PR #5, opened 2025-05-15
feat: Implement grpc-gateway in public API

forestileao:
- Implement grpc-gateway in public API
- Add protoc-gen for open api specs


Makefile
```diff
@@ -61,4 +61,21 @@ pb.gen:
...
+
+# Generate REST API and OpenAPI spec for Delivery service
+pb.gen.rest: pb.deps.rest
+	@echo "Generating gRPC-Gateway and OpenAPI files for Delivery service..."
+	rm -rf $(TMP_REPO_DIR)
+	git clone git@github.com:renderedtext/internal_api.git $(TMP_REPO_DIR) && (cd $(TMP_REPO_DIR) && git checkout $(INTERNAL_API_BRANCH) && cd -)
+	docker-compose run --rm --no-deps app /app/scripts/protoc_gateway.sh delivery $(INTERNAL_API_BRANCH) $(TMP_REPO_DIR)
 	rm -rf $(TMP_REPO_DIR)
+
+pb.deps.rest:
```
lucaspin: Why do we need to install them here if they are installed in the dockerfile?

scripts/protoc_gateway.sh
```diff
@@ -0,0 +1,106 @@
...
+  if [ ! -d "$OPENAPI_DIR" ]; then
+    echo "$(bold "Installing OpenAPI proto files")"
+    mkdir -p $OPENAPI_DIR
+    curl -L https://raw.githubusercontent.com/grpc-ecosystem/grpc-gateway/main/protoc-gen-openapiv2/options/annotations.proto > $OPENAPI_DIR/annotations.proto
+    curl -L https://raw.githubusercontent.com/grpc-ecosystem/grpc-gateway/main/protoc-gen-openapiv2/options/openapiv2.proto > $OPENAPI_DIR/openapiv2.proto
+  fi
+}
+
+# Check if the proto file already has the necessary imports
+check_and_add_imports() {
```
lucaspin: This will be made a lot simpler after https://github.com/superplanehq/superplane/pull/4 is merged, since then we can move the proto definition to this repo as well, and put all the annotations we need directly on it

api/swagger/delivery.swagger.json
lucaspin, line 1: Should we add this to version control or generate it every time during the build?

lucaspin (approved):

PR #8, opened 2025-05-15
fix: remove organization ID from resources

lucaspin:
Before, we were associating canvases/sources/stages to a Semaphore organization ID, but now that superplane is its own thing, we can't have that anymore, and it is causing issues when running the application locally – no x-semaphore-org-id headers, for example. Here, I'm removing the organization relationship altogether, which allows us to run superplane locally.

NOTE: we will probably need some kind of namespaces (organizations) in superplane as well, but for now, we don't need to worry about that.

### Other changes

- Add the swagger JSON to source control
- Fix issues with calling the Semaphore API for triggering tasks


.gitignore
```diff
@@ -1,2 +1 @@
 junit-report.xml
-api/swagger/superplane.swagger.json
```
hamir-suspect, line 7: 👍

hamir-suspect (dismissed):

hamir-suspect (approved):

PR #10, opened 2025-05-20
feat: update stage + initial CLI

lucaspin:
### Initial CLI

Initial version of CLI is added in cmd/cli and pkg/cli. It has following commands: get, list, create, update, and approve. The CLI uses an API client generated with [openapi-generator](https://github.com/OpenAPITools/openapi-generator), and placed in pkg/openapi_client

### Cleaning up tag related endpoints/fields from API

Our understanding of tags is still not clear, but we do know that the way they are represented currently in the API is not OK, so I'm removing the ListTags and UpdateTagState endpoints, and also the TagUsageDefinition from stage resource.
 
### UpdateStage endpoint implementation

Not having a way to update the stage makes it hard to play around with things, so an implementation of that is also added in this PR.


hamir-suspect (approved):

hamir-suspect:
@lucaspin can you also introduce API for executions?

lucaspin:
@hamir-suspect yes. I'll do it in a separate PR though

PR #22, opened 2025-05-27
feat: add name queries for canvas

forestileao:
##  Changes
- Add name queries for Canvas
- Add cli option (canvasName) for yaml files
- Updated OpenApiSpec

### DIFF
--- a/pkg/cli/approve.go
+++ b/pkg/cli/approve.go
@@ -11,16 +11,17 @@ import (
 )
 
 var approveEventCmd = &cobra.Command{
-	Use:     "event [CANVAS_ID_OR_NAME] [STAGE_ID] [EVENT_ID]",
+	Use:     "event [EVENT_ID]",
 	Short:   "Approve a stage event",
 	Long:    `Approve a pending stage event that requires approval.`,
 	Aliases: []string{"events"},
-	Args:    cobra.ExactArgs(3),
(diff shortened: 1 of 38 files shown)


forestileao:
@lucaspin I know we don't have canvases aggregators, like org_id. But should I add an unique index for canvas names?

pkg/cli/approve.go
```diff
@@ -11,14 +11,14 @@ import (
 )
 
 var approveEventCmd = &cobra.Command{
-	Use:     "event [CANVAS_ID] [STAGE_ID] [EVENT_ID]",
+	Use:     "event [CANVAS_ID_OR_NAME] [STAGE_ID] [EVENT_ID]",
```
lucaspin: I think it would be better to have the canvas ID or name and stage ID or name as flags instead of parameters here, kinda like the [semaphoreci/cli](https://github.com/semaphoreci/cli) does with `--project-id` and `--project-name` flags. What do you think?

lucaspin:
@forestileao 

>  I know we don't have canvases aggregators, like org_id. But should I add an unique index for canvas names?

Yes. We can change that index later to include the org_id

forestileao:
@lucaspin canvas name unique index are already added there.
Added stage names queries as well.
Set these things as flags in the cli.

PR #24, opened 2025-05-30
feat: add canvas sidebar + show stage executions

forestileao:
### Changes
- Add template sidebar to show the full state of the stages
- Set NoOpEncryptor for workers, on tests only

https://github.com/user-attachments/assets/39e6fe54-22e2-44e0-9f7e-8084051d37d7


forestileao:
@hamir-suspect please check if the directory structure is good as well. Or if we are following a different pattern.

hamir-suspect (approved):

PR #25, opened 2025-06-02
feat: stage inputs and outputs

lucaspin:
This PR introduces an input/output system for stages in Superplane, enhancing the ability to manage and validate data flow between stages.

### Inputs

- Added support for inputs and input_mappings in stages
- Updated the execution flow to handle inputs and store them in stage events
- Updated CLI commands to support new input/output functionality
- Added comprehensive test coverage for input/output validation and processing

### Outputs

- Added output definitions to stages
- Updated API endpoint to use `/outputs` instead of `/executions/{id}/outputs`. This allows outputs to be pushed for the local environment as well.
- If a stage defines a required output, and that output is not pushed in the execution, the execution fails.

## Other changes

- Updated Go version to 1.24, due to the new version of Air requiring it.
- After iterating on the inputs/outputs design, it became quite clear that `runTemplate` is poor choice of word, so we also  rename `run_template` to `executor_spec` in this pull request.


docs/concepts/inputs.md
```diff
@@ -0,0 +1,115 @@
...
+      parameters:
+        - name: DOCS_VERSION
+          value: ${{ inputs.DOCS_VERSION }}
+        - name: TERRAFORM_VERSION
+          value: ${{ inputs.TERRAFORM_VERSION }}
+```
+
+- The `inputs` field are the input definitions.
+- The `inputMappings` field allow you to assign values to your inputs based on the connection that is currently hitting your stage. When an event comes from a connection, the input mapping for that connection is chosen and applied to compute the input set that will be used when the execution starts.
+- The values that come from the `inputMappings` are made available in the executor spec field through the use of the `${{ inputs.* }}` syntax.
```
forestileao, line 71: 👍

pkg/cli/create.go
```diff
@@ -93,8 +93,8 @@ var createCmd = &cobra.Command{
 
 			fmt.Printf("Event Source '%s' created with ID '%s'.\n",
 				*response.EventSource.Name, *response.EventSource.Id)
-			fmt.Printf("API Key: %s\n", *response.Key)
-			fmt.Println("Save this key as it won't be shown again.")
+			fmt.Printf("Key: %s\n", *response.Key)
+			fmt.Println("! Save this key as it won't be shown again.")
```
forestileao, line 97:
```suggestion
			fmt.Println("\033[31m! Save this key as it won't be shown again.\033[0m")
```

Maybe we can make it RED, just a suggestion

pkg/inputs/input_builder.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (b *InputBuilder) getValueFromMap(m map[string]any, inputName string) (any, error) {
+	for k, v := range m {
+		if k == inputName {
+			return v, nil
+		}
+	}
+
+	return nil, fmt.Errorf("value for %s not found in map: %v", m, inputName)
+}
```
forestileao:
```suggestion
func (b *InputBuilder) getValueFromMap(m map[string]any, inputName string) (any, error) {
    if value, exists := m[inputName]; exists {
        return value, nil
    }
    return nil, fmt.Errorf("value for %s not found in map", inputName)
}
```

forestileao (dismissed):

pkg/cli/create.go
lucaspin: I would prefer to use something like https://github.com/fatih/color instead of hardcoding these things here. But I wouldn't like to pull in that dependency in this PR

forestileao (approved):

PR #28, opened 2025-06-02
feat: add inputs and outputs to UI

forestileao:
## Changes

Created ui to display inputs and ouputs.
Input mappings are also displayed within the Inputs.
It handles static values or stage-referenced inputs from ValueFrom.

https://github.com/user-attachments/assets/26d8f448-434f-4475-a824-e569f7723478

### DIFF
--- a/web_src/src/pages/canvas/components/tabs/SettingsTab.tsx
+++ b/web_src/src/pages/canvas/components/tabs/SettingsTab.tsx
@@ -1,3 +1,4 @@
+import { InputMappingValueDefinition } from "@/api-client";
 import { StageWithEventQueue } from "../../store/types";
 
 interface SettingsTabProps {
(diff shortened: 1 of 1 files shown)


hamir-suspect (approved):

web_src/src/pages/canvas/components/tabs/SettingsTab.tsx
```diff
@@ -1,10 +1,54 @@
...
 interface SettingsTabProps {
   selectedStage: StageWithEventQueue;
 }
 
 export const SettingsTab = ({ selectedStage }: SettingsTabProps) => {
+  const getInputMapping = (inputName: string) => {
+    if (!selectedStage.inputMappings) return null;
+    
+    for (const mapping of selectedStage.inputMappings) {
+      const valueMapping = mapping.values?.find(v => v.name === inputName);
```
lucaspin, line 13: I don't think this is correct. It seems like it's always finding the first one, but we can have cases where the same input is defined differently depending on the source currently triggering the stage.

For example, this stage:

```
connections:
  - type: TYPE_EVENT_SOURCE
    name: source-1
  - type: TYPE_EVENT_SOURCE
    name: source-2

inputs:
  - name: VERSION_A
  - name: VERSION_B

inputMappings:
  - when:
      triggeredBy:
        connection: source-1
    values:
      - name: VERSION_A
        valueFrom:
          eventData:
            connection: source-1
            expression: ref
      - name: VERSION_B
        valueFrom:
          lastExecution:
            results: [RESULT_PASSED]
  
  - when:
      triggeredBy:
        connection: source-2
    values:
      - name: VERSION_A
        valueFrom:
          lastExecution:
            results: [RESULT_PASSED]
      - name: VERSION_B
        valueFrom:
          eventData:
            connection: source-1
            expression: ref
```

In this case, `VERSION_A` is defined from the event data when the connection triggering the stage is source-1, but it is defined from the last execution when the connection triggering the stage is source-2

PR #29, opened 2025-06-03
fix: websocket connection reliable and stable

hamir-suspect:
## Websocket changes
- Use different websocket library in react app
- better event routing in react
-  heartbeat (pings) from server and set to 10s for quick detection of dropped or inactive connections

## Other changes

- Use logrus for logging in tackle and logging middleware

### DIFF
--- a/web_src/index.html
+++ b/web_src/index.html
@@ -9,7 +9,7 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
     <link href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" rel="stylesheet">
 
-    <script src="http://localhost:8097"></script>
+    <!-- <script src="http://localhost:8097"></script> -->
 
     <title>Vite + React + TS</title>
   </head>


cmd/server/main.go
```diff
@@ -14,6 +14,10 @@ import (
 )
 
 func startWorkers(jwtSigner *jwt.Signer, encryptor encryptor.Encryptor) {
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp: true,
+		TimestampFormat: time.StampMilli,
+	})
```
pedroforesti: Do we need to set up 2 times

cmd/server/main.go
```diff
@@ -131,7 +135,10 @@ func startPublicAPI(encryptor encryptor.Encryptor, jwtSigner *jwt.Signer) {
 }
 
 func main() {
-	log.SetFormatter(&log.TextFormatter{TimestampFormat: time.StampMilli})
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp:    true,
+		TimestampFormat: time.StampMilli,
+	})
```
pedroforesti, line 137: Related to the comment from before

web_src/index.html
```diff
@@ -9,7 +9,7 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
     <link href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" rel="stylesheet">
 
-    <!-- <script src="http://localhost:8097"></script> -->
+    <script src="http://localhost:8097"></script>
```
forestileao: isnt it going to break in sxmoon

PR #33, opened 2025-06-04
fix: clean up event source node type

hamir-suspect:


docs/examples/secret_example.yml
```diff
@@ -0,0 +1,10 @@
+kind: Secret
+metadata:
+  name: api-credentials
+  canvasId: my-application-canvas  # Reference to the canvas this secret belongs to
+spec:
+  provider: local  # Currently only local provider is supported
```
lucaspin: spec.provider is `PROVIDER_LOCAL` and keys are under the `local.data` field instead

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: api-credentials
  canvasId: my-application-canvas
spec:
  provider: PROVIDER_LOCAL
  local:
    data:
      api-key: "your-api-key-here"
      api-secret: "your-api-secret-here"
```

docs/examples/stage2_example.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  # Define connections to event sources or other stages
+  connections:
+    - type: TYPE_STAGE
+      name: deploy-to-staging # Reference to the stage defined earlier
+      # Optional filters can be defined to determine when this stage should be triggered
+      filters:
+        - type: FILTER_TYPE_DATA
+          data:
+            expression: "$.DEPLOY_URL" # Only trigger on main branch
```
lucaspin: We don't need to use `$` to access fields in the event, and the filter expression must return a boolean

docs/examples/stage2_example.yml
```diff
@@ -0,0 +1,68 @@
...
+  # Define inputs that the stage requires
+  inputs:
+    - name: VERSION
+      description: "Version to deploy"
+    - name: ENVIRONMENT
+      description: "Target environment"
+
+  # Define how inputs should be mapped based on what triggered the stage
+  inputMappings:
+    - when:
```
lucaspin, line 48: When you only have one connection, you can omit the `when` field, and just have `values`

docs/examples/stage_example.yml
```diff
@@ -0,0 +1,70 @@
...
+
+  # Define connections to event sources or other stages
+  connections:
+    - type: TYPE_EVENT_SOURCE
+      name: github-webhook  # Reference to the event source defined earlier
+      # Optional filters can be defined to determine when this stage should be triggered
+      filters:
+        - type: FILTER_TYPE_DATA
+          data:
+            expression: "$.ref == 'refs/heads/main'"  # Only trigger on main branch
```
lucaspin: Same comment about `$` usage

docs/examples/stage_example.yml
```diff
@@ -0,0 +1,70 @@
...
+        - type: FILTER_TYPE_HEADER
+          header:
+            expression: "$.X-GitHub-Event == 'push'"  # Only trigger on push events
```
lucaspin: I think this condition should be `headers['X-GitHub-Event'] == 'push'`, right @forestileao ?

docs/examples/stage_example.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Define inputs that the stage requires
+  inputs:
+    - name: VERSION
+      description: "Version to deploy"
+    - name: ENVIRONMENT
+      description: "Target environment"
+
+  # Define how inputs should be mapped based on what triggered the stage
+  inputMappings:
+    - when:
```
lucaspin: Same comment about single connection mappings

docs/examples/stage_example.yml
```diff
@@ -0,0 +1,70 @@
...
+        triggeredBy:
+          connection: github-webhook
+      values:
+        - name: VERSION
+          valueFrom:
+            eventData:
+              connection: github-webhook
+              expression: "$.after.substring(0, 7)"  # Use first 7 chars of commit SHA
```
lucaspin: [expr-lang/lang](https://expr-lang.org/docs/language-definition) doesn't have a substring() function. I think you'd need to do something like `after | split("") | take(7) | join("")` here

docs/examples/secret_example.yml
lucaspin: Also, the `metadata.canvasId` expects the canvas UUID, not the name

docs/examples/stage2_example.yml
hamir-suspect: It's there to showcase how it's done

docs/examples/stage_example.yml
lucaspin: Actually, I think you can do `after[0:7]` too

docs/examples/stage_example.yml
pedroforesti: that is correct, since headers are case insensitive we needed to add this way, and it doesnt matter if you write 
headers['X-GitHub-Event') or headers['x-github-event') or headers['X-GITHub-EveNt')

lucaspin:
@hamir-suspect shouldn't you also remove the `cmd/fixtures/main.go` file?

PR #37, opened 2025-06-06
feat: Add RBAC with domains Authorization

forestileao:
## Changes
- Created the Authorization Service implementation to deal with
  - Roles
  - Inheritance
  - Groups
  - Permission checking
- Implemented RBAC with domain logic described by [Casbin](https://casbin.org/pt/docs/rbac-with-domains)

- Created middleware to be used once authentication is done
- Exposed endpoints to the API

image

## Future Improvements
- With the authentication design, it will need to be implemented:
  - Organizations Api
  - Users Api


docker-compose.dev.yml
```diff
@@ -11,14 +11,17 @@ services:
     environment:
       DB_HOST: "db"
       DB_PORT: "5432"
-      DB_NAME: "superplane_dev"
+      DB_NAME: "superplane"
```
lucaspin: Why this change?

pkg/grpc/.DS_Store
lucaspin, line 1: Do not commit .DS_Store files

pkg/grpc/actions/assign_role.go
```diff
@@ -0,0 +1,46 @@
+package actions
```
lucaspin: Lots of things into `pkg/grpc/actions` now. Maybe we should create `pkg/grpc/actions/auth` and `pkg/grpc/actions/stage`, `pkg/grpc/actions/canvas` to separate things a bit. What do you think?

pedroforesti: No problem, I will make this refactor

docker-compose.dev.yml
pedroforesti: I messed up with some env variables at the time that I was testing. This should not have been commited. I'll remove, just like the .DS_Store

pkg/authorization/interface.go
```diff
@@ -0,0 +1,52 @@
+package authorization
+
+type AuthorizationServiceInterface interface {
```
lucaspin: Appending an `Interface` to the end of an interface's name is a bit weird, maybe just `Authorization` is OK here

pkg/grpc/actions/check_permission.go
```diff
@@ -0,0 +1,42 @@
...
+		return nil, status.Error(codes.InvalidArgument, "domain type must be specified")
+	}
+	if req.Resource == "" || req.Action == "" {
+		return nil, status.Error(codes.InvalidArgument, "resource and action must be specified")
+	}
+
+	var allowed bool
+	switch req.DomainType {
+	case pb.DomainType_DOMAIN_TYPE_ORGANIZATION:
+		allowed, err = authService.CheckOrganizationPermission(req.UserId, req.DomainId, req.Resource, req.Action)
```
lucaspin: Do we really need separate `CheckOrganizationPermission` and `CheckCanvasPermission` methods, if they both receive the same arguments?

pkg/authorization/interface.go
lucaspin: Also, this interface has a lot of methods, which is usually considered a not-so-great practice. Is there any way to split into more meaningful ones?

pkg/grpc/actions/check_permission.go
pedroforesti: Inside their implementation, there is a generic checkPermission that receives a DomainType to differentiate it.
IMO this is easier to use than adding another parameter.

pkg/authorization/interface.go
pedroforesti: It's possible to define many interfaces and make the authservice implement a bunch of them. But we cannot split the struct due to the cached enforcer. Unless we inject the enforcer, but then it would not make sense in case we switch the library in the future (Even if the probability is very low).

forestileao:
@lucaspin, I Removed the unecessary endpoints that were exposed

PR #39, opened 2025-06-09
chore: refactor usage of multiple stores

hamir-suspect:


web_src/src/pages/canvas/store/canvasStore.ts
```diff
@@ -124,4 +129,92 @@ export const useCanvasStore = create<CanvasState>((set, get) => ({
   updateWebSocketConnectionStatus: (status) => {
     set({ webSocketConnectionStatus: status });
   },
+
+  syncToReactFlow: () => {
+    const { stages, event_sources, nodePositions, approveStageEvent } = get();
+
+    // Use the transformer functions from flowTransformers.ts
+    const stageNodes = transformStagesToNodes(stages, nodePositions, approveStageEvent);
+    const eventSourceNodes = transformEventSourcesToNodes(event_sources, nodePositions);
```
forestileao, line 137: This will make the autolayout be activated all the time an event appears? But I also don't see other way to make it better than it.

forestileao (dismissed):

web_src/src/pages/canvas/store/canvasStore.ts
pedroforesti: Removed the use effect from the flow rendered and applied the auto layout directly on initialization. This will also avoid the inconsistency of showing the nodes unlayouted at first render.

pedroforesti (approved):

forestileao (approved):

PR #46, opened 2025-06-16
Feat/authentication

forestileao:


forestileao:
Added a new authentication method for cli as well. Now the CLI authenticates using github Device flow:
https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#device-flow

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/superplanehq/superplane/pkg/jwt"
+	"github.com/superplanehq/superplane/pkg/models"
+)
+
+type AuthenticationHandler struct {
+	jwtSigner *jwt.Signer
+}
+
+type AuthenticationUser struct {
```
lucaspin: No need to prefix the struct with `Authentication`, since the package is already called `authentication`, so having `authentication.User` and `authentication.Handler` is probably better than `authentication.AuthenticationHandler` and authentication.AuthenticationUser

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+
+type GitHubUserInfo struct {
+	ID        int    `json:"id"`
+	Login     string `json:"login"`
+	Name      string `json:"name"`
+	Email     string `json:"email"`
+	AvatarURL string `json:"avatar_url"`
+}
+
+func NewAuthHandler(jwtSigner *jwt.Signer) *AuthenticationHandler {
```
lucaspin: Same comment about the prefix - `NewHandler` is probably better

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+// RegisterRoutes adds authentication routes to the router
+func (a *AuthenticationHandler) RegisterRoutes(router *mux.Router) {
+	router.HandleFunc("/auth/me", a.handleMe).Methods("GET")
+	router.HandleFunc("/logout", a.handleLogout).Methods("GET")
+	router.HandleFunc("/login", a.handleLoginPage).Methods("GET")
+
+	// Token exchange route for CLI
+	router.HandleFunc("/auth/token/exchange", a.handleTokenExchange).Methods("POST")
+
+	if os.Getenv("APP_ENV") == "development" {
```
lucaspin: We should replace the `os.Getenv("APP_ENV")` with a configuration that is passed to the Handler. Hiding os.Getenv() calls on lower level packages can be hard to follow

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+	}
+
+	dbUser, err := models.FindUserByID(accountProvider.UserID.String())
+	if err != nil {
+		log.Errorf("Failed to find user by ID %s: %v", accountProvider.UserID.String(), err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+
+	dbUser.Email = githubUser.Email
```
lucaspin: Why do we need name/email/avatar in the user record too?

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+		log.Infof("User already authenticated: %s", gothUser.Email)
+		a.handleSuccessfulAuth(w, r, gothUser)
+	} else {
+		log.Info("Starting OAuth flow")
+		gothic.BeginAuthHandler(w, r)
+	}
+}
+
+func (a *AuthenticationHandler) handleDevAuth(w http.ResponseWriter, r *http.Request) {
+	if os.Getenv("APP_ENV") != "development" {
```
lucaspin: Same comment about os.Getenv

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+
+	if callbackURL != "" {
+		parsedURL, err := url.Parse(callbackURL)
+		if err != nil {
+			http.Error(w, "Invalid callback URL", http.StatusBadRequest)
+			return
+		}
+
+		// Security check - only allow localhost callbacks
+		if parsedURL.Hostname() != "localhost" && parsedURL.Hostname() != "127.0.0.1" {
```
lucaspin: Why do we need this check?

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+func (a *AuthenticationHandler) findOrCreateUserAndAccount(gothUser goth.User) (*models.User, *models.AccountProvider, error) {
+	accountProvider, err := models.FindAccountProviderByProviderID(gothUser.Provider, gothUser.UserID)
+	if err == nil {
+		accountProvider.Username = gothUser.NickName
+		accountProvider.Email = gothUser.Email
+		accountProvider.Name = gothUser.Name
+		accountProvider.AvatarURL = gothUser.AvatarURL
+		accountProvider.AccessToken = gothUser.AccessToken
+		accountProvider.RefreshToken = gothUser.RefreshToken
+		if gothUser.ExpiresAt != (time.Time{}) {
```
lucaspin: You can use `gothUser.ExpiresAt.IsZero()` for this check

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+	userID := claims["sub"].(string)
+	user, err := models.FindUserByID(userID)
+	if err != nil {
+		return nil, fmt.Errorf("user not found: %v", err)
+	}
+
+	return user, nil
+}
+
+func (a *AuthenticationHandler) AuthMiddleware(next http.Handler) http.Handler {
```
lucaspin: Same comment about `Auth` prefix

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,654 @@
...
+		if err != nil {
+			log.Errorf("User not found: %v", err)
+			if r.Header.Get("Accept") == "application/json" {
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
+			} else {
+				http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+			}
+			return
+		}
+		log.Infof("User %s authenticated", user.Email)
```
lucaspin: Logging this every time will generate too many logs. Let's just rely on the logging middleware for now

pkg/models/repo_host_account.go
lucaspin, line 1: Please rename the file to account_provider.go too

pkg/authentication/authentication.go
forestileao: I'll keep only the Name, so we can have a different display name in the future

pkg/authentication/authentication.go
forestileao: We don't need this entire block of code anymore, since it was for the first cli authentication method I tried. Before the implementation of the Device Flow

pkg/cli/login.go
```diff
@@ -0,0 +1,445 @@
...
+		fmt.Println("Please open the URL manually in your browser.")
+	}
+}
+
+var logoutCmd = &cobra.Command{
+	Use:   "logout",
+	Short: "Logout from GitHub",
+	Long:  `Remove stored authentication token.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		viper.Set(ConfigKeyAuthToken, "")
```
lucaspin, line 398: Why isn't this calling the /logout endpoint?

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,613 @@
...
+			http.Error(w, "Account exists but GitHub provider not connected. Please connect GitHub through the web interface.", http.StatusNotFound)
+			return
+		}
+	}
+
+	accountProvider.Username = githubUser.Login
+	accountProvider.Email = githubUser.Email
+	accountProvider.Name = githubUser.Name
+	accountProvider.AvatarURL = githubUser.AvatarURL
+	accountProvider.AccessToken = req.GitHubToken
```
lucaspin: We shouldn't store this as plain-text in the database

pkg/cli/login.go
forestileao: In CLI, we are just storing the JWT token.
The flow for CLI authentication is: we get the github token from "device flow" (that we enabled when we created the client and secret keys). Then, we just make a token exchange from github to our system's token.
There is no gothic session / cookie associated with the authentication from CLI.

pkg/authentication/authentication.go
forestileao: Sure, let me add the encryptor

pkg/authentication/authentication.go
```diff
@@ -0,0 +1,630 @@
...
+		}
+	}
+
+	encryptedAccessToken, err := a.encryptor.Encrypt(context.Background(), []byte(req.GitHubToken), []byte(githubUser.Email))
+	if err != nil {
+		log.Errorf("Failed to encrypt access token: %v", err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+	accountProvider.AccessToken = string(encryptedAccessToken)
```
lucaspin: Not sure you can just `string()` like this. I think we should either make that column a bytea, or if we keep that column as a string, we should base64 encode it here

forestileao: Updated to use base64

lucaspin (approved):

PR #48, opened 2025-06-24
feat: implement organizations and auth interceptor

forestileao:
### Changes
- Implemented organizations CRUD
- Implemented Permissions interceptor for each service we have
- Updated CLI to work with organization ID as well
- Regenerated open api clients and swagger files


pkg/authorization/interceptor.go
```diff
@@ -0,0 +1,170 @@
...
+func (a *AuthorizationInterceptor) UnaryInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		rule, requiresAuth := a.rules[info.FullMethod]
+		if !requiresAuth {
+			return handler(ctx, req)
+		}
+
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			log.Errorf("User not found in context, metadata %v", md)
```
lucaspin: Log message here is misleading. Something like "Metadata not found in context" is probably better

pkg/authorization/service.go
```diff
@@ -326,6 +326,28 @@ func (a *AuthService) SetupOrganizationRoles(orgID string) error {
 	return nil
 }
 
+func (a *AuthService) DestroyOrganizationRoles(orgID string) error {
+	domain := fmt.Sprintf("org:%s", orgID)
+	ok, err := a.enforcer.RemoveFilteredGroupingPolicy(2, domain)
```
lucaspin, line 331: What is 2?

pkg/authorization/service.go
```diff
@@ -326,6 +326,28 @@ func (a *AuthService) SetupOrganizationRoles(orgID string) error {
...
+
+	if err != nil {
+		return fmt.Errorf("failed to remove organization roles: %w", err)
+	}
+	if !ok {
+		return fmt.Errorf("organization roles not found for %s", orgID)
+	}
+
+	ok, err = a.enforcer.RemoveFilteredPolicy(1, domain)
```
lucaspin, line 340: What is 1?

pkg/grpc/actions/canvases/list_canvases_test.go
```diff
@@ -4,22 +4,73 @@ import (
...
+		require.NotNil(t, res)
+		require.Len(t, res.Canvases, 1)
+		require.NotNil(t, res.Canvases[0].Metadata)
+		assert.Equal(t, r.Canvas.ID.String(), res.Canvases[0].Metadata.Id)
+		assert.Equal(t, r.Canvas.Name, res.Canvases[0].Metadata.Name)
+		assert.Equal(t, r.Canvas.CreatedBy.String(), res.Canvases[0].Metadata.CreatedBy)
+		assert.NotNil(t, res.Canvases[0].Metadata.CreatedAt)
+	})
+
+	t.Run("invalid organization ID -> list canvases from organization", func(t *testing.T) {
```
lucaspin: Test name is misleading. "Organization with no canvases -> empty list" or something

pkg/grpc/actions/organizations/delete_organization.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/superplanehq/superplane/pkg/database"
+	"github.com/superplanehq/superplane/pkg/grpc/actions"
+	"github.com/superplanehq/superplane/pkg/models"
+	pb "github.com/superplanehq/superplane/pkg/protos/organizations"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"gorm.io/gorm"
+)
+
+func DeleteOrganization(ctx context.Context, req *pb.DeleteOrganizationRequest, authorizationService authorization.Authorization) (*pb.DeleteOrganizationResponse, error) {
```
lucaspin, line 19: Is this deleting all the canvases/sources/stages/events/... for the organization?

pkg/grpc/actions/organizations/describe_organization.go
```diff
@@ -0,0 +1,59 @@
...
+
+func DescribeOrganization(ctx context.Context, req *pb.DescribeOrganizationRequest) (*pb.DescribeOrganizationResponse, error) {
+	if req.IdOrName == "" {
+		return nil, status.Error(codes.InvalidArgument, "id_or_name is required")
+	}
+
+	var organization *models.Organization
+	var err error
+
+	if _, parseErr := uuid.Parse(req.IdOrName); parseErr == nil {
```
lucaspin: Why uuid.Parse() and then actions.ValidateUUIDs() again?

pkg/grpc/actions/organizations/list_organizations_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/authentication"
+	"github.com/superplanehq/superplane/pkg/authorization"
+	"github.com/superplanehq/superplane/pkg/database"
+	"github.com/superplanehq/superplane/pkg/grpc/actions/auth"
+	"github.com/superplanehq/superplane/pkg/models"
+	protos "github.com/superplanehq/superplane/pkg/protos/organizations"
+)
+
+func Test__ListOrganizations(t *testing.T) {
```
lucaspin, line 18: We should have tests to check that an organization that is not accessible to a user is not returned in the response when listing organizations.

pkg/public/server.go
```diff
@@ -139,17 +145,34 @@ func (s *Server) RegisterGRPCGateway(grpcServerAddr string) error {
...
 	s.Router.PathPrefix("/api/v1/authorization").Handler(protectedGRPCHandler)
 	s.Router.PathPrefix("/api/v1/canvases").Handler(protectedGRPCHandler)
+	s.Router.PathPrefix("/api/v1/organizations").Handler(protectedGRPCHandler)
 
 	return nil
 }
 
+func headersMatcher(key string) (string, bool) {
+	switch key {
+	case "X-User-Id":
```
lucaspin, line 166: Should we allow the X-User-Id header to be passed directly like this instead of the cookie/token? That allows me to impersonate another user, no?

forestileao: Actually, this header is something that we set right here: https://github.com/superplanehq/superplane/pull/48/files#diff-a5008ad28108159760e66faef71fc7683c161b2aafd94f03782528aef51f04beR174

We make this way by header because the grpc gateway re-creates the context and only translates headers to metadata. And cookies as well. But since this is something that WE set from the authentication, I don't see it happening. If it exists in the request, it would be overridden by us.

pkg/grpc/actions/organizations/delete_organization.go
forestileao: It's not, should we also do this? or make a soft-delete instead?

pkg/authorization/service.go
forestileao: This is from the Casbin API, which means that the filter I am applying is on index 1 of the policy, for example:

Here is the format for a policy:
`role:owner,domainId,resource,action`
The domainId is on index 1.

This command would be the same as filtering like:
RemoveFilteredPolicy(0, "",  domain)
Then the empty string would be ignored.

pkg/authorization/service.go
forestileao: Same thing from comment below, but the format for grouping policies are:

`role_or_user_or_group,role_name,domain`

and domain is on index 2

pkg/grpc/actions/organizations/delete_organization.go
lucaspin: We should probably have both, but for now, let's make it a soft-delete instead. We can add the hard deletion later

pkg/public/server.go
lucaspin: Ok. Can we reject requests that come with this header set then instead?

forestileao: Yep, good idea. I'll apply this change

db/migrations/20250618171050_add-organizations.up.sql
```diff
@@ -0,0 +1,15 @@
...
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name VARCHAR(255) NOT NULL UNIQUE,
+    display_name VARCHAR(255) NOT NULL,
+    created_by UUID NOT NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+
+ALTER TABLE CANVASES ADD COLUMN organization_id UUID NOT NULL DEFAULT '00000000-0000-0000-0000-000000000000';
```
lucaspin: We should make organization_id a foreign key too, no?

pkg/grpc/actions/organizations/delete_organization.go
forestileao: Implemented organization soft deletion with gorm.DeletedAt managed columns. It automatically set `Deleted_at is NULL` in queries as well.

PR #56, opened 2025-06-30
feat: timeout for connection group field sets

lucaspin:
This PR introduces a timeout mechanism for connection groups, allowing users to define how the system should handle incomplete field sets when waiting for events from multiple connections.

## Changes

- New `timeout` field to control how the system should handle incomplete field sets. By default, no timeout is used. Timeout must be between 60s and 24h.
- New `timeout_behavior` field to control what to do when the timeout for a connection group field set is reached: (1) `drop`, which does not emit any events for a field set that timed out, and (2) `emit`, which emits an event but indicate some connections were missing.
- New `PendingFieldSetsWorker` to monitor and process pending field sets that have timeouts applied.


pkg/workers/pending_field_sets_worker.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			log.Errorf("Error processing field set %s: %v", fieldSet.ID, err)
+		}
+	}
+
+	return nil
+}
+
+func (w *PendingFieldSetsWorker) ProcessFieldSet(fieldSet models.ConnectionGroupFieldSet) error {
+	connectionGroup, err := models.FindConnectionGroupByID(database.Conn(), fieldSet.ConnectionGroupID)
```
forestileao: Why do we pass database.conn() as a parameter in some places, and in some places we don't?
i.e.: models.FindConnectionGroupByID and models.ListPendingConnectionGroupFieldSets()

lucaspin: Oversight on my part. Usually, I use the `X` and `XInTransaction` pattern, but didn't apply it here. Fixed it now.

forestileao (approved):

PR #68, opened 2025-07-02
feat: groups for canvas

forestileao:
## Changes
This PR:
- Modularizes the group management in many go files
- Adds the ability to add groups for canvas roles
- Changes the group responses to a more structured way
- Change proto to listgroups instead of list organization groups

### DIFF
--- a/pkg/grpc/actions/auth/get_group_users.go
+++ b/pkg/grpc/actions/auth/get_group_users.go
@@ -37,14 +37,9 @@ func GetGroupUsers(ctx context.Context, req *GetGroupUsersRequest, authService a
 		return nil, status.Error(codes.Internal, "failed to get group users")
 	}
 
-	roles, err := authService.GetGroupRoles(req.DomainID, domainType, req.GroupName)
+	role, err := authService.GetGroupRole(req.DomainID, domainType, req.GroupName)
 	if err != nil {
 		return nil, status.Error(codes.Internal, "failed to get group roles")
-	}
-
-	var role string
-	if len(roles) > 0 {
-		role = roles[0]
 	}
 
 	group := &pb.Group{
--- a/pkg/grpc/actions/auth/list_groups.go
+++ b/pkg/grpc/actions/auth/list_groups.go
@@ -32,14 +32,9 @@ func ListGroups(ctx context.Context, req *GroupRequest, authService authorizatio
 
 	groups := make([]*pb.Group, len(groupNames))
 	for i, groupName := range groupNames {
-		roles, err := authService.GetGroupRoles(req.DomainID, domainType, groupName)
+		role, err := authService.GetGroupRole(req.DomainID, domainType, groupName)
 		if err != nil {
 			return nil, status.Error(codes.Internal, "failed to get group roles")
-		}
-
-		var role string
-		if len(roles) > 0 {
-			role = roles[0]
 		}
 
 		groups[i] = &pb.Group{
--- a/pkg/authorization/interface.go
+++ b/pkg/authorization/interface.go
@@ -13,7 +13,7 @@ type GroupManager interface {
 	RemoveUserFromGroup(domainID string, domainType string, userID string, group string) error
 	GetGroupUsers(domainID string, domainType string, group string) ([]string, error)
 	GetGroups(domainID string, domainType string) ([]string, error)
-	GetGroupRoles(domainID string, domainType string, group string) ([]string, error)
+	GetGroupRole(domainID string, domainType string, group string) (string, error)
 }
 
 // Role management interface
--- a/pkg/authorization/service.go
+++ b/pkg/authorization/service.go
@@ -230,7 +230,7 @@ func (a *AuthService) GetGroups(domainID string, domainType string) ([]string, e
 	return groups, nil
 }
 
-func (a *AuthService) GetGroupRoles(domainID string, domainType string, group string) ([]string, error) {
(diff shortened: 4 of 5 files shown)


protos/authorization.proto
```diff
@@ -184,11 +185,11 @@ service Authorization {
   //
   rpc GetGroupUsers(GetGroupUsersRequest) returns (GetGroupUsersResponse) {
     option (google.api.http) = {
-      get: "/api/v1/authorization/organizations/{org_id}/groups/{group_name}/users"
+      get: "/api/v1/authorization/groups/users"
```
lucaspin: I think think it makes more sense to have the `{group_id_or_name}` in the URL, since we use this pattern in all the other endpoints

pedroforesti: No problem, just wanted to make them shorter.

forestileao: Updated. BTW, I set {group_name} instead of {group_id_or_name} since on casbin we use the role/group name  as their id

pkg/grpc/actions/auth/create_group.go
```diff
@@ -11,29 +11,41 @@ import (
...
 
-	if req.Role == "" {
-		return nil, status.Error(codes.InvalidArgument, "role must be specified")
+	domainType, err := ConvertDomainType(req.DomainType)
+	if err != nil {
+		return nil, err
 	}
 
-	// TODO: once orgs are implemented, check if the org exists
+	// TODO: once orgs/canvases are implemented, check if the domain exists
```
lucaspin: Should we address this TODO?

pkg/grpc/actions/auth/list_groups.go
```diff
@@ -0,0 +1,49 @@
...
+	// Convert group names to Group objects
+	groups := make([]*pb.Group, len(groupNames))
+	for i, groupName := range groupNames {
+		// Get the role for this group - for now we'll leave it empty as the interface doesn't provide it
+		// TODO: Update authorization service to return group details including roles
+		groups[i] = &pb.Group{
+			Name:       groupName,
+			DomainType: req.DomainType,
+			DomainId:   req.DomainID,
+			Role:       "", // TODO: get actual role from service
```
lucaspin: When are we doing this?

pkg/grpc/actions/auth/types.go
```diff
@@ -0,0 +1,223 @@
...
+		return status.Error(codes.InvalidArgument, "domain type must be specified")
+	}
+
+	return nil
+}
+
+func ConvertDomainType(domainType pb.DomainType) (string, error) {
+	switch domainType {
+	case pb.DomainType_DOMAIN_TYPE_ORGANIZATION:
+		return "org", nil
```
lucaspin: We should use the `DomainOrg` and `DomainCanvas` constants

protos/authorization.proto
```diff
@@ -132,18 +132,50 @@ service Authorization {
     };
   }
 
+  //
+  // Endpoint for creating a group within a canvas
+  // Operation is synchronous and idempotent.
+  //
+  rpc CreateCanvasGroup(CreateCanvasGroupRequest) returns (CreateCanvasGroupResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/canvases/{canvas_id}/groups"
```
lucaspin: Why not use the `canvas_id_or_name` approach?

protos/authorization.proto
```diff
@@ -132,18 +132,50 @@ service Authorization {
...
+    };
+  }
+
+  //
+  // Endpoint for adding a user to a canvas group
+  // Operation is synchronous and idempotent.
+  //
+  rpc AddUserToCanvasGroup(AddUserToCanvasGroupRequest) returns (AddUserToCanvasGroupResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/canvases/{canvas_id}/groups/{group_name}/users"
```
lucaspin: Same comment about canvas_id_or_name

pkg/grpc/actions/auth/create_group.go
forestileao: We can remove this line, since we verify if the user has permission to make actions in the domain in the authorization interceptor.

protos/authorization.proto
forestileao: Updated only for canvas, not org. Since we are going to get the orgId by the domain in the future

pkg/grpc/actions/auth/list_groups.go
forestileao: Implemented

pkg/grpc/actions/auth/get_group_users.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	roles, err := authService.GetGroupRoles(req.DomainID, domainType, req.GroupName)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "failed to get group roles")
+	}
+
+	var role string
+	if len(roles) > 0 {
+		role = roles[0]
```
lucaspin: If multiple roles can be assigned to a group, why do we have just one here?

pedroforesti: Actually, just one role can be assigned to a group. This topic is still on discussion if we should allow more than one role per group

pedroforesti: Regarding this, the interface GetGroupRoles should be GetGroupRole. Updated to avoid misunderstandings

PR #69, opened 2025-07-07
feat: permissions sync

forestileao:
## Changes
- Implement permissions sync at startup to sync permissions for canvases and orgs


pkg/authorization/service_test.go
```diff
@@ -1167,3 +1168,400 @@ func Test__AuthService_GetRoleHierarchy(t *testing.T) {
...
+
+		// Now detect missing permissions
+		missingOrgs, missingCanvases, err := authService.DetectMissingPermissions()
+		require.NoError(t, err)
+
+		// Should not detect any missing permissions for the setup org and canvas
+		orgFound := false
+		canvasFound := false
+
+		for _, missing := range missingOrgs {
```
lucaspin: You can use slices.Contains instead of a loop here

pkg/authorization/service_test.go
```diff
@@ -1167,3 +1168,400 @@ func Test__AuthService_GetRoleHierarchy(t *testing.T) {
...
+
+		// Test that viewer doesn't have write permissions
+		allowed, err = authService.CheckCanvasPermission(userID, canvasID, "stage", "create")
+		require.NoError(t, err)
+		assert.False(t, allowed)
+	})
+}
+
+
+func Test__AuthService_OptimizedSync(t *testing.T) {
```
lucaspin: Does this mean we have an optimized and an non-optimal sync? If so, why?

pkg/models/canvas.go
```diff
@@ -380,3 +380,23 @@ func CreateCanvas(requesterID uuid.UUID, orgID uuid.UUID, name string) (*Canvas,
...
+	err := database.Conn().Model(&Canvas{}).
+		Select("id").
+		Pluck("id", &canvasIDs).Error
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert UUIDs to strings
+	var stringIDs []string
```
lucaspin: We are just de-assigning one thing to another, so no need for this

pkg/models/organization.go
```diff
@@ -125,3 +125,24 @@ func HardDeleteOrganization(id string) error {
...
+		Select("id").
+		Where("deleted_at IS NULL").
+		Pluck("id", &orgIDs).Error
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert UUIDs to strings
+	var stringIDs []string
```
lucaspin: Same comment as above

pkg/authorization/service_test.go
forestileao: This was not a good name for the test. By optimized, I wanted to say something like: "only updates the ones that are missing". Let me change that

PR #71, opened 2025-07-11
Feat/implement approvals by auth

forestileao:
## Changes
Changed the structure of the approval requirement following the new model:
```yml
name: production-deployment
conditions:
  - type: approval
    approval:
      from:
        # Require approval from specific user, no count is required
        - type: user
          id: "12345678-1234-1234-1234-123456789abc"
        
        # Require approval from user by username, no count is required
        - type: user
          name: "john.doe"
          
        # Require approvals from users with specific role
        - type: role
          name: "canvas_admin"
          count: 2
          
        # Require approval from users in specific group
        - type: group
          name: "security-team"
          count: 1
```

Implemented approval requirement by:
- specific user (username or id)
- role
- group


db/migrations/20250710170208_add-username-to-users.up.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+-- Add username column to users table
+ALTER TABLE users
```
lucaspin, line 4: Not so sure about this. Are we doing this just to avoid a JOIN with the account_providers table? What if the user has multiple account providers?

web_src/src/pages/canvas/components/nodes/handle.tsx
```diff
@@ -194,12 +194,14 @@ function TooltipContent({ connections = [], conditions = [] }: { connections?: S
   }
 
   function ApprovalContent({ approval }: { approval: SuperplaneConditionApproval }) {
+    const totalCount = approval.from?.reduce((sum, req) => sum + (req.count || 1), 0) || 0;
```
shiroyasha, line 197: This semi-complex logic appears two times in the code, here and for the settings tab. A shared function that encapsulates this logic would help. e.g. `requiredApprovals(approval)`.

PR #73, opened 2025-07-17
feat: integrations

lucaspin:
This pull request adds integrations, a new way to connect Superplane to external services. The `pkg/integrations` package defines our new `Integration` and `Resource` interfaces, and implements them for Semaphore.

### API changes

- New API endpoints to create canvas/describe/list canvas integrations
- Event source and stage creation endpoints were updated to support integrations

### Event source changes

Event sources can be created in two ways:
- Without referencing an integration - current behavior. If you create an event source this way, you are responsible for injecting the events for it.
- With an integration - SuperPlane will use the integration to take care of provisioning the required resources on the integration side for the webhooks to work.

To support event sources with integrations, a new `PendingEventSourcesWorker` for provisioning the required resources on the integration side for the webhooks to work. In order for that work, we need to have a `BASE_URL` environment variable now, which the worker will use to build the URLs used to provision the webhooks.

### Executor changes

The Semaphore executor was updated to rely on the event source for an integration resource for receiving updates about the execution states. When creating a new stage, if an event source for the targeted resource does not exist, a new internally-scoped event source will be created. If it already exists, the executor will re-use it. Also, to make things easier to query, the `executor_spec` column was removed from the `stages` table, and moved to the `stage_executors` table. The database schema can be seen in [this whimsical](https://whimsical.com/integrations-BCKN3t1ez3YqZzH5Ts5tFd).

### Execution changes

The `reference_id` column was removed from the `stage_executions` table. Now, the resources created for the execution are kept in the `execution_resources` table.


pkg/grpc/actions/integrations/create_integration.go
```diff
@@ -0,0 +1,222 @@
...
+
+		return &models.IntegrationAuth{
+			Token: &models.IntegrationAuthToken{
+				ValueFrom: models.ValueDefinitionFrom{
+					Secret: &models.ValueDefinitionFromSecret{
+						Name: auth.Token.ValueFrom.Secret.Name,
+						Key:  auth.Token.ValueFrom.Secret.Key,
+					},
+				},
+			},
```
pedroforesti, line 147: Is there a possibility to validate this key before creating the integration?

pkg/models/event_source.go
```diff
@@ -23,6 +36,34 @@ func (s *EventSource) UpdateKey(key []byte) error {
...
+}
+
+func (s *EventSource) UpdateState(state string) error {
+	return s.UpdateStateInTransaction(database.Conn(), state)
+}
+
+func (s *EventSource) UpdateStateInTransaction(tx *gorm.DB, state string) error {
+	now := time.Now()
+	s.State = state
+	s.UpdatedAt = &now
```
pedroforesti: I think Gorm already handles that  https://gorm.io/docs/models.html

pkg/models/integration.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type IntegrationOIDC struct {
+	Enabled bool `json:"enabled"`
+}
+
+func CreateIntegration(integration *Integration) (*Integration, error) {
+	now := time.Now()
+	integration.CreatedAt = &now
+	integration.UpdatedAt = &now
```
pedroforesti: Same here

pkg/grpc/actions/integrations/create_integration.go
lucaspin: Oh, good idea. I'll add a new `CheckConnection()` method to the Integration interface for that, and we can call it here. But I'll do that in another PR, if that's OK

pedroforesti: No problem. It will be better

forestileao (approved):

PR #74, opened 2025-07-18
feat: Organization settings UI

forestileao:
## Changes
- Created Organization settings ui based on the design. That includes:
  -  New navigation  dropdown
  - Organization settings sidebar
  - Roles page
  - Members page
  - Groups page (for organizations only)
  - Role / Group Creation Page
- It was also added in the responses from the roles and groups to have a display name and description. That's why we also added new tables to carry this information.
- Added description column to orgs
- Fixed some permissions in the interceptor and services that were not exposed in the grpc


pkg/authentication/authentication.go
```diff
@@ -78,7 +78,7 @@ func (a *Handler) InitializeProviders(providers map[string]ProviderConfig) {
 
 		switch providerName {
 		case "github":
-			gothProviders = append(gothProviders, github.New(config.Key, config.Secret, config.CallbackURL))
+			gothProviders = append(gothProviders, github.New(config.Key, config.Secret, config.CallbackURL, "user"))
```
lucaspin: Can we use a constant for "user"? Same thing for the "github" provider

pkg/authentication/authentication.go
```diff
@@ -281,9 +313,20 @@ func (a *Handler) handleSuccessfulAuth(w http.ResponseWriter, r *http.Request, g
...
 	if r.Header.Get("Accept") == "application/json" {
 		accountProviders, _ := dbUser.GetAccountProviders()
+
+		// Extract primary email and avatar from account providers
+		primaryEmail := ""
+		primaryAvatar := ""
+		if len(accountProviders) > 0 {
+			primaryEmail = accountProviders[0].Email
+			primaryAvatar = accountProviders[0].AvatarURL
+		}
```
lucaspin: This is repeated multiple times, we can put it in a function like `getPrimaryEmail(accountProviders)` and `getPrimaryAvatar(accountProviders)`

pkg/authentication/authentication.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/superplanehq/superplane/pkg/models"
 )
 
+const (
+	ProviderGitHub = "github"
```
lucaspin: Let's keep these constants in the `pkg/models` package. While working on the task to allow the secrets/integrations on the organization level too, I've ran into issues with import cycles by having the domain type scope (org/canvas) in the pkg/authorization

pkg/authentication/authentication.go
```diff
@@ -134,21 +139,41 @@ func (a *Handler) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
...
+			}
+		} else {
+			accountProvider, err = user.GetAccountProvider(ProviderGitHub)
+			if err != nil {
+				log.Errorf("User %s exists but has no GitHub account provider", githubUser.Email)
+				http.Error(w, "Account exists but GitHub provider not connected. Please connect GitHub through the web interface.", http.StatusNotFound)
+				return
+			}
 		}
 	}
```
lucaspin: The branching on this part makes it hard to follow. Can we move this logic to a `findOrCreateAccountProvider` function?

pkg/authentication/authentication.go
```diff
@@ -192,9 +218,16 @@ func (a *Handler) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
...
 
 	accountProviders, _ := dbUser.GetAccountProviders()
```
lucaspin: Why are we not checking the error here?

pkg/authentication/authentication.go
```diff
@@ -192,9 +218,16 @@ func (a *Handler) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
...
+
+	// Extract primary email and avatar from account providers
+	primaryEmail := getPrimaryEmail(accountProviders)
+	primaryAvatar := getPrimaryAvatar(accountProviders)
+
 	authUser := User{
 		ID:               dbUser.ID.String(),
 		Name:             dbUser.Name,
+		Email:            primaryEmail,
```
lucaspin: No need to use a `primaryEmail` variable, just use getPrimaryEmail(accountProviders) directly here

pkg/authentication/authentication.go
```diff
@@ -423,22 +469,32 @@ func (a *Handler) findOrCreateUserAndAccount(gothUser goth.User) (*models.User,
...
+		} else {
+			user.Name = gothUser.Name
+			user.IsActive = true
+			user.Update()
 		}
 	} else {
 		user.Name = gothUser.Name
+		user.IsActive = true
 		user.Update()
 	}
```
lucaspin: Same comment about too much branching making it difficult to follow the code

pkg/authorization/interface.go
```diff
@@ -9,9 +9,12 @@ type PermissionChecker interface {
 // Group management interface
 type GroupManager interface {
 	CreateGroup(domainID string, domainType string, groupName string, role string) error
+	DeleteGroup(domainID string, domainType string, groupName string) error
+	UpdateGroupRole(domainID string, domainType string, groupName string, newRole string) error
 	AddUserToGroup(domainID string, domainType string, userID string, group string) error
 	RemoveUserFromGroup(domainID string, domainType string, userID string, group string) error
 	GetGroupUsers(domainID string, domainType string, group string) ([]string, error)
+	GetGroupMembersCount(domainID string, domainType string, group string) (int, error)
```
lucaspin: Do we really need this? Can't we just use the `GetGroupUsers` for this? The implementation is just calling that anyway

pkg/authorization/service.go
```diff
@@ -107,23 +107,21 @@ func (a *AuthService) checkPermission(userID, domainID, domainType, resource, ac
...
-		DomainCanvas: {RoleCanvasViewer, RoleCanvasAdmin, RoleCanvasOwner},
-	}
-
-	if roles, exists := validRoles[domainType]; exists {
-		if !contains(roles, role) {
-			return fmt.Errorf("invalid role %s for domain type %s", role, domainType)
-		}
-	} else {
+	// Validate domain type
+	if domainType != DomainOrg && domainType != DomainCanvas {
```
lucaspin: Let's move the `DomainOrg` and `DomainCanvas` to the `pkg/models` package too

lucaspin: And this domain type check should be in a shared function that all methods re-use

pkg/authorization/service.go
```diff
@@ -107,23 +107,21 @@ func (a *AuthService) checkPermission(userID, domainID, domainType, resource, ac
...
 	domain := fmt.Sprintf("%s:%s", domainType, domainID)
-	prefixedGroupName := fmt.Sprintf("group:%s", groupName)
 	prefixedRole := fmt.Sprintf("role:%s", role)
 
+	// Check if role exists in this domain using the enforcer
+	if !a.roleExistsInDomain(role, domain) {
+		return fmt.Errorf("invalid role %s for domain type %s", role, domainType)
+	}
+
+	prefixedGroupName := fmt.Sprintf("group:%s", groupName)
```
lucaspin: This domain (`%s:%s`), prefixed group name (`group:%s`) and prefixed role name (`role:%s`) logic should be in a single function that is used by all these methods.

pkg/authorization/service.go
```diff
@@ -1310,3 +1410,69 @@ func (a *AuthService) getDomainTypeFromDomain(domain string) string {
...
+// setupDefaultCanvasRoleMetadata creates metadata for default canvas roles
+func (a *AuthService) setupDefaultCanvasRoleMetadata(canvasID string) error {
+	defaultRoles := []struct {
+		name        string
+		displayName string
+		description string
+	}{
+		{
+			name:        RoleCanvasOwner,
+			displayName: "Owner",
```
lucaspin: We should not use hardcoded values. "Owner" is hardcoded here and in `pkg/models/role_metadata.go`. We should move all these values into the `pkg/models` package.

pkg/grpc/actions/auth/add_user_to_group.go
```diff
@@ -25,10 +26,40 @@ func AddUserToGroup(ctx context.Context, req *GroupUserRequest, authService auth
...
 
-	err = authService.AddUserToGroup(req.DomainID, domainType, req.UserID, req.GroupName)
+	// Handle user identification - either by user_id or user_email
+	var userID string
+	if req.UserID != "" {
+		err := actions.ValidateUUIDs(req.UserID)
+		if err != nil {
+			return status.Error(codes.InvalidArgument, "invalid user ID")
+		}
+		userID = req.UserID
```
lucaspin: We should still validate that this user ID exists here. Also, same comment about excessive branching making this code hard to read

pkg/grpc/actions/auth/assign_role.go
```diff
@@ -36,9 +32,40 @@ func AssignRole(ctx context.Context, req *pb.AssignRoleRequest, authService auth
...
 	}
 
-	// TODO: once users are implemented, check if the user exists
+	err := actions.ValidateUUIDs(req.RoleAssignment.DomainId)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid domain ID")
+	}
+
+	var userID string
+	switch identifier := req.UserIdentifier.(type) {
```
lucaspin: In AddUserToGroup, you have this same logic about finding or creating a new user, and returning its ID, so this should be in a shared function that all endpoints use. Also, let's not use this `oneof` protobuf thingy as discussed.

pkg/grpc/actions/auth/list_groups_test.go
```diff
@@ -42,6 +43,8 @@ func Test_ListGroups(t *testing.T) {
 			assert.Equal(t, pb.DomainType_DOMAIN_TYPE_ORGANIZATION, group.DomainType)
 			assert.Equal(t, orgID, group.DomainId)
 			assert.Contains(t, []string{"org_admin", "org_viewer"}, group.Role)
+			assert.GreaterOrEqual(t, group.MembersCount, int32(0))
+			// CreatedAt and UpdatedAt may be empty if no metadata exists
```
lucaspin: Why would no metadata exist?

forestileao: needed to add the metadata upsert at the start of the test. I was thinking in database inconsistency just in case. But it shouldnt happen.

pkg/authentication/authentication.go
```diff
@@ -548,6 +564,73 @@ func (a *Handler) getGitHubUserInfo(token string) (*GitHubUserInfo, error) {
...
+		return nil, err
+	}
+
+	return user, nil
+}
+
+func (a *Handler) updateUserInfo(user *models.User, name string) *models.User {
+	user.Name = name
+	user.IsActive = true
+	user.Update()
```
lucaspin: We are ignoring the error that Update() could return. We should use `return user.Update()` here instead.

pkg/authorization/service.go
```diff
@@ -470,6 +546,12 @@ func (a *AuthService) SetupCanvasRoles(canvasID string) error {
 		}
 	}
 
+	// Set up metadata for default canvas roles
+	if err := a.setupDefaultCanvasRoleMetadata(canvasID); err != nil {
+		log.Errorf("Error setting up default canvas role metadata: %v", err)
+		// Don't fail the entire setup if metadata fails
```
lucaspin: Why not fail the entire setup if the metadata fails? Isn't the metadata required for this to work properly?

pkg/authorization/service.go
```diff
@@ -81,7 +79,7 @@ func NewAuthService() (*AuthService, error) {
 
 	service := &AuthService{
 		enforcer:              enforcer,
-		db:                    database.Conn(),
+		db:                    adapter.GetDb(),
```
lucaspin: This field doesn't seem to be used

pkg/authorization/service.go
```diff
@@ -98,46 +96,180 @@ func (a *AuthService) CheckOrganizationPermission(userID, orgID, resource, actio
...
 	return a.enforcer.Enforce(prefixedUserID, domain, resource, action)
 }
 
-func (a *AuthService) CreateGroup(domainID string, domainType string, groupName string, role string) error {
-	validRoles := map[string][]string{
-		models.DomainTypeOrganization: {RoleOrgViewer, RoleOrgAdmin, RoleOrgOwner},
-		models.DomainTypeCanvas:       {RoleCanvasViewer, RoleCanvasAdmin, RoleCanvasOwner},
+func (a *AuthService) CreateGroup(domainID string, domainType string, groupName string, role string, displayName string, description string) error {
+	var err error
+	if err := models.ValidateDomainType(domainType); err != nil {
```
lucaspin: Why do we need to validate it here again?

pkg/grpc/actions/auth/common.go
```diff
@@ -6,23 +6,50 @@ import (
...
+	pbRoles "github.com/superplanehq/superplane/pkg/protos/roles"
+	pb "github.com/superplanehq/superplane/pkg/protos/users"
+	pbUsers "github.com/superplanehq/superplane/pkg/protos/users"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-func convertRoleDefinitionToProto(roleDef *authorization.RoleDefinition, authService authorization.Authorization, domainID string) (*pbAuth.Role, error) {
+func convertRoleDefinitionToProto(roleDef *authorization.RoleDefinition, authService authorization.Authorization, domainID string, roleMetadataMap map[string]*models.RoleMetadata) (*pbRoles.Role, error) {
```
lucaspin: authService variable is not used in this function

pkg/grpc/actions/auth/common.go
```diff
@@ -6,23 +6,50 @@ import (
...
+	role := &pbRoles.Role{
+		Metadata: &pbRoles.Role_Metadata{
+			Name:       roleDef.Name,
+			DomainType: actions.DomainTypeToProto(roleDef.DomainType),
+			DomainId:   domainID,
+			CreatedAt:  timestamppb.New(roleMetadataMap[roleDef.Name].CreatedAt),
+			UpdatedAt:  timestamppb.New(roleMetadataMap[roleDef.Name].UpdatedAt),
+		},
+		Spec: &pbRoles.Role_Spec{
+			DisplayName: models.GetRoleDisplayNameWithFallback(roleDef.Name, roleDef.DomainType, domainID, roleMetadata),
```
lucaspin: Now we are ensuring that metadata is always available, we don't need this fallback mechanism, right?

pkg/grpc/actions/auth/delete_group.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		log.Errorf("failed to delete group %s: %v", groupName, err)
+		return nil, status.Error(codes.Internal, "failed to delete group")
+	}
+
+	err = models.DeleteGroupMetadata(groupName, domainType, domainID)
+	if err != nil {
+		log.Errorf("failed to delete group metadata for %s: %v", groupName, err)
+		return nil, status.Error(codes.Internal, "failed to delete group metadata")
+	}
```
lucaspin: This should all happen inside of a transaction

pkg/grpc/actions/auth/delete_role.go
```diff
@@ -5,41 +5,37 @@ import (
...
 	}
 
-	err = authService.DeleteCustomRole(req.DomainId, domainType, req.RoleName)
+	err = models.DeleteRoleMetadata(roleName, domainType, domainID)
 	if err != nil {
-		log.Errorf("failed to delete role %s: %v", req.RoleName, err)
-		return nil, status.Error(codes.Internal, err.Error())
+		log.Errorf("failed to delete role metadata for %s: %v", roleName, err)
+		return nil, status.Error(codes.Internal, "failed to delete role metadata")
 	}
```
lucaspin: Same comment about transaction

pkg/grpc/actions/auth/describe_group.go
```diff
@@ -0,0 +1,64 @@
...
+
+	role, err := authService.GetGroupRole(domainID, domainType, groupName)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, "group not found")
+	}
+
+	groupMetadata, err := models.FindGroupMetadata(groupName, domainType, domainID)
+	var displayName, description string
+	var createdAt, updatedAt *timestamppb.Timestamp
+	if err == nil {
```
lucaspin, line 28: If metadata is not found, we should fail

pkg/grpc/actions/auth/describe_role.go
```diff
@@ -4,36 +4,33 @@ import (
...
-		return nil, status.Error(codes.InvalidArgument, "invalid role specified")
+	roleNames := []string{roleDefinition.Name}
+	if roleDefinition.InheritsFrom != nil {
+		roleNames = append(roleNames, roleDefinition.InheritsFrom.Name)
 	}
 
-	roleDefinition, err := authService.GetRoleDefinition(req.Role, domainType, req.DomainId)
+	roleMetadataMap, err := models.FindRoleMetadataByNames(roleNames, domainType, domainID)
 	if err != nil {
-		return nil, status.Error(codes.NotFound, "role not found")
```
lucaspin, line 33: If metadata is not found, we should fail

pkg/grpc/actions/auth/get_user_roles.go
```diff
@@ -34,19 +32,32 @@ func GetUserRoles(ctx context.Context, req *pb.GetUserRolesRequest, authService
...
+	for i, role := range roles {
+		roleNames[i] = role.Name
+		if role.InheritsFrom != nil {
+			roleNames = append(roleNames, role.InheritsFrom.Name)
+		}
+	}
+
+	roleMetadataMap, err := models.FindRoleMetadataByNames(roleNames, domainType, domainID)
+	if err != nil {
+		roleMetadataMap = make(map[string]*models.RoleMetadata)
```
lucaspin: If metadata is not found, we should fail

pkg/grpc/actions/auth/list_group_users.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	role, err := authService.GetGroupRole(domainID, domainType, groupName)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "failed to get group roles")
+	}
+
+	roleMetadataMap, err := models.FindRoleMetadataByNames([]string{role}, domainType, domainID)
+	if err != nil {
+		roleMetadataMap = make(map[string]*models.RoleMetadata)
```
lucaspin: If metadata is not found, we should fail

pkg/grpc/actions/auth/list_group_users.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		user, err := convertUserToProto(userID, []*pbUsers.UserRoleAssignment{roleAssignment})
+		if err != nil {
+			continue
+		}
+		users = append(users, user)
+	}
+
+	groupMetadata, err := models.FindGroupMetadata(groupName, domainType, domainID)
```
lucaspin, line 56: If metadata is not found, we should fail

pkg/grpc/actions/auth/list_groups.go
```diff
@@ -5,47 +5,64 @@ import (
...
 
+		groupUsers, err := authService.GetGroupUsers(domainID, domainType, groupName)
+		if err != nil {
+			return nil, status.Error(codes.Internal, "failed to get group members count")
+		}
+
+		groupMetadata, err := models.FindGroupMetadata(groupName, domainType, domainID)
+		var createdAt, updatedAt *timestamppb.Timestamp
+		var displayName, description string
+		if err == nil {
```
lucaspin, line 36: If metadata is not found, we should fail

pkg/grpc/actions/auth/list_roles.go
```diff
@@ -4,34 +4,34 @@ import (
...
+	for i, roleDef := range roleDefinitions {
+		roleNames[i] = roleDef.Name
+		if roleDef.InheritsFrom != nil {
+			roleNames = append(roleNames, roleDef.InheritsFrom.Name)
+		}
 	}
 
-	roleDefinitions, err := authService.GetAllRoleDefinitions(domainType, req.DomainId)
+	roleMetadataMap, err := models.FindRoleMetadataByNames(roleNames, domainType, domainID)
 	if err != nil {
```
lucaspin, line 28: If metadata is not found, we should fail

pkg/grpc/actions/auth/update_group.go
```diff
@@ -0,0 +1,90 @@
...
+
+	var displayName string
+	var description string
+	groupModelMetadata, err := models.FindGroupMetadata(groupName, domainType, domainID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "failed to get group metadata")
+	}
+
+	if groupSpec != nil && (groupSpec.DisplayName != "" || groupSpec.Description != "") {
+		displayName = groupSpec.DisplayName
```
lucaspin, line 34: Shouldn't we just use what's in the metadata always?

pkg/grpc/actions/auth/delete_group.go
forestileao: actually we dont eeven need this DeleteGroupMetadata, since we already have it internally in authService.DeleteGroup. and it is inside a transaction

pkg/grpc/actions/auth/update_group.go
forestileao: I didn't get your question. This case is for updating the display name and description only if the user specifies it.

pkg/grpc/actions/auth/describe_role.go
forestileao: just a note: we are already going to get a Internal server error if the role metadata does not exist. What we expect

	roleMetadata := roleMetadataMap[roleDef.Name]
	DisplayName: roleMetadata.DisplayName,

pkg/grpc/actions/auth/update_group.go
lucaspin: Right, but these if/else statements are after we have already updated the group and on line 38, we do a `models.FindGroupMetadata`, can't we just use the result of that here, like:

```golang
groupModelMetadata, err := models.FindGroupMetadata(groupName, domainType, domainID)
if err != nil {
    return nil, status.Error(codes.Internal, "failed to get group metadata")
}

updatedRole := groupSpec.Role
if updatedRole == "" {
    updatedRole = currentRole
}

groupUsers, err := authService.GetGroupUsers(domainID, domainType, groupName)
if err != nil {
    return nil, status.Error(codes.Internal, "failed to get group members count")
}

group := &pb.Group{
    Metadata: &pb.Group_Metadata{
        Name:       groupName,
        DomainType: actions.DomainTypeToProto(domainType),
        DomainId:   domainID,
        CreatedAt:  timestamppb.Now(),
        UpdatedAt:  timestamppb.Now(),
    },
    Spec: &pb.Group_Spec{
         Role:        updatedRole,
         DisplayName: groupModelMetadata.DisplayName,
         Description: groupModelMetadata.Description,
    },
    Status: &pb.Group_Status{
        MembersCount: int32(len(groupUsers)),
    },
}

return &pb.UpdateGroupResponse{
    Group: group,
}, nil
```

pkg/grpc/actions/auth/delete_group.go
forestileao: Also, inside the delete group transaction we already deletes the user using  _, err = e.RemoveFilteredGroupingPolicy(1, prefixedRoleName, domain) and  _, err = e.RemoveFilteredGroupingPolicy(1, prefixedGroupName, domain)

So I just removed it from the action

pkg/grpc/actions/auth/update_group.go
forestileao: My bad, I also found an errror in the logic. This thing should be before the update

pkg/grpc/actions/auth/describe_role.go
forestileao: Added the error after the reviewing this line as well

PR #75, opened 2025-07-22
feat: secrets and integrations on organization level

lucaspin:
### API changes

- Move the secrets API into its own secrets.proto file
- Move the integrations API into its own integrations.proto file
- Rename superplane.proto to canvases.proto
- Update interceptor to handle endpoints that support multiple domain types

### Database changes

- Add domain type and domain ID fields to secrets table

### Notes

Most of the changes in this pull request are due to moving the domain type constants into the `pkg/models` package. The authorization API endpoints will still need to be updated to follow the new pattern, but that will be done in another PR.


pkg/grpc/superplane_service.go
```diff
@@ -9,127 +9,93 @@ import (
...
 	return groups.DescribeConnectionGroup(ctx, req)
 }
 
-func (s *SuperplaneService) ListConnectionGroups(ctx context.Context, req *pb.ListConnectionGroupsRequest) (*pb.ListConnectionGroupsResponse, error) {
+func (s *CanvasService) ListConnectionGroups(ctx context.Context, req *pb.ListConnectionGroupsRequest) (*pb.ListConnectionGroupsResponse, error) {
 	return groups.ListConnectionGroups(ctx, req)
 }
 
-func (s *SuperplaneService) ListConnectionGroupFieldSets(ctx context.Context, req *pb.ListConnectionGroupFieldSetsRequest) (*pb.ListConnectionGroupFieldSetsResponse, error) {
+func (s *CanvasService) ListConnectionGroupFieldSets(ctx context.Context, req *pb.ListConnectionGroupFieldSetsRequest) (*pb.ListConnectionGroupFieldSetsResponse, error) {
```
forestileao: why not create a file just for canvas_service.go?

pkg/models/organization.go
```diff
@@ -10,6 +10,11 @@ import (
 	"gorm.io/gorm/clause"
 )
 
+const (
+	DomainTypeOrganization = "org"
+	DomainTypeCanvas       = "canvas"
+)
+
```
forestileao: this will be highly shared, we could put in a `models/constants.go`

pkg/grpc/superplane_service.go
lucaspin: Renamed this file to canvas_service.go

pkg/models/organization.go
lucaspin: Done

PR #79, opened 2025-07-25
refactor: add SetupWebhook() method to integration interface

lucaspin:
The code for setting up a integration webhook is currently in the PendingEventSourcesWorker, which makes it required to update that worker when adding a new integration type. That is not what we want. This pull request moves that logic to the Integration interface, behind a new SetupWebhook() method. 

Additionally, I moved the code for the Semaphore integration into the `pkg/integrations/semaphore` package, and updated the logic for instantiating a new instance of an integration. Now, we use a integration factory that allows new integration types to register themselves into it.


pkg/integrations/semaphore/semaphore.go
```diff
@@ -1,32 +1,151 @@
...
+	// Secret does not exist, create it.
+	//
+	secret, err = s.Create(ResourceTypeSecret, &Secret{
+		Metadata: SecretMetadata{
+			Name: name,
+		},
+		Data: SecretSpecData{
+			EnvVars: []SecretSpecDataEnvVar{
+				{
+					Name:  "WEBHOOK_SECRET",
```
forestileao, line 90: We could store this in a constant

PR #80, opened 2025-07-28
refactor: new integration / executor should not require API / core code changes

lucaspin:
Adding a new integration or executor requires updating the API and some logic in the workers for the integration/executor to work properly. That shouldn't be the case. Ideally, adding a new integration/executor should be just about adding a new implementation for the interface you want, and registering it. The changes in this pull request allow for that to happen.

#### Key Changes

- Introduced pkg/registry package to centrally manage integrations and executors using factory patterns
- Moved integration-specific executors from pkg/executors/ to pkg/integrations/ for better organization. Executors that don't require an integration are still placed in pkg/executors/.
- Updated the API to use generic types for the execution spec. That way, the executor is the one responsible for what the spec looks like, and when adding a new executor, we don't need to update the API.
- Moved executor spec validation logic into the executor interface
- Moved the integration webhook setup logic into the integration interface


pkg/workers/pending_executions_worker.go
```diff
@@ -167,8 +163,44 @@ func (w *PendingExecutionsWorker) handleSyncResource(logger *log.Entry, response
...
+	if err != nil {
+		return err
+	}
+
+	integrationExecutor, err := w.Registry.NewIntegrationExecutor(integration, resource)
+	if err != nil {
+		return err
+	}
+
+	token, err := w.JwtSigner.Generate(execution.ID.String(), 24*time.Hour)
```
forestileao: We can store this time into a constant. Or add some configuration for this kind of token

lucaspin: Moved into a constant for now

forestileao (approved):

PR #81, opened 2025-07-30
refactor: new EventHandler and ResourceManager interfaces

lucaspin:
For better separation of concerns, this pull request splits the monolithic Integration interface into two focused interfaces:
- EventHandler: handles webhook data conversion and event processing. Since the event handler does not need to make any interactions with the external integration system for this, and the logic is the same for all integrations of the same type, the same event handler implementation instance is shared.
- ResourceManager: manages integration resources. This one requires a specific connection (URL + token) to an integration instance, so it cannot be shared, and must be initialized every time.

Also, now that the EventHandler interface can provide custom event handling logic per integration, we can add a new type column to the events table, and update the HTTP server to defer the webhook handling to the integration itself.


pkg/models/events.go
```diff
@@ -0,0 +1,98 @@
...
+
+	//
+	// Include the missing field, if any.
+	//
+	e.Missing = []string{}
+	if len(missingConnections) > 0 {
+		for _, connection := range missingConnections {
+			e.Missing = append(e.Missing, connection.SourceName)
+		}
+	}
```
forestileao: Here we have a duplicated verification for missing connections using `len(missingConnections)`

lucaspin: Good catch. Fixed.

forestileao (approved):

PR #82, opened 2025-07-31
feat: OIDC support for requests from integrations

lucaspin:
This PR implements OpenID Connect (OIDC) authentication support for integrations, enabling secure authentication using OIDC ID tokens from external services like Semaphore CI. The implementation adds a new OIDCVerifier interface and removes the previous OIDC configuration from the integration model.

Key Changes

- New crypto.OIDCVerifier with caching support for multiple issuers and OIDCVerifier interface in integrations package, using [coreos/go-oidc](https://github.com/coreos/go-oidc) library
- Implemented SemaphoreOIDCVerifier with proper token validation including workflow ID and project ID claims verification
- Removed unused oidc column from integrations table and updated migration files
- Updated public API server to accept and use OIDC verifier for authenticated requests


pkg/crypto/oidc.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	provider, err := oidc.NewProvider(ctx, issuer)
+	if err != nil {
+		return nil, err
+	}
+
+	verifier = provider.VerifierContext(ctx, &oidc.Config{ClientID: audience})
+	v.verifiers[issuer] = verifier
+	return verifier.Verify(ctx, token)
```
forestileao, line 40: If this is a shared map that will be handled by a lot of asyncronous requests. It would be nice to add a mutex here

pkg/workers/pending_executions_worker_test.go
```diff
@@ -209,14 +209,10 @@ func assertParameters(t *testing.T, req *semaphore.CreateWorkflowRequest, execut
 		all[k] = v
 	}
 
-	assert.Len(t, req.Parameters, len(all)+1)
+	assert.Len(t, req.Parameters, len(all))
```
forestileao, line 212: why this change?

pkg/crypto/oidc.go
lucaspin: Oh, yes. Good catch. Done

pkg/workers/pending_executions_worker_test.go
lucaspin: The `SUPERPLANE_STAGE_EXECUTION_TOKEN` is no longer added to the parameters now

forestileao (approved):

PR #83, opened 2025-08-01
feat: filters on event sources

lucaspin:
Adds support for event type filters on event sources, allowing fine-grained control over which events are processed.

### Key Changes

- New event_types JSONB column to event_sources table to store event type configurations
- Updated API + generated clients for event sources
- Updated event processing logic to validate events against configured event type filters before processing the connections for the source.

### Other changes

Also updated the expression processor to use the `$.*` syntax for accessing fields in the event payload.

### DIFF
--- a/pkg/grpc/actions/event_sources/create_event_source.go
+++ b/pkg/grpc/actions/event_sources/create_event_source.go
@@ -110,11 +110,7 @@ func CreateEventSource(ctx context.Context, encryptor crypto.Encryptor, registry
 }
 
 func validateEventTypes(spec *pb.EventSource_Spec) ([]models.EventType, error) {
-	if spec == nil {
-		return []models.EventType{}, nil
-	}
-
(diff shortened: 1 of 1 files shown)


pkg/grpc/actions/event_sources/create_event_source.go
```diff
@@ -110,8 +109,56 @@ func CreateEventSource(ctx context.Context, encryptor crypto.Encryptor, registry
...
 }
 
+func validateEventTypes(spec *pb.EventSource_Spec) ([]models.EventType, error) {
+	if spec == nil {
+		return []models.EventType{}, nil
+	}
+
+	if spec.Events == nil {
+		return []models.EventType{}, nil
+	}
```
pedroforesti:
```suggestion
	if spec == nil || spec.Events == nil {
		return []models.EventType{}, nil
	}
```

PR #86, opened 2025-08-04
feat: add canvas management UI pages

forestileao:
### Changes
In this PR, Canvas Settings pages were added that can handle:
- Secrets
- Integrations
- General Settings
- Deletion

Also, it was added edit mode for nodes, including validation of fields. All these things are integrated using the generated SDK + React Query hooks.

### Fixes
Some states were being. updated only in the node.data directly instead of using the hooks. Generating some inconsistencies. Now this is fixed.


web_src/src/components/SettingsPage/components/CanvasSecrets.tsx
```diff
@@ -0,0 +1,337 @@
+import { useState, useEffect } from 'react'
+import { Heading } from '../../Heading/heading'
+import { Text } from '../../Text/text'
+import { Button } from '../../Button/button'
+import { MaterialSymbol } from '../../MaterialSymbol/material-symbol'
+import { Input } from '../../Input/input'
+import { useSecrets, useCreateSecret, useDeleteSecret, useUpdateSecret, useSecret } from '../../../pages/canvas/hooks/useSecrets'
+
+// Utility function to format relative time
+const formatRelativeTime = (dateString: string | undefined): string => {
```
lucaspin: We can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat for this

web_src/src/pages/canvas/index.tsx
```diff
@@ -149,29 +175,87 @@ export function Canvas() {
...
   }
 
+  // Handle adding nodes using the modular system
+  const handleAddNodeByType = (nodeType: string, executorType?: string, eventSourceType?: string) => {
+    try {
+      // Map string types to NodeType enum
+      const nodeTypeMap: Record<string, NodeType> = {
+        'stage': 'stage',
+        'event_source': 'event_source',
+        'eventSource': 'event_source', // Support both naming conventions
```
lucaspin: Why do we need to support both "naming conventions"?

web_src/src/pages/canvas/index.tsx
```diff
@@ -149,29 +175,87 @@ export function Canvas() {
...
+        'connectionGroup': 'connection_group' // Support both naming conventions
+      };
+
+      const mappedNodeType = nodeTypeMap[nodeType];
+      if (!mappedNodeType) {
+        console.error(`Unknown node type: ${nodeType}`);
+        return;
+      }
+
+      // Handle stages with executor types
```
lucaspin: These conditionals are hard to follow. Some switches and separate functions and maps could make this easier to read

PR #89, opened 2025-08-06
feat: github integration

lucaspin:
New GitHub Integration added in `pkg/integrations/github` package, allowing GitHub integrations to be created using a fine-grained personal access token. With that, you can:
- Create event sources for receiving [GitHub events](https://docs.github.com/en/webhooks/webhook-events-and-payloads) for specific repositories.
- Execute GitHub Actions workflows via [workflow dispatch events](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) during stage executions.
- Use [OIDC ID tokens](https://docs.github.com/en/actions/concepts/security/openid-connect) for pushing stage outputs from workflow runs.

### Notes

Since the GitHub API doesn't expose the ID for the workflow that was just triggered, we need to poll to find the triggered workflow run. Additionally, since the GitHub API doesn't expose the inputs passed to the workflow, we need to rely on the `superplane_execution_id` input being used in the `run-name` field in the workflow YAML definition, to dynamically set the workflow run name in order for us to reliably identify the workflow run created. See [this discussion](https://github.com/orgs/community/discussions/9752) for more details.


web_src/src/pages/canvas/utils/nodeFactories.ts
```diff
@@ -39,6 +39,25 @@ export const createEmptyStage = ({ canvasId, name = 'New Stage', executorType }:
...
+          },
+          spec: {
+            workflow: '.github/workflows/task.yaml',
+            ref: 'main',
+            inputs: {
+              VERSION: '${{ inputs.VERSION }}',
+              ENVIRONMENT: '${{ inputs.ENVIRONMENT }}'
+            }
+          }
+        };
```
pedroforesti, line 60: We are planning to avoid adding too much boilerplate. But since I am working on the validation and UI/UX for these forms I will handle that in my PR

forestileao (approved):

PR #90, opened 2025-08-08
fix: scope sources / stages / connection groups / events queries to canvas

lucaspin:


pkg/models/connection_group.go
```diff
@@ -266,13 +293,36 @@ func (c *Canvas) UpdateConnectionGroup(id, name, description, requesterID string
 	})
 }
 
-func FindConnectionGroupByID(id uuid.UUID) (*ConnectionGroup, error) {
-	return FindConnectionGroupByIDInTransaction(database.Conn(), id)
+// NOTE: we are not querying scoped by canvas here,
+// so this should be used only in the workers.
+func FindConnectionGroup(id string) (*ConnectionGroup, error) {
```
forestileao: Not a requirement, but it would be good to rename it to FindUnscoperConnectionGroup

pkg/models/stage.go
```diff
@@ -192,15 +192,49 @@ type ApprovalCondition struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return stages, nil
+}
+
+// NOTE: we are not querying scoped by canvas here,
+// so this should be used only in the workers.
+func FindStage(id string) (*Stage, error) {
```
forestileao: Same here

forestileao (approved):

PR #99, opened 2025-08-13
refactor: relationship between users and organizations

lucaspin:
### Changes

- The relationship between users and organization was updated, with a new account model being used. Users are now a concrete association between an account and an organization.
- Currently, we are using an `is_active` flag on the users table to determine if a user is part of the organization or it it has been only invited. That is confusing and prone to errors. Now, we use a `organization_invitations` table to keep track of the invitations, and once an account that has been invited logs in, we add them to the organizations that have invited it. This change also makes it easier for us to extend things in the future, and make invitations expire, or support invitations based on tokens that can be used by anyone.
- Removed /auth/me, /auth/token/exchange, and /auth/{provider}/disconnect endpoints. the CLI will be updated to use user API tokens in a following pull request.
- The API routes were updated to expect an account cookie and an `x-organization-id` header. From those values, we can check if the user is part of the organization. Once we have support for user API tokens, we will update this middleware to accept a user API token (scoped to the organization) to authenticate as well.
- New /account and GET /organizations endpoints to fetch the current authenticated account and organizations associated with it.
- Since organization creation requires account-based authentication (no organization context), a new POST /organizations endpoint was added to allow authenticated accounts to create organizations.

### DIFF
--- a/pkg/grpc/actions/stages/create_stage.go
+++ b/pkg/grpc/actions/stages/create_stage.go
@@ -26,8 +26,8 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-func CreateStage(ctx context.Context, encryptor crypto.Encryptor, registry *registry.Registry, canvasID string, stage *pb.Stage) (*pb.CreateStageResponse, error) {
-	canvas, err := models.FindUnscopedCanvasByID(canvasID)
+func CreateStage(ctx context.Context, encryptor crypto.Encryptor, registry *registry.Registry, orgID, canvasID string, stage *pb.Stage) (*pb.CreateStageResponse, error) {
+	canvas, err := models.FindCanvasByID(canvasID, uuid.MustParse(orgID))
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, "canvas not found")
 	}
--- a/pkg/grpc/actions/event_sources/create_event_source.go
+++ b/pkg/grpc/actions/event_sources/create_event_source.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	uuid "github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/superplanehq/superplane/pkg/builders"
 	"github.com/superplanehq/superplane/pkg/crypto"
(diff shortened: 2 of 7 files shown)


pkg/authentication/authentication.go
```diff
@@ -92,184 +63,130 @@ func (a *Handler) InitializeProviders(providers map[string]ProviderConfig) {
 	}
 }
 
-// RegisterRoutes adds authentication routes to the router
 func (a *Handler) RegisterRoutes(router *mux.Router) {
-	router.HandleFunc("/auth/me", a.handleMe).Methods("GET")
```
forestileao, line 97: Is it a good idea to remove this route completely? since this kind of route is a pattern I believe. and we were using it in the frontend

pkg/grpc/actions/stages/create_stage.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func CreateStage(ctx context.Context, encryptor crypto.Encryptor, registry *registry.Registry, canvasID string, stage *pb.Stage) (*pb.CreateStageResponse, error) {
-	canvas, err := models.FindCanvasByID(canvasID)
+	canvas, err := models.FindUnscopedCanvasByID(canvasID)
```
forestileao: can't we use scoped here? By the organization id injected in the context?

web_src/src/App.tsx
```diff
@@ -24,26 +27,39 @@ const queryClient = new QueryClient({
...
-        </Routes>
-      </BrowserRouter>
+      <AccountProvider>
+        <BrowserRouter basename={BASE_PATH}>
+          <Routes>
+ 
+            {/* Organization-scoped protected routes */}
+            <Route path=":organizationId" element={withAuthAndNavigation(HomePage)} />
+            <Route path=":organizationId/canvas/:canvasId" element={withAuthOnly(Canvas)} />
+            <Route path=":organizationId/settings/*" element={withAuthAndNavigation(OrganizationSettings)} />
```
forestileao, line 55: Did you check all references on links, useNavigate, useParams hooks? I will probably need to update some in my end, but that is no problem

web_src/src/components/SettingsPage/components/CanvasDelete.tsx
```diff
@@ -37,7 +37,7 @@ export function CanvasDelete({ canvasId, organizationId }: CanvasDeleteProps) {
   const handleConfirmDelete = async () => {
     if (confirmationText === 'DELETE') {
       await deleteCanvasMutation.mutateAsync({ canvasId })
-      navigate(`/organization/${organizationId}`)
+      navigate(`/`)
```
forestileao: why not `/${organizationId}`

pkg/authentication/authentication.go
lucaspin: This route was replaced by the /account one, and the UI was properly updated to use the new /account one. Check the AccountContext context.

web_src/src/App.tsx
lucaspin: Yes

web_src/src/components/SettingsPage/components/CanvasDelete.tsx
lucaspin: That's better, yes. Updated.

pkg/grpc/actions/stages/create_stage.go
lucaspin: Yes. Updated it

PR #111, opened 2025-08-14
refactor: UI/UX for event sources

forestileao:
## Changes

This PR contains changes specific related to Event sources.

### Frontend Changes
- Auto complete for event source names
- New Event Source sidebar with view headers and payload
- Event Source events poller until events are not pending
- New websocket event handler
- Webhook key regeneration
- Event source filters

### Backend Changes
- New general `event_created` event, both for webhook and rabbitmq
- New ListEventsService


pkg/grpc/actions/events/list_events.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, status.Error(codes.InvalidArgument, "invalid canvas ID")
+	}
+
+	_, err = models.FindCanvasByID(canvasID, orgUUID)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, status.Error(codes.InvalidArgument, "canvas not found")
+		}
+		return nil, err
+	}
```
lucaspin: We don't need to do all of this again here, right?

All this is done in the interceptor

pkg/grpc/actions/events/list_events.go
```diff
@@ -0,0 +1,150 @@
...
+
+	sourceTypeStr := EventSourceTypeToString(sourceType)
```
lucaspin: No need for an additional variable, just use EventSourceTypeToString(sourceType) in the function call

pkg/grpc/actions/events/list_events.go
```diff
@@ -0,0 +1,150 @@
...
+		serialized, err := serializeEvent(event)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, serialized)
+	}
+	return out, nil
+}
+
+func serializeEvent(in models.Event) (*pb.Event, error) {
```
lucaspin, line 102: We are calling structpb.NewStruct() 4 times for no reason, and we are ignoring errors. This function can be rewritten as:

```go
func serializeEvent(in models.Event) (*pb.Event, error) {
	event := &pb.Event{
		Id:         in.ID.String(),
		SourceId:   in.SourceID.String(),
		SourceName: in.SourceName,
		SourceType: StringToEventSourceType(in.SourceType),
		Type:       in.Type,
		State:      StringToEventStateProto(in.State),
		ReceivedAt: timestamppb.New(*in.ReceivedAt),
	}

	if len(in.Raw) > 0 {
		data, err := in.GetData()
		if err != nil {
			return nil, err
		}

		event.Raw = structpb.NewStruct(data)
	}

	if len(in.Headers) > 0 {
		headers, err := in.GetHeaders()
		if err != nil {
			return nil, err
		}

		event.Headers = structpb.NewStruct(headers)
	}

	return event, nil
}
```

pkg/grpc/actions/events/list_events_test.go
```diff
@@ -0,0 +1,111 @@
...
+		assert.Equal(t, r.Source.Name, e.SourceName)
+		assert.Equal(t, protos.EventSourceType_EVENT_SOURCE_TYPE_EVENT_SOURCE, e.SourceType)
+		assert.Equal(t, "webhook", e.Type)
+		assert.Equal(t, protos.Event_STATE_PENDING, e.State)
+		assert.NotNil(t, e.ReceivedAt)
+		assert.NotNil(t, e.Raw)
+		assert.NotNil(t, e.Headers)
+	})
+
+	t.Run("filter by source type", func(t *testing.T) {
```
lucaspin, line 63: we are testing that we can filter by source type, but we only have one event source, no?

lucaspin: Ah, nevermind, forgot about the stage

pkg/models/event.go
```diff
@@ -267,6 +267,38 @@ func FindLastEventBySourceID(sourceID uuid.UUID) (map[string]any, error) {
 	return m, nil
 }
 
+func ListEventsByCanvasID(canvasID uuid.UUID, sourceType string, sourceIDStr string) ([]Event, error) {
+	var events []Event
+	
+	query := database.Conn().Table("events")
+	
+	query = query.Joins("LEFT JOIN event_sources ON events.source_id = event_sources.id AND events.source_type = ?", SourceTypeEventSource).
```
lucaspin: Why do we need all these joins? All the information we need to filter events are already in the events table

forestileao: The joins are actually necessary because the events table doesn't contain
  canvas_id - it's stored in the related event_sources, stages, and
  connection_groups tables. The query needs to join with these tables to filter
  events by canvas_id.

lucaspin: Then let's add a canvas_id to the events table too

PR #113, opened 2025-08-14
feat: include inputs in stage completion event

lucaspin:
Currently, there's no way to use the inputs from one stage in another stage in the chain, unless you push that input as an output in the execution. Here, we add the inputs in the stage completion event, so the next stages in the chain can also use them.


docs/simulating-source-event.md
```diff
@@ -7,7 +7,7 @@ export EVENT="{\"ref\":\"v1.0\",\"ref_type\":\"tag\"}"
 export SIGNATURE=$(echo -n "$EVENT" | openssl dgst -sha256 -hmac "$SOURCE_KEY" | awk '{print $2}')
 
 curl -X POST \
-  -H "X-Hub-Signature-256: sha256=$SIGNATURE" \
+  -H "X-Signature-256: sha256=$SIGNATURE" \
```
forestileao: Actually this whole markdown is wrong. Since this speficicaly for simulating for github.
Not specifically for testing webhooks. That's why we had "X-Hub-Signature-256".
This markdown should be: "How to test your webhook event source"

forestileao (approved):

PR #114, opened 2025-08-14
fix: removing user from organization

lucaspin:
Now that adding a user to an organization is no longer just about casbin rules, we have an issue when removing users from the organization. The issue happens because we are currently using the `PATCH /api/v1/roles/remove` endpoint for this, which does remove the role from the user, but leaves the user record in the database.

This pull request removes that endpoint, and replaces it with two:
- DELETE /api/v1/users/{user_id} endpoint for removing users from the org, which removes all the casbin rules for the organization and any canvases the user might be in, alongside with the user record
- DELETE /api/v1/canvases/{id}/users/{user_id} endpoint for removing users from a canvas

### Other changes

- Since we have a new `DELETE /api/v1/canvases/{id}/users/{user_id}`, having a matching `POST /api/v1/canvases/{id}/users` endpoint to add users to the canvas makes a bit more sense.
- Describing a role was using the `GET /api/v1/roles/describe`. It makes more sense to use `GET /api/v1/roles/describe/{role_name}` instead
- POST /api/v1/roles/{roleName}/users - instead of POST /api/v1/roles/assign - for assigning a role to a user


pkg/grpc/actions/canvases/remove_user.go
```diff
@@ -0,0 +1,35 @@
...
+	//
+	// TODO: this should be in transaction
+	//
+	for _, role := range roles {
+		err = authService.RemoveRole(user.ID.String(), role.Name, canvasID, models.DomainTypeCanvas)
+		if err != nil {
+			return nil, status.Error(codes.Internal, "error removing user")
+		}
+	}
+
```
forestileao, line 33: The get user roles from canvas also gets the implicit ones. But since it ignores whether the user does not have the role, I think it would be okay. Maybe if we add a test for trying to remove an admin/owner user

pkg/grpc/actions/organizations/remove_user.go
```diff
@@ -0,0 +1,68 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func RemoveUser(ctx context.Context, authService authorization.Authorization, orgID, userID string) (*pb.RemoveUserResponse, error) {
+	user, err := models.FindUserByID(orgID, userID)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, "user not found")
+	}
+
```
forestileao, line 19: also add a TODO that this thing should be in a transaction

Issue #106, opened 2025-08-14
"Name already used" while creating a new canvas with a duplicate name


shiroyasha:
Here is how to reproduce:

1. Created a new canvas, tried out various things, deleted it
2. Tried to create a new canvas
3. No error on the screen, console says 400 name already used

<img width="2764" height="1508" alt="Image" src="https://github.com/user-attachments/assets/a437211a-c46c-40a5-a505-bc2522639383" />

AleksandarCole:
@forestileao are we actually deleting canvases or is that in that same list as deleting components (TBD)?

forestileao:
@AleksandarCole deleting canvases is currently a soft deletion. Until we have a "Garbage Collector" with grace period to in fact delete forever. This fix is already in our list of TODOs

AleksandarCole:
All clear. Thank you!

PR #126, opened 2025-08-19
fix: conflicting canvas name after deletion

forestileao:
## Issue https://github.com/superplanehq/superplane/issues/106

[fix: conflicting canvas name after deletion](https://github.com/superplanehq/superplane/commit/d91fe69bcec46d3c411b09a8f79ee0da6a23dc97)

### DIFF
--- a/pkg/models/canvas.go
+++ b/pkg/models/canvas.go
@@ -122,6 +122,16 @@ func (c *Canvas) CreateStageInTransaction(
 }
 
 func (c *Canvas) DeleteInTransaction(tx *gorm.DB) error {
+	deletedName := fmt.Sprintf("%s-deleted-%d", c.Name, time.Now().Unix())
+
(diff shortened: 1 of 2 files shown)


pkg/models/canvas.go
```diff
@@ -122,6 +122,16 @@ func (c *Canvas) CreateStageInTransaction(
...
+		Where("id = ?", c.ID).
+		Update("name", deletedName).
+		Error
+	if err != nil {
+		return err
+	}
+
 	return tx.
 		Where("id = ?", c.ID).
 		Delete(&Canvas{}).
```
lucaspin: We don't need 2 queries for this. We can update the deleted_at timestamp on the first one, and remove this one

forestileao: Addressed!

PR #173, opened 2025-08-21
feat: intuitive run names

forestileao:
### Changes
- Implement Intuitive Run Names in Backend and Frontend
- Now, there is a new option under executor called "label". This label can be configured to  manage how the executions text will be displayed in UI:

image
image
image

### DIFF
--- a/db/migrations/20250821200205_add-executor-label-fields.down.sql
+++ b/db/migrations/20250821200205_add-executor-label-fields.down.sql
@@ -1,2 +1,0 @@
-ALTER TABLE stages DROP COLUMN executor_label;
-ALTER TABLE stage_events DROP COLUMN executor_label;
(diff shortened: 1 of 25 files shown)


db/migrations/20250821200205_add-executor-label-fields.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE stages DROP COLUMN executor_label;
```
lucaspin: We don't use these, so let's just not have it for this too

forestileao: My bad. Removed

db/migrations/20250821200205_add-executor-label-fields.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE stages ADD COLUMN executor_label text;
+ALTER TABLE stage_events ADD COLUMN executor_label text;
```
lucaspin: Not sure about `executor_label` as the name for this column. What about just `name`?

lucaspin (approved):

PR #177, opened 2025-08-22
feat: edit and cleanup stages and event sources

forestileao:
### Changes
- Added Update Event Source
- Added Clean webhooks procedure if the resource name is changed
- Added CLI update
- Added UI update


pkg/builders/stage.go
```diff
@@ -325,3 +333,65 @@ func (b *StageBuilder) Update() (*models.Stage, error) {
...
+
+	oldIntegration, err := models.FindIntegrationByID(oldResource.IntegrationID)
+	if err != nil {
+		return nil
+	}
+
+	isIntegrationChanging := b.integration != nil && oldIntegration.ID != b.integration.ID
+	isResourceChanging := b.resource != nil && (oldResource.Name() != b.resource.Name() || oldResource.Type() != b.resource.Type())
+
+	if !isIntegrationChanging && !isResourceChanging {
```
lucaspin: I don't think this is the only thing we should take into account when determining if an event source and its webhook should be deleted.

There are a couple more things to take into account:
- We shouldn't delete anything if the stage is re-using a externally-scoped (created by the user) event source
- Even if the event source is internally-scoped, if another stage is re-using it, we shouldn't delete it as well

Also, since this is all a bit complex, we should probably try to make it all async, and not part of the stage update API call

forestileao: The solution I found was adding two consumers that check for both of these cases. According to my tests everything goes well for all these cases.

pkg/grpc/actions/event_sources/update_event_source.go
```diff
@@ -0,0 +1,124 @@
...
+	pb "github.com/superplanehq/superplane/pkg/protos/canvases"
+	"github.com/superplanehq/superplane/pkg/registry"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"gorm.io/gorm"
+)
+
+func UpdateEventSource(ctx context.Context, encryptor crypto.Encryptor, registry *registry.Registry, orgID, canvasID, idOrName string, newSource *pb.EventSource) (*pb.UpdateEventSourceResponse, error) {
+	_, userIsSet := authentication.GetUserIdFromMetadata(ctx)
+	if !userIsSet {
```
hamir-suspect, line 24: This looks like it should be implemented as a middleware

pkg/grpc/canvas_service.go
```diff
@@ -75,6 +75,12 @@ func (s *CanvasService) CreateEventSource(ctx context.Context, req *pb.CreateEve
 	return eventsources.CreateEventSource(ctx, s.encryptor, s.registry, orgID, canvasID, req.EventSource)
 }
 
+func (s *CanvasService) UpdateEventSource(ctx context.Context, req *pb.UpdateEventSourceRequest) (*pb.UpdateEventSourceResponse, error) {
+	orgID := ctx.Value(authorization.OrganizationContextKey).(string)
+	canvasID := ctx.Value(authorization.DomainIdContextKey).(string)
+	return eventsources.UpdateEventSource(ctx, s.encryptor, s.registry, orgID, canvasID, req.IdOrName, req.EventSource)
```
hamir-suspect, line 81: if these values are in ctx why are then also an argument?

pkg/logging/logging.go
```diff
@@ -73,3 +73,17 @@ func ForEvent(event *models.Event) *log.Entry {
 		},
 	)
 }
+
+func ForEventSource(eventSource *models.EventSource) *log.Entry {
```
hamir-suspect: unused

pkg/grpc/actions/event_sources/update_event_source.go
forestileao: It is implemented in a middleware: https://github.com/superplanehq/superplane/blob/main/pkg/authorization/interceptor.go#L130

For this case, we are using metadata because grpc gateway transforms headers and put them right there.

pkg/grpc/canvas_service.go
forestileao: For those patterns, we have a justification. And the answer is: Grpc Gateway.
We have an interceptor logic that works like a middleware. Since Grpc Gateway creates a new context for itself,  we use it directly because it is set by the auth interceptor. then, We are guaranteeing that what we have was verified by the authorization interceptor.

pkg/logging/logging.go
forestileao: Good catch! Removed!

hamir-suspect (approved):

PR #182, opened 2025-08-27
fix: nil pointer dereferencing for http executor

forestileao:
fix: nil pointer dereferencing for http executor for HTTP status codes


darkofabijan (approved):
Where is a test? :)

darkofabijan (approved):

Issue #227, opened 2025-09-05
Pending Executions Stuck


forestileao:
### Issue
After the creation of a new canvas (I added you there as well) and the errors are the same, I have an event, but the stage does not start any pipelines on Semaphore.

<img width="2968" height="1528" alt="Image" src="https://github.com/user-attachments/assets/c370abe2-83e8-41c0-b692-3b3dd2a6195e" />

We have a worker who checks for pending executions and starts processing them. But currently it is stopping to process the rest of the executions that are pending once one of them if broken. 

### How the issue happened:
A Stage was deleted when an execution was pending
Once the execution was still pending, the worker was trying to process that, failing all the others.

### Fix
First, delete the pending execution from the deleted stage
Fix hard deletion worker (for some reason didn't delete the execution)
Fix the Pending Executions worker to not stop processing all others once one of them fails

forestileao:
@AleksandarCole Can we close this one?

AleksandarCole:
@forestileao I am not sure how to verify this, let's push the fix to production and ask Radek to confirm since he originally reported it. 

Let's keep in review until verified.

PR #228, opened 2025-09-05
fix: pending executions worker should filter errored executions

forestileao:
Issue: https://github.com/superplanehq/superplane/issues/227

### Description
Since we are hard-deleting components async after soft-deletion and the pending executions worker was breaking every time it was trying to get a stage that was soft-deleted (our query would not find since it is auto-managed by gorm), we should just log these errors and proceed with the next executions. Or the entire application would stop working for all canvases.

Also, fixed event sources and stage resource hard deletions that were avoiding the hard deletion of some stages and event sources.

### Changes

- [fix: Pending Executions Worker should filter errored executions and log it](https://github.com/superplanehq/superplane/commit/bb1b89961a9b0c776a9c99423f23c352204add78)
- [fix: foreign key constraint violation while deleting event sources](https://github.com/superplanehq/superplane/commit/6b79e2c6bcd62bab88209c0660803273931cd307)
- [fix: stage execution resource deletion](https://github.com/superplanehq/superplane/commit/edf0edc47b0bc3268e61544d4c5cbd590064d4cd)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.24
+ARG GO_VERSION=1.25
 ARG UBUNTU_VERSION=22.04
 ARG BUILDER_IMAGE="golang:${GO_VERSION}"
 ARG RUNNER_IMAGE="ubuntu:${UBUNTU_VERSION}"
@@ -71,7 +71,7 @@ RUN go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
 RUN go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@latest
 RUN go install github.com/air-verse/air@latest
 RUN go install github.com/mgechev/revive@v1.8.0
-RUN go install gotest.tools/gotestsum@v1.12.1
+RUN go install gotest.tools/gotestsum@v1.12.3
 
 CMD [ "/bin/bash",  "-c \"while sleep 1000; do :; done\"" ]
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/superplanehq/superplane
 
-go 1.24
+go 1.25
 
 require (
 	github.com/casbin/casbin/v2 v2.106.0


pkg/workers/pending_executions_worker_test.go
```diff
@@ -197,6 +198,145 @@ func Test__PendingExecutionsWorker(t *testing.T) {
...
+			Create()
+		require.NoError(t, err)
+
+		validExecution := support.CreateExecution(t, r.Source, validStage)
+
+		// Soft delete the stage after creating the execution to simulate production issue
+		err = database.Conn().Delete(&invalidStage).Error
+		require.NoError(t, err)
+
+		//
```
lucaspin: This comment makes sense in the context of this PR, but it won't make a lot of sense after it, since "original bug" and "fix" only exist in the context of this PR

pkg/models/stage_execution.go
```diff
@@ -425,3 +425,28 @@ func (e *StageExecution) AddResource(externalID string, externalType string, par
 
 	return r, nil
 }
+
+func DeleteStageExecutionsBySourceInTransaction(tx *gorm.DB, sourceID uuid.UUID, sourceType string) error {
+	if err := tx.Unscoped().
+		Where("execution_id IN (SELECT id FROM stage_executions WHERE stage_event_id IN (SELECT id FROM stage_events WHERE source_id = ? AND source_type = ?))", sourceID, sourceType).
```
lucaspin, line 431: What's the reason for using nested queries instead of join here? Join is easier to understand

pkg/workers/pending_executions_worker_test.go
forestileao: My Bad, used IA to help with test setup and forgot to remove these comments

pkg/models/stage_execution.go
forestileao: Postgres do not support joins in delete queries: https://stackoverflow.com/questions/76896069/how-to-use-joins-in-delete-using-gorm

forestileao: Rolled back to nested queries version

Issue #210, opened 2025-09-02
Stage broken after event source rename


AleksandarCole:
1. Add event source name it (Listen something)
2. Add stage, connect to "Listen something"
3. Rename "Listen something" to "Listen something-else"
4. Stage goes to `broken` state as connection doesn't exist anymore
5. Try to update stage by deleting old connection and adding new connection to "Listen something-else" 
6. Fails to save (there's no error, no validation, but console error still shows old name of connection)

Solution:
- Source rename should not break stage (update stages that listen to it)
- There should be some validation when I try to save new connection (give me error if it fails)

AleksandarCole:
@forestileao this one is not good. Check the screenshot
1. I renamed event source from "Listen terraform" to "Listen to: terraform"
2. Deploy Devel appears broken
3. I go to Edit `Deploy Devel`
4. Connection seems to be updated properly
5. I bang my head against the table trying to figure out what's wrong
6. I remember that I have input mapping depending on connection
7. They are the ones that are broken

<img width="671" height="870" alt="Image" src="https://github.com/user-attachments/assets/1fdf9323-90cf-4352-a1e0-dd4949996ffc" />


We need to follow the same patterns. If I change the connection name either:
1. Stage breaks and faulty connection and faulty inputs are highlighted as broken when I try to edit it (worse option)
2. OR - Stage doesn't break because connection name was updated everywhere (preferred)

AleksandarCole:
Reassigning to @lucaspin and @AleksandarCole to figure out the name/display-name thing

PR #238, opened 2025-09-10
fix: when udpating component names change connection source names

forestileao:
https://github.com/superplanehq/superplane/issues/210
### Changes
- Add connection update in transaction for components + tests
- Updated frontend to automatically update connections in UI

https://github.com/user-attachments/assets/a8194edc-c346-4713-8bff-df7f0c75508b


lucaspin (approved):

pkg/models/connection.go
```diff
@@ -87,3 +88,77 @@ func ListConnectionsInTransaction(tx *gorm.DB, targetID uuid.UUID, targetType st
 
 	return connections, nil
 }
+
+type StageNotifier interface {
+	NotifyStageUpdated(stage *Stage)
+}
+
+func UpdateConnectionSourceNameInTransaction(tx *gorm.DB, canvasID uuid.UUID, sourceID uuid.UUID, sourceType string, oldName string, newName string, notifier StageNotifier) error {
```
lucaspin: This function name is no longer applicable. Maybe `UpdateReferencesAfterNameUpdate()`?

pkg/builders/notifiers.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/superplanehq/superplane/pkg/grpc/actions/messages"
+	"github.com/superplanehq/superplane/pkg/models"
+)
+
+type MessageStageNotifier struct{}
```
lucaspin: I don't understand why we need this. Don't we do this from pkg/grpc/actions/stages/update_stage.go?

pkg/models/connection.go
forestileao: Name updated to UpdateReferencesAfterNameUpdateInTransaction.

pkg/builders/notifiers.go
forestileao: we need this to tell the UI the stage connection was updated, so it does not sha ow as broken stage (the user does not need to refresh the page). 
I created MessageStageNotifier because of import cycles errors

forestileao: Removed it, we can avoid async inconsistency by updating directly the connections in the frontend

pkg/executors/semaphore/semaphore.go
lucaspin, line 1: Why do we need this?

forestileao: my bad, that should not have been commited

lucaspin (approved):

Issue #239, opened 2025-09-10
Superplane requesting read/write scopes when logging in with GitHub


lucaspin:
Superplane is requesting read/write scopes when logging in with GitHub

<img width="575" height="345" alt="Image" src="https://github.com/user-attachments/assets/caf01dd4-07d3-47cb-9a04-41b48ade01e6" />

---

We should use the read-only scope instead.

PR #240, opened 2025-09-10
fix: do not specify GitHub scope

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/239

Not using a scope grants read-only access to public information only (including user profile info, repository info, and gists). See https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps

### DIFF
--- a/pkg/models/constants.go
+++ b/pkg/models/constants.go
@@ -4,7 +4,6 @@ import "fmt"
 
 const (
 	ProviderGitHub = "github"
-	ScopeUser      = "user"
 
 	DomainTypeOrganization = "org"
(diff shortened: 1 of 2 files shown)


pkg/models/constants.go
```diff
@@ -4,7 +4,6 @@ import "fmt"
 
 const (
 	ProviderGitHub = "github"
-	ScopeUser      = "user"
```
forestileao, line 7: Do not remove, we need that to get the email information `read:user`

lucaspin: Removed because pkg/models/constants.go is for constants used in our database models.

PR #244, opened 2025-09-11
refactor: split execution list into a separate endpoint

forestileao:
This pull request updates the state transitions for stage events. This [whimsical](https://whimsical.com/event-states-2u4PCE3FLmjNYeybobKQTK)  of how the state of a stage event is updated while the event is being processed.

### Main changes

- When a stage execution is started, the stage event no longer is moved to `waiting(execution)` state. Instead, the stage event is moved to `processed` state. The `waiting(X)` state is reserved to stage events that are waiting on conditions to be met.
- The already existing `GET /stages/{id}/events` endpoint no longer includes execution information.
- A new `GET /stages/{id}/executions` exposes information about stage executions.
- A new `POST /stages/{id}/executions/{executionId}/cancel` was added to allow stage executions to be cancelled.
- The `GET /event-sources` and `GET /stages` endpoints were updated to include status information about the resources being returned. That allows for us to implement the lazy loading without the need of the bulk endpoints, which were removed.
- Cleaned up unused state reasons

### DIFF
--- a/web_src/src/pages/canvas/components/StageEditModeContent.tsx
+++ b/web_src/src/pages/canvas/components/StageEditModeContent.tsx
@@ -17,7 +17,6 @@ import IntegrationZeroState from '@/components/IntegrationZeroState';
 import { createInputMappingHandlers } from '../utils/inputMappingHandlers';
 import { twMerge } from 'tailwind-merge';
 import { showErrorToast } from '@/utils/toast';
-import { Button } from '@/components/Button/button';
 
 interface StageEditModeContentProps {
   data: StageNodeType['data'];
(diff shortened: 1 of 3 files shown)


pkg/workers/eventdistributer/execution_cancelled.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to marshal websocket event: %w", err)
+	}
+
+	wsHub.BroadcastToCanvas(executionCancelled.CanvasId, wsEventJSON)
+	log.Debugf("Broadcasted execution_cancelled event to canvas %s", executionCancelled.CanvasId)
+
+	return nil
+}
```
forestileao, line 57: I am trying to avoid using snake case in the Api. Also because our pattern in GRPC gateway is camel case the best thing here would be use the same pattern of serialization we have for stage_created.

https://github.com/superplanehq/superplane/blob/main/pkg/workers/eventdistributer/stage_created.go

Note that we already have a proto message. you just need to use the pb masrshler to correctly parse it to json (unless you are going to parse Enums as numbers instead of strings). This should also remove the amount of code.

web_src/src/pages/canvas/hooks/useWebsocketEvents.ts
```diff
@@ -136,24 +154,60 @@ export function useWebsocketEvents(canvasId: string, organizationId: string): vo
...
+        queryClient.invalidateQueries({
+          queryKey: canvasKeys.stageExecutions(canvasId, executionCancelledPayload.stage_id)
+        });
         break;
       default:
     }
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [lastJsonMessage, addEventSource, addStage, updateCanvas, updateStage, syncStageEvents]);
+  }, [lastJsonMessage, addEventSource, addStage, updateCanvas, updateStage, syncStageEvents, syncStageExecutions]);
```
forestileao, line 212: hmm but actually we have a bunch of things here also using snake case. Specially for stage events. Maybe we can do it later i another PR. Since this is not a component update it will not break.,

pkg/workers/eventdistributer/execution_cancelled.go
forestileao: Let's keep like it rn, because of https://github.com/superplanehq/superplane/pull/244#discussion_r2349188484

PR #247, opened 2025-09-12
refactor: expand connections and executor config if they are not configured

forestileao:
## Changes

This PR applies changes in 'StageEditModeContent.tsx'. More specifically to the validation of the stage and and executor.

Including: 
- When creating - Connections is expanded (if not configured), Executor is expanded (both are showing red text that they have to be configured) 
- There's no Add executor button you can have only one executor
- When executor is saved, it's collapsed, no need to have that "semaphore executor" minimised item
- On Api errors it should also open the executor config. The error fallback with a toast will be added in another pr.
- Added to open sections on Stage Actions "Save" validation

image
image


forestileao:
@lucaspin Added a fix to also open the errored sections when we click on "Save" from the floating actions bar.

PR #256, opened 2025-09-16
feat: track event rejections while processing events

lucaspin:
### Issue

Currently, if an event is filtered, or something goes wrong while applying the filters, or building the inputs, the event is rejected, but we only log an error message, which doesn't allow users to see what happened with their events that were rejected. That makes it hard to troubleshoot issues with filters or inputs.

### Solution

This pull request addresses the issue in two ways:
- New state_reason and state_message fields in events table. Those fields are used when processing event source filters. If something goes wrong, or filters don't pass, the state_reason is updated accordingly
-  New event_rejections table and GET /event_rejections endpoint to expose them. These are used when processing filters for stages and connection groups (and any future component type as well)

Note: the UI updates to display these event rejections for stages will come in a later pull request


pkg/grpc/actions/canvases/list_event_rejections.go
```diff
@@ -0,0 +1,153 @@
...
+		HasNextPage:   result.hasNextPage(limit),
+		LastTimestamp: result.lastTimestamp(),
+	}
+
+	return response, nil
+}
+
+type listAndCountEventRejectionsResult struct {
+	rejections []models.EventRejection
+	totalCount int64
```
forestileao, line 63: The proto response is uint32, why are we using int64 in a bunch of places?
uint32 max value is more than enough (4,294,967,295).

But do not use uint64 in proto responses. since when parsing to json it will become a string in javascript

pkg/models/event_rejection.go
```diff
@@ -0,0 +1,99 @@
...
+	query = query.Order("rejected_at DESC").Limit(limit)
+	err := query.Find(&rejections).Error
+	if err != nil {
+		return nil, err
+	}
+
+	return rejections, nil
+}
+
+func CountEventRejections(targetType string, targetID uuid.UUID) (int64, error) {
```
forestileao, line 86: here too

pkg/grpc/actions/canvases/list_event_rejections.go
```diff
@@ -0,0 +1,153 @@
...
+	listErr    error
+	countErr   error
+}
+
+func (r *listAndCountEventRejectionsResult) hasNextPage(limit uint32) bool {
+	return len(r.rejections) == int(limit) && r.totalCount > int64(limit)
```
forestileao, line 69: here too, we can remove the casting

pkg/grpc/actions/canvases/list_event_rejections.go
lucaspin: We are using int64 because the Count() method in gorm expects a int64

PR #294, opened 2025-09-19
feat: add update integration endpoint

forestileao:
feat: add simple update integration service


pkg/grpc/actions/integrations/update_integration.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	integration.Name = updatedIntegration.Name
+	integration.Type = updatedIntegration.Type
+	integration.URL = updatedIntegration.URL
+	integration.AuthType = updatedIntegration.AuthType
+	integration.Auth = updatedIntegration.Auth
+
+	err = integration.Update()
```
lucaspin, line 62: What happens if an integration was already used to set up some webhooks into an account, and we update it to point to something else?

forestileao: we should block updating integrations that are being used by a resources. Added a verification for that and a toast error

Issue #145, opened 2025-08-20
Timed/Scheduled Runs


AleksandarCole:
# Timed/Scheduled Runs - Design Discussion

## Why Is It Important
**User Pain Points:**
- No way to automate recurring workflows like daily builds, nightly tests, or periodic deployments
- Cannot schedule individual stages within workflows for different timing requirements
- Lack of built-in scheduling forces users to rely on external automation tools

## Problem
User feedback indicates need for timed/scheduled runs triggered at user-defined intervals (e.g., daily builds). We need both stage-level scheduling to trigger single nodes in existing workflows, and event source elements that start entire workflow chains on schedule.

```mermaid
graph LR
    SE[⏰ Scheduled Event Source<br/>Daily at 9 AM] --> B1[🔨 Nightly Build<br/>Project A]
    SE --> B2[🔨 Nightly Build<br/>Project B]
    
    B1 --> MS[🎯 Manual Deploy to Staging<br/>User selects version]
    B2 --> MS
    
    MS --> MR[🚀 Manual Release<br/>User selects version<br/>for official release]
    
    style SE fill:#e8f5e8
    style B1 fill:#e3f2fd
    style B2 fill:#e3f2fd
    style MS fill:#fff3e0
    style MR fill:#fce4ec
```

# Scheduled Event Source proposal

### Summary
Implement a new event source type that allows users to create multiple time-based schedules within a single event source. Each schedule uses cron expressions to trigger events with predefined JSON payloads.

### Functional Requirements
- **Event Source Type**: New "Scheduled" event source type
- **Multiple Schedules**: Support up to 5 individual schedules per event source
- **Cron Scheduling**: Standard 5-field cron expressions (`minute hour day month weekday`)
- **Timezone**: All schedules operate in UTC timezone
- **JSON Payloads**: Each schedule emits a static, user-defined JSON payload
- **Schedule Control**: Individual enable/disable functionality for each schedule
- **Event Delivery**: SuperPlane needs to be able to emit internal events

### Technical Requirements
- Standard 5-field cron expression validation
- UTC-based scheduling engine
- Error handling and logging for schedule execution

<img width="1076" height="808" alt="Image" src="https://github.com/user-attachments/assets/6d5aed07-35f1-4244-b957-360019109194" />

lucaspin:
A few notes/questions:
- Why does the source need to support multiple schedules? To me, it seems more natural if each single source had a single schedule, and if you need multiple schedules, you create multiple sources
[...]

PR #296, opened 2025-09-22
feat: event source with schedule

lucaspin:
This PR introduces the ability to schedule event sources to trigger automatically at specified intervals. See: https://github.com/superplanehq/superplane/issues/145

### Key Features

- Scheduling Types: Support for hourly, daily, and weekly schedules
- Backend Infrastructure: New database schema, protobuf definitions, and schedule worker
- UI Components: Schedule configuration interface with time/day selection
- Worker System: Background worker to process scheduled triggers, controlled by `START_EVENT_SOURCE_SCHEDULE_WORKER` environment variable

### Other changes

- The new component (`ScheduleConfiguration`) was already in `src/components` and has storybook stories for it. Other components required by it were also moved to that folder, and some storybook stories were added for them as well. Those are: EventSourceZeroState and ValidationField.

### Examples

This is what the new event source looks like in the canvas:

Screenshot 2025-09-21 at 11 57 54


pkg/models/event_source.go
```diff
@@ -42,6 +58,207 @@ type EventType struct {
...
+
+	targetWeekday, err := parseWeekday(weekDay)
+	if err != nil {
+		return nil, fmt.Errorf("invalid weekday: %v", err)
+	}
+
+	nowUTC := now.UTC()
+	currentWeekday := nowUTC.Weekday()
+
+	daysUntilTarget := int(targetWeekday - currentWeekday)
```
forestileao, line 150: This logic assumes we always want the next occurrence, but when daysUntilTarget == 0 (today), it sets the trigger for today regardless of whether the time has passed. The subsequent check tries to fix this, but it's cleaner to handle this upfront, no?

forestileao (approved):

PR #297, opened 2025-09-22
refactor: update integrations page design

forestileao:
Look at: https://github.com/orgs/superplanehq/projects/2/views/1?pane=issue&itemId=125156903&issue=superplanehq%7Csuperplane%7C164

- the only thing different from the issue is that we should be able to update the other fields if it is not used by any other resource

image


lucaspin:
@pedroforesti 

I don't understand why the "Value" field is shown when I select "Import from existing secret". Is that correct? Shouldn't I just need to select the secret key in that case?

Screenshot 2025-09-22 at 18 00 03

forestileao:
> @pedroforesti
> 
> I don't understand why the "Value" field is shown when I select "Import from existing secret". Is that correct? Shouldn't I just need to select the secret key in that case?
> 
> Screenshot 2025-09-22 at 18 00 03

@lucaspin That case is just used when you are creating the secret for first time in the modal. The real value is not there, it is just to show that it is a value from a secret that cannot be displayed. And this input is not updatable.

web_src/src/components/IntegrationForm/GitHubIntegrationForm.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+        <Story />
+      </div>
+    ),
+  ],
+}
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+const mockSecrets = [
```
lucaspin: These secrets seem to be shared between all these stories. It would be nice if we had some kind of factory for resources that could be reused across all the stories

forestileao: addressed

web_src/src/components/IntegrationForm/__mocks__/storyFactory.ts
```diff
@@ -0,0 +1,122 @@
+import { IntegrationsIntegration, SecretsSecret } from "@/api-client"
+
+export const createMockSecrets = (): SecretsSecret[] => [
```
lucaspin: Why do we need different functions, createMockSecrets, createGitHubSecrets, createSemaphoreMockSecrets, createFlowMockSecrets? Isn't it easier to just have one mockedSecrets function?

web_src/src/components/IntegrationForm/ApiTokenForm.stories.tsx
```diff
@@ -0,0 +1,270 @@
+import React, { useState } from 'react'
+import type { Meta, StoryObj } from '@storybook/react'
+import { MemoryRouter } from 'react-router-dom'
+import { ApiTokenForm } from './ApiTokenForm'
+import type { IntegrationData, FormErrors } from './types'
+import { createMockSecrets, defaultProps } from './__mocks__/storyFactory'
```
lucaspin: Also, not sure if this is the best place to put this kind of thing. Wouldn't it be better to put those in a web_src/test folder or something? What are the patterns for this kind of scenario?

Issue #118, opened 2025-08-15
Manual Stage Run Triggers


AleksandarCole:
## Problem
Currently, stages must be connected to event sources to execute - there's no option for users to manually trigger run executions. We want users to be able to start workflows by manually running any stage in the event chain, making event source connections optional rather than required.

## Current State
- Stages require event source connections to execute
- No manual trigger option exists
- Workflows can only start from event sources (webhooks, GitHub, Semaphore)
- Users cannot test individual stages or start workflows on-demand
- Event connections are mandatory for stage functionality

## Questions to Resolve
- [ ] How do we handle dependencies when manually triggering a stage in the middle of a chain?
- [ ] How does the UI look when event source connections become optional vs required?
- [ ] How do we handle inputs on manually triggered stages (no event payload available)?
- [ ] Should we allow parameterization for manual triggers (user-provided inputs)?
- [ ] What happens to downstream stages when manually triggering an upstream stage?
- [ ] Should manual runs show differently in the execution history?
- [ ] How do we distinguish between event-triggered runs vs manual runs?
- [ ] Do we need a "Run with parameters" modal for complex inputs?

### Manual Trigger Scenarios
```mermaid
graph LR
    E1[Event Source] -.-> |optional| S1[Stage 1]
    S1 --> S2[Stage 2]
    S2 --> S3[Stage 3]
    
    M1[🚀 Manual Trigger] -.-> S1
    M2[🚀 Manual Trigger] -.-> S2
    M3[🚀 Manual Trigger] -.-> S3
    
    S2 -.-> |needs S1 inputs?| Q1[Input Problem]
    S3 -.-> |needs S1 + S2 inputs?| Q2[Dependency Chain]
    
    style E1 stroke-dasharray: 5 5
    style M1 fill:#e8f5e8
    style M2 fill:#e8f5e8
    style M3 fill:#e8f5e8
    style Q1 fill:#fff3e0
    style Q2 fill:#fff3e0
```

## Constraints
- Must maintain backward compatibility with existing event-driven workflows
- Cannot break existing stage configurations
- Manual triggers should feel intuitive and not confusing
- Need to handle missing input data gracefully
- Performance considerations for manual trigger UI responsiveness

## Next Steps
- [ ] Schedule discussion session
- [ ] Design manual trigger UI patterns
- [ ] Define input handling for manual runs
- [ ] Create dependency resolution strategies
- [ ] Design parameterization interface for complex inputs

lucaspin:
[...]

PR #298, opened 2025-09-22
feat: manually emit from event source

lucaspin:
Task: https://github.com/superplanehq/superplane/issues/118

### API changes

- New `POST /canvases/{id}/events` endpoint to manually create an event for a source

### UI changes

The event source node gets a new "Send" button in the event source top left corner

Screenshot 2025-09-22 at 17 23 16

When clicked, a modal comes up with a form that allows the user to input the event type and payload:

Screenshot 2025-09-22 at 17 23 07

To make things even easier, the payload editor is pre-filled with last emitted event for the source.


pkg/grpc/canvas_service.go
```diff
@@ -196,6 +196,12 @@ func (s *CanvasService) ListEvents(ctx context.Context, req *pb.ListEventsReques
 	return events.ListEvents(ctx, canvasID, req.SourceType, req.SourceId, req.Limit, req.Before)
 }
 
+func (s *CanvasService) CreateEvent(ctx context.Context, req *pb.CreateEventRequest) (*pb.CreateEventResponse, error) {
+	canvasID := ctx.Value(authorization.DomainIdContextKey).(string)
+	rawData := req.Raw.AsMap()
```
forestileao: can req.Raw be nil?

pkg/grpc/actions/events/create_event.go
```diff
@@ -0,0 +1,84 @@
...
+
+	sourceName, err := findSourceName(parsedCanvasID, sourceType, parsedSourceID)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "source not found: %v", err)
+	}
+
+	rawBytes, err := json.Marshal(raw)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid raw data: %v", err)
+	}
```
forestileao, line 50: if raw is nil, would it still be a valid Json String?
asking this because in Javascript, null is a valid json
image

pkg/grpc/canvas_service.go
lucaspin: Added a condition to not allow nil raw data

PR #301, opened 2025-09-22
refactor: remove unique name and display name from org

forestileao:
refactor: remove unique name and db constraint from org

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,7 +2,7 @@
 -- PostgreSQL database dump
 --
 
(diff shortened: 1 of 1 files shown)


db/migrations/20250922200128_remove-display-name-from-organizations.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE organizations DROP COLUMN display_name;
+ALTER TABLE organizations DROP CONSTRAINT organizations_name_key;
```
lucaspin: I think we should still keep this unique constraint. Allowing organizations with the same name could be confusing

forestileao: removed this alter table

PR #306, opened 2025-09-23
feat: manually emit event for stage

lucaspin:
Allow the EmitEventModal component to be used for stage nodes as well. Since the last emitted event by a stage is not exposed by the API, we need to load it using the ListEvents endpoint.

### Other changes

- I was getting collation mismatch errors - `The database was created using collation version 2.36, but the operating system provides version 2.41` - when running make `dev.setup` and `test.setup`. I suspect this is due to updates in the Postgres image we are using in our docker compose, so I updated the `db.create` Makefile target to always run `REFRESH COLLATION VERSION` for `template1` and `postgres` databases before creating new ones.
- Emit a RabbitMQ message when CreateEvent endpoint is used.


web_src/src/components/EmitEventModal/EmitEventModal.stories.tsx
```diff
@@ -2,27 +2,40 @@ import type { Meta, StoryObj } from '@storybook/react'
...
       <EmitEventModal
         isOpen={isOpen}
         onClose={() => setIsOpen(false)}
         sourceName={sourceName}
-        lastEvent={lastEvent}
+        loadLastEvent={loadLastEvent || defaultLoadLastEvent}
         onCancel={() => {}}
-        onSubmit={async () => {}}
+        onSubmit={async (eventType, eventData) => {
+          console.log('Emitting event:', { eventType, eventData })
```
forestileao:
```suggestion
```

avoid console.logs, unless it is console.error for something it should not happen

PR #308, opened 2025-09-23
fix: event source resource deletion

forestileao:
## Changes

- Added resource deletion in for event sources if allowed to delete
- Added InTransacion method for that, since it was causing deadlocks with nested transactions for some reason
image

### DIFF
--- a/pkg/workers/hard_deletion_worker.go
+++ b/pkg/workers/hard_deletion_worker.go
@@ -93,7 +93,7 @@ func (w *HardDeletionWorker) processStages() error {
 // hardDeleteStage handles hard deletion of soft-deleted stages following hierarchical dependency chain:
 // 1. Delete stage executions and their execution resources FIRST
 // (executions have foreign key to stage_events, so must be deleted first)
-// 2. Now delete stage events and their associated events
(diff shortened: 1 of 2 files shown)


pkg/workers/resource_cleanup_test.go
```diff
@@ -151,3 +151,44 @@ func Test__ResourceCleanupService(t *testing.T) {
...
+			WithName("test-event-source-single").
+			WithScope(models.EventSourceScopeExternal).
+			ForIntegration(r.Integration).
+			ForResource(resource).
+			Create()
+		require.NoError(t, err)
+
+		_, err = service.CleanupEventSourceWebhooks(eventSource, resource)
+		assert.NoError(t, err)
+		// The result depends on whether other event sources use this resource
```
lucaspin, line 187: Why can't we predict it on this test?

forestileao: because the cleanup of webhooks uses an external api, it will always return false in this case, since it is a fake webhook. We don't consider it an error because it is possible that the user might manually delete a webhook and the api get a 404. It should not brake the resource deletion process

PR #339, opened 2025-09-29
fix: process executions in parallel in PendingExecutionsWorker

lucaspin:
### Issue

The PendingExecutionsWorker processes executions sequentially. That allows for a slow execution (HTTP call that takes a long time) to block other executions to proceed.

### Solution

We combine a couple of things here:
- We use goroutines executions in parallel
- We use a semaphore from [golang.org/x/sync/semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore) to control the number of concurrent goroutines
- We use a transaction for each execution, locking the record with `FOR UPDATE SKIP LOCKED` to avoid other workers from processing it at the same time.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,5 +33,6 @@
 - Tests end with _test.go
 - Always prefer early returns over else blocks when possible
 - GoLang: prefer `any` over `interface{}` types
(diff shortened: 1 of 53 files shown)


forestileao (approved):

AleksandarCole:
@lucaspin tested on `prod` - works a lot faster now 👍

PR #353, opened 2025-09-30
fix: delete hard deletion for shared resource

forestileao:
### Fixes
Before, when you were deleting the event source before the stage that had a shared resource, the event source was cleaning the webhook. The behaviour wasnt happenning when you were deleting the stage before the event source.

It was happening due to 2 things: repeated code for cleaning webhooks and a wrong verification, since GORM already discards soft-deleted entities in scoped count queries.

### DIFF
--- a/pkg/workers/execution_poller.go
+++ b/pkg/workers/execution_poller.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	DefaultExecutionTimeout          = 3 * time.Hour
-	ExecutionResourcePollingInterval = 5 * time.Minute
+	ExecutionResourcePollingInterval = 1 * time.Minute
 )
 
 type ExecutionPoller struct {


pkg/workers/resource_cleanup.go
```diff
@@ -192,13 +144,25 @@ func (s *ResourceCleanupService) CleanupEventSourceWebhooksInTransaction(tx *gor
 
 	logger := log.WithField("event_source_id", eventSource.ID)
 
-	count, err := models.CountExternalEventSourcesUsingResourceInTransaction(tx, *eventSource.ResourceID)
+	externalEventSourceCount, err := models.CountExternalEventSourcesUsingResourceInTransaction(tx, *eventSource.ResourceID)
```
lucaspin, line 147: externalEventSourceCount is always 0 or 1, right?

forestileao: yes, it could be an "existExternalEventSources", but since we are going allow many event sources in the same canvas in the future I kept that as a number

forestileao: Actually it can be an exist in this case as well, WDYT, is it better?

lucaspin: Right. And what if externalEventSourceCount is 0, but stagesCount is > 0? I still don't understand why we don't need to move the event source scope back to `internal` in that case

forestileao: I see, the webhook would still exist but we would lose the internal event source reference. Good catch, let me fix this

forestileao: Updated the implementation and tests. Now when we delete an event source with shared resources, it becomes internal scoped and we set deleted_at to null. This internal event source will be deleted once all stages that uses them are deleted as well.
Updated the shared resource test for that situation

lucaspin (approved):

PR #361, opened 2025-10-07
feat: auto complete for input mappings

forestileao:


forestileao:
@lucaspin I made some UI / UX updates, can you re-review it?
image

PR #365, opened 2025-10-10
feat: add expression validation on filters

forestileao:
- Added expression validation using web assembly module from expr lang to validate boolean expressions before saving

image


shiroyasha:
This is a deprecated PR and we can close it, right? @forestileao

forestileao:
@shiroyasha Closing it, we kept it open to see how we made this WebAssembly expr-validation on the frontend.
But we can go back here if we need a reference.

PR #369, opened 2025-10-15
feat: add invitation by email + assign canvas to invititations

forestileao:
### Changes
This PR adds an invitation Rabbit MQ consumer to notify users by email that they were invited to a certain organization.
In this PR, we also added canvasIDs field to an invitation, so, when the user signs in for the first time, they will be assigned as a viewer of the canvas as well.

In the api side, the signature of the RemoveUser endpoints was changed to RemoveSubject to accept invitation and users at the same time. Besides, the assign role endpoint also allows invitations to be assigned, but only canvas viewer roles.

### DIFF
--- a/pkg/services/email.go
+++ b/pkg/services/email.go
@@ -20,6 +20,7 @@ type SendGridEmailService struct {
 	fromName    string
 	fromEmail   string
 	templateDir string
+	client      *sendgrid.Client
 }
 
 func NewSendGridEmailService(apiKey, fromName, fromEmail, templateDir string) *SendGridEmailService {
@@ -28,6 +29,7 @@ func NewSendGridEmailService(apiKey, fromName, fromEmail, templateDir string) *S
 		fromName:    fromName,
 		fromEmail:   fromEmail,
 		templateDir: templateDir,
+		client:      sendgrid.NewSendClient(apiKey),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/models/organization_invitation.go
```diff
@@ -17,6 +17,7 @@ const (
 type OrganizationInvitation struct {
 	ID             uuid.UUID `gorm:"type:uuid;primary_key;default:gen_random_uuid()"`
 	OrganizationID uuid.UUID
+	CanvasIDs      UUIDArray `gorm:"type:text[]"`
```
lucaspin: Probably easier just to use a datatypes.JSONType[[]string] since we already use datatypes.JSONType everywhere else

forestileao: Updated to jsonb

pkg/services/email.go
```diff
@@ -0,0 +1,97 @@
...
+
+	htmlContent, err := s.renderTemplate("invitation.html", templateData)
+	if err != nil {
+		log.Errorf("Error rendering HTML template: %v", err)
+		return fmt.Errorf("failed to render HTML template: %w", err)
+	}
+
+	message := mail.NewSingleEmail(from, subject, to, plainTextContent, htmlContent)
+
+	client := sendgrid.NewSendClient(s.apiKey)
```
lucaspin: Can we reuse the same client instead of creating a new one every time?

lucaspin (approved):

PR #411, opened 2025-10-24
Story ephemeral environments

AleksandarCole:
A new story showing provisioning/deprovisioning ephemeral environments use-case


shiroyasha:
I'm planning bigger changes that would leave this branch non-mergable on monday.
Merging it as it is. You can continue with the rest of the flow on monday.

Issue #541, opened 2025-10-31
(Homepage) Show who created the canvas


shiroyasha:
Previously this was visible on the canvas cards.
Now, this is no longer available when we converted to the new system.

PR #584, opened 2025-11-01
feat: Show canvas creator

shiroyasha:
fixes https://github.com/superplanehq/superplane/issues/541

CleanShot 2025-11-01 at 15 29 24@2x


pkg/grpc/actions/workflows/create_workflow.go
```diff
@@ -20,16 +21,21 @@ func CreateWorkflow(ctx context.Context, registry *registry.Registry, organizati
...
+        ID:             uuid.New(),
+        OrganizationID: uuid.MustParse(organizationID),
+        Name:           pbWorkflow.Name,
+        Description:    pbWorkflow.Description,
+        CreatedAt:      &now,
+        UpdatedAt:      &now,
+        Edges:          datatypes.NewJSONSlice(edges),
+    }
+
+    if userID, ok := authentication.GetUserIdFromMetadata(ctx); ok {
```
lucaspin: If no userID is set, we shouldn't create the workflow, so I think this check should be at the top of the function, and return an error if for some reason, no user ID is set

api/swagger/superplane.swagger.json
```diff
@@ -7191,6 +7191,14 @@
             "type": "object",
             "$ref": "#/definitions/ComponentsEdge"
           }
+        },
+        "createdBy": {
```
lucaspin, line 7206: What if we had a `createdBy` field was an object with id/name fields, instead of having `createdBy` and `createdByName` fields. In the proto, this could be a UserRef message that would be reused across all the objects that return user information

PR #661, opened 2025-11-02
fix: Don't remount the canvas after a save

### LINKED ISSUE #646: Don't refresh page on Save
1. Add new component on Canvas
2. Click Save
3. Page gets refreshed

markoa:
Fixes #646.

The page wasn't really refreshing but the canvas was being "remounted" so it appeared like a refresh.


markoa:
fyi @lucaspin since you wrote the original logic, just in case this change is going against something that I haven't noticed

PR #675, opened 2025-11-02
feat: HTTP payload is optional

### LINKED ISSUE #637: Payload for HTTP component should not be required
### Changes

- Disable the requirement for the payload component in its definition
- Allow sending null and array values as json payloads

markoa:
Resolves #637.

### DIFF
--- a/pkg/components/http/http.go
+++ b/pkg/components/http/http.go
@@ -85,6 +85,9 @@ func (e *HTTP) Configuration() []components.ConfigurationField {
 			Type:     components.FieldTypeObject,
 			Label:    "Payload",
 			Required: false,
+			VisibilityConditions: []components.VisibilityCondition{
+				{Field: "method", Values: []string{"POST", "PUT"}},
+			},
 		},
 		{
 			Name:     "headers",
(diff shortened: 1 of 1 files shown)


pkg/components/http/http.go
```diff
@@ -84,8 +84,7 @@ func (e *HTTP) Configuration() []components.ConfigurationField {
 			Name:     "payload",
```
lucaspin, line 84: We should also only show this field when POST/PUT method is chosen. You can do this with the usage of visibility conditions. Check how we do it on the schedule trigger: https://github.com/superplanehq/superplane/blob/develop/pkg/triggers/schedule/schedule.go#L94

pkg/components/http/http.go
```diff
@@ -132,7 +131,13 @@ func (e *HTTP) Execute(ctx components.ExecutionContext) error {
 		return err
 	}
 
-	body, err := e.getBody(spec)
+	configMap, ok := ctx.Configuration.(map[string]any)
```
lucaspin: At this point, we already have Spec parsed, so we probably don't need to touch `ctx.Configuration` again. Also, all this logic could be inside of `getBody()`

pkg/components/http/http.go
```diff
@@ -179,12 +184,20 @@ func (e *HTTP) Execute(ctx components.ExecutionContext) error {
 	})
 }
 
-func (e *HTTP) getBody(spec Spec) (io.Reader, error) {
-	if spec.Method == http.MethodGet {
+func (e *HTTP) getBody(method string, payload any, payloadProvided bool) (io.Reader, error) {
```
lucaspin: payloadProvided boolean parameter is extraneous. We can just check payload directly, probably like:

```golang
func (e *HTTP getBody(method string, payload any) (io.Reader, error) {
  if method == http.MethodGet || payload == nil {
    return nil, nil
  }

  bodyData, err := json.Marshal(spec.Payload)
  if err != nil {
    return nil, fmt.Errorf("failed to marshal request body: %w", err)
  }

  return bytes.NewReader(bodyData), nil
}

markoa:
Thanks for the feedback @lucaspin. I've updated the code.

PR #763, opened 2025-11-04
feat: semaphore component

lucaspin:
Workflow status is being monitored through polling. Ideal solution should use webhook-based updates, but this makes it a lot simpler for now.


web_src/src/pages/workflowv2/index.tsx
```diff
@@ -1159,6 +1160,8 @@ function prepareComponentNode(
       return prepareFilterNode(nodes, node, components, nodeExecutionsMap);
     case "http":
       return prepareHttpNode(node, components, nodeExecutionsMap);
+    case "semaphore":
+      return prepareSemaphoreNode(node, components, nodeExecutionsMap);
```
shiroyasha: You probably need nodeQueueItemsMap as well here.

PR #764, opened 2025-11-04
feat: time gate component

forestileao:
image

### DIFF
--- a/pkg/components/timegate/time_gate.go
+++ b/pkg/components/timegate/time_gate.go
@@ -1,4 +1,4 @@
-package time_gate
+package timegate
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


web_src/src/pages/workflowv2/index.tsx
```diff
@@ -1161,6 +1161,8 @@ function prepareComponentNode(
       return prepareHttpNode(node, components, nodeExecutionsMap);
     case "wait":
       return prepareWaitNode(node, components, nodeExecutionsMap, nodeQueueItemsMap);
+    case "time_gate":
+      return prepareTimeGateNode(node, components, nodeExecutionsMap);
```
shiroyasha, line 1165: You probably need the `nodeQueueItemsMap` here as well.

Issue #522, opened 2025-10-31
Merge component


AleksandarCole:
(edits by igor)

Steps to get there:

- [x] New component added to UI, called "merge"
- [x] Engine prep work is done (see dependent issues)
- [x] Offload queue processing to Components
- [x] Support merge behavior on the "root" level

---

Details from the sync (Lucas x Igor):

```
1. Blueprint are a component and implement the Component interface
   - Users should not see them listed in the sidebar
   - Engine should not use this interface yet

2. Move blueprint "execution" to the "execute" method on the compenent
   - Implement executions via the blueprint interface
   - Update the engine

3. Continue with shifting queue processing from worker -> components
   - Just rebase with "develop" and it should work

4. Implement the "merge behaviour" on the workflow level
   - Keep it as "noop" on the blueprint level

5. "Explode" the blueprint for every node
   - Create workflow nodes for every subcomponent

6. Implement queue for blueprint level nodes

7. Implement the "merge behaviour" on non-root level
```

---

(original)

We need to support fan in both as a stand alone component and as an option for `custom component`.

Fan in description:
This component listens to 2 or more components, it waits for events from all of them, and emits an event only when all of them arrive.

AleksandarCole:
- Introduce timeout: 
  - set time in `x` `seconds/minutes/hours`
  - select behaviour if they don't arrive in time - emit/drop/emit through timeout channel
  - if `emit through timeout` is selected new output channel exists
  - Show these parameters in the component UI

shiroyasha:
The core feature is shipped. Logging various upgrades:

- https://github.com/superplanehq/superplane/issues/838
- https://github.com/superplanehq/superplane/issues/840
- https://github.com/superplanehq/superplane/issues/839
- https://github.com/superplanehq/superplane/issues/841

PR #781, opened 2025-11-05
chore: Shift the queue processing from the Queue Worker to Components

shiroyasha:
Previously, the QueueWorker was pop-ing the first item from the queue, and creating the execution item.

Now, it is the responsibility of the Component to manage its queue.

related to https://github.com/superplanehq/superplane/issues/522

### DIFF
--- a/pkg/workers/workflow_node_queue_worker_test.go
+++ b/pkg/workers/workflow_node_queue_worker_test.go
@@ -207,7 +207,7 @@ func Test__WorkflowNodeQueueWorker_PicksOldestQueueItem(t *testing.T) {
 		r.Organization.ID,
 		[]models.WorkflowNode{
 			{NodeID: triggerNode, Type: models.NodeTypeTrigger},
-			{NodeID: componentNode, Type: models.NodeTypeComponent},
+			{NodeID: componentNode, Type: models.NodeTypeComponent, Ref: datatypes.NewJSONType(models.NodeRef{Component: &models.ComponentRef{Name: "noop"}})},
 		},
 		[]models.Edge{
 			{SourceID: triggerNode, TargetID: componentNode, Channel: "default"},
@@ -298,7 +298,7 @@ func Test__WorkflowNodeQueueWorker_EmptyQueue(t *testing.T) {
 		r.Organization.ID,
 		[]models.WorkflowNode{
 			{NodeID: triggerNode, Type: models.NodeTypeTrigger},
-			{NodeID: componentNode, Type: models.NodeTypeComponent},
+			{NodeID: componentNode, Type: models.NodeTypeComponent, Ref: datatypes.NewJSONType(models.NodeRef{Component: &models.ComponentRef{Name: "noop"}})},
 		},
 		[]models.Edge{
 			{SourceID: triggerNode, TargetID: componentNode, Channel: "default"},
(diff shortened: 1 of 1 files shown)


pkg/workers/workflow_node_queue_worker.go
```diff
@@ -81,52 +92,117 @@ func (w *WorkflowNodeQueueWorker) processNode(tx *gorm.DB, node *models.Workflow
 		return err
 	}
 
-	w.log("De-queueing item %s for node=%s workflow=%s", queueItem.ID, node.NodeID, node.WorkflowID)
```
shiroyasha, line 84: Note: Re-add this message. We lost it in the new implementation.

pkg/workers/workflow_node_queue_worker.go
```diff
@@ -11,18 +11,29 @@ import (
...
+	"github.com/superplanehq/superplane/pkg/registry"
 	"github.com/superplanehq/superplane/pkg/workers/contexts"
 )
 
+func asMap(v any) map[string]any {
+	if m, ok := v.(map[string]any); ok {
+		return m
+	}
+	return map[string]any{}
+}
```
shiroyasha: We don't need this anymore.

pkg/workers/workflow_node_queue_worker.go
```diff
@@ -81,52 +92,117 @@ func (w *WorkflowNodeQueueWorker) processNode(tx *gorm.DB, node *models.Workflow
...
+		now := time.Now()
+
+		execution := models.WorkflowNodeExecution{
+			WorkflowID:          queueItem.WorkflowID,
+			NodeID:              node.NodeID,
+			RootEventID:         queueItem.RootEventID,
+			EventID:             event.ID,
+			PreviousExecutionID: event.ExecutionID,
+			State:               models.WorkflowNodeExecutionStatePending,
+			Configuration:       datatypes.NewJSONType(asMap(config)),
```
shiroyasha: No need to call `asMap` here.

pkg/workers/workflow_node_queue_worker.go
```diff
@@ -81,52 +85,125 @@ func (w *WorkflowNodeQueueWorker) processNode(tx *gorm.DB, node *models.Workflow
...
+
+	ctx.DequeueItem = func() error {
+		return queueItem.Delete(tx)
+	}
+
+	ctx.UpdateNodeState = func(state string) error {
+		return node.UpdateState(tx, state)
+	}
+
+	ctx.UpdateNodeState = func(state string) error {
```
lucaspin: UpdateNodeState is defined twice

pkg/workers/workflow_node_queue_worker.go
```diff
@@ -81,52 +85,125 @@ func (w *WorkflowNodeQueueWorker) processNode(tx *gorm.DB, node *models.Workflow
...
-	err = queueItem.Delete(tx)
-	if err != nil {
-		return err
-	}
+    if w.registry == nil {
+        return nil, fmt.Errorf("registry is not initialized")
+    }
+
+    if ref.Component == nil || ref.Component.Name == "" {
+        return nil, fmt.Errorf("node %s has no component reference", node.NodeID)
```
lucaspin: I think we need to handle blueprint nodes here too, right?

PR #807, opened 2025-11-06
chore: Merge queue processing

shiroyasha:


shiroyasha:
Introduced a `workflow_node_execution_lookup_keys` cc @lucaspin

db/migrations/20251106120000_add-workflow-node-execution-lookup-keys.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS workflow_node_execution_lookup_keys (
```
lucaspin: Should we add this as part of this change? For now, we don't need any special keys, right? Maybe we can add this later, when we actually need it?

pkg/components/component.go
```diff
@@ -209,9 +209,22 @@ type ProcessQueueContext struct {
...
+	CountIncomingEdges func() (int, error)
+
+	// FinishExecution marks the execution as finished with the provided outputs.
+	FinishExecution func(executionID string, outputs map[string][]any) error
+
+	// FindExecutionByLookupKey returns an execution id registered under key=value for this node.
+	FindExecutionByLookupKey func(key string, value any) (string, bool, error)
+
+	// SetExecutionLookupKVPair registers a lookup key/value pair for the given execution id.
+	SetExecutionLookupKVPair func(executionID string, key string, value string) error
```
lucaspin: I think this would be cleaner if we had sub-contexts for specific things, like we do in the [ExecutionContext](https://github.com/superplanehq/superplane/blob/develop/pkg/components/component.go#L102). What do you think?

pkg/workers/workflow_node_queue_worker.go
```diff
@@ -152,8 +153,28 @@ func (w *WorkflowNodeQueueWorker) buildProcessQueueContext(tx *gorm.DB, node *mo
 		Input:         event.Data.Data(),
 	}
 
-	ctx.CreateExecution = func() error {
-		now := time.Now()
+    ctx.FindExecutionByLookupKey = func(key string, value any) (string, bool, error) {
```
lucaspin: Another reason for having clear and separated contexts for each concern is it would make this code a little bit cleaner. We would keep all the sub-contexts in [pkg/workers/contexts](https://github.com/superplanehq/superplane/tree/develop/pkg/workers/contexts), and we can just build the ProcessQueueContext like we build the ExecutionContext in [WorkflowNodeExecutor](https://github.com/superplanehq/superplane/blob/develop/pkg/workers/workflow_node_executor.go#L167)

pkg/components/merge/merge.go
lucaspin, line 47: The Execute() needs to change too, right?

pkg/workers/workflow_node_queue_worker.go
shiroyasha: We can split it, but the construction needs to live outside of the worker so we can write tests for the components and pass them an execution/queueprocessing context, without the need to spin up a whole worker.

lucaspin: We could move the contexts from `pkg/workers/contexts` to `pkg/contexts`, and have a `contexts.NewExecutionContext()` that both the worker and the component tests can use

pkg/components/merge/merge.go
```diff
@@ -48,7 +50,105 @@ func (m *Merge) Execute(ctx components.ExecutionContext) error {
 }
 
 func (m *Merge) ProcessQueueItem(ctx components.ProcessQueueContext) error {
-	// This will become 10x more complex when we add support for
-	// merging strategies, timeouts, etc.
-	return ctx.DefaultProcessing()
+	merge_group := ctx.RootEventID
```
lucaspin: Linter should've picked this snake_case variable here :)

shiroyasha: Heh, it is a pain moving between snake case and non-snake case languages.

shiroyasha: Wanted to create an issue, but
All is good, it did pick it up 👌 

CleanShot 2025-11-08 at 13 05 36@2x

PR #808, opened 2025-11-06
fix: use transactional casbin enforcer / adapter

lucaspin:
Updates authorization/service.go to use the new transaction enforcer (see https://github.com/casbin/casbin/pull/1552), and  transactional adapter (see https://github.com/casbin/gorm-adapter/pull/273).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/superplanehq/superplane
 go 1.25
 
 require (
-	github.com/casbin/casbin/v2 v2.132.0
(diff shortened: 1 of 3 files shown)


lucaspin:
Just leaving a note here since this change isn't super important right now, so I'll probably not look at it right now, and no one merges this.

---

There's an issue with custom roles after migrating to the transactional enforcer: after creating a custom role, all the permission checks start to return `undefined function g`. Only after a environment restart, everything goes back to working. This seems related to TransactionalEnforcer not updating the role manager after making policy changes in the transaction.

PR #817, opened 2025-11-07
feat: Indexed key/value store for node executions

shiroyasha:


db/migrations/20251107100000_rename-workflow-node-execution-lookup-keys-to-kvs.up.sql
```diff
@@ -0,0 +1,16 @@
+-- Rename table and index: workflow_node_execution_lookup_keys -> workflow_node_execution_kvs
+ALTER TABLE IF EXISTS workflow_node_execution_lookup_keys
+  RENAME TO workflow_node_execution_kvs;
+
+DO $$
```
lucaspin: Why not just `ALTER INDEX idx_workflow_node_execution_lookup_keys_ekv RENAME TO idx_workflow_node_execution_kvs_ekv`?

PR #825, opened 2025-11-07
feat: add exeuction chain

forestileao:
image


web_src/src/ui/CanvasPage/index.tsx
```diff
@@ -29,6 +29,7 @@ export interface TabData {
   current?: Record<string, any>;
   root?: Record<string, any>;
   payload?: any;
+  executionChain?: Array<{ name: string; completed: boolean; children?: Array<{ name: string; completed: boolean }> }>;
```
shiroyasha: Create a named type for this pls 🙏

forestileao: Saw this now, doing another PR

Issue #818, opened 2025-11-07
Avoid duplicating data in the database for already emitted events


forestileao:
Avoid duplicating data in the database for already emitted events: 
if, noop, and filter should emit empty events

PR #835, opened 2025-11-08
refactor: if, noop and filter should emmit empty events

forestileao:
Related to: https://github.com/superplanehq/superplane/issues/818

### DIFF
--- a/pkg/components/filter/filter_test.go
+++ b/pkg/components/filter/filter_test.go
@@ -30,29 +30,34 @@ func (m *MockExecutionStateContext) IsFinished() bool {
 
 func TestFilter_Execute_EmitsEmptyEvents(t *testing.T) {
 	tests := []struct {
-		name          string
-		configuration map[string]any
-		inputData     any
+		name            string
+		configuration   map[string]any
+		inputData       any
+		expectedOutputs map[string][]any
 	}{
 		{
-			name:          "filter with true condition emits empty event",
-			configuration: map[string]any{"expression": "true"},
-			inputData:     map[string]any{"test": "value"},
+			name:            "filter with true condition emits empty event",
+			configuration:   map[string]any{"expression": "true"},
+			inputData:       map[string]any{"test": "value"},
+			expectedOutputs: map[string][]any{"default": {make(map[string]any)}},
 		},
 		{
-			name:          "filter with false condition emits empty event",
-			configuration: map[string]any{"expression": "false"},
-			inputData:     map[string]any{"test": "value"},
+			name:            "filter with false condition emits empty event",
+			configuration:   map[string]any{"expression": "false"},
+			inputData:       map[string]any{"test": "value"},
+			expectedOutputs: map[string][]any{},
 		},
 		{
-			name:          "filter with complex true condition emits empty event",
-			configuration: map[string]any{"expression": "$.test == 'value'"},
-			inputData:     map[string]any{"test": "value"},
+			name:            "filter with complex true condition emits empty event",
+			configuration:   map[string]any{"expression": "$.test == 'value'"},
+			inputData:       map[string]any{"test": "value"},
+			expectedOutputs: map[string][]any{"default": {make(map[string]any)}},
 		},
 		{
-			name:          "filter with complex false condition emits empty event",
-			configuration: map[string]any{"expression": "$.test == 'different'"},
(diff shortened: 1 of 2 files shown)


pkg/components/filter/filter.go
```diff
@@ -86,17 +86,12 @@ func (f *Filter) Execute(ctx components.ExecutionContext) error {
...
 	if !ok {
 		return fmt.Errorf("expression must evaluate to boolean, got %T", output)
 	}
 
-	outputs := map[string][]any{}
-	if matches {
-		outputs[components.DefaultOutputChannel.Name] = []any{ctx.Data}
-	}
-
-	return ctx.ExecutionStateContext.Pass(outputs)
```
shiroyasha, line 99: This is changing the behavior as well, right?

Previously, filter didn't emit anything if it failed. 
Now it always emits.

This kinda breaks the behavior of the filter? Turning it into noop?

pkg/components/filter/filter.go
```diff
@@ -86,17 +86,12 @@ func (f *Filter) Execute(ctx components.ExecutionContext) error {
 		return fmt.Errorf("expression evaluation failed: %w", err)
 	}
 
-	matches, ok := output.(bool)
+	_, ok := output.(bool)
```
lucaspin: We still need to check the result of the expression, right?

pkg/components/if/if.go
```diff
@@ -90,23 +90,12 @@ func (f *If) Execute(ctx components.ExecutionContext) error {
 		return fmt.Errorf("expression evaluation failed: %w", err)
 	}
 
-	matches, ok := output.(bool)
+	_, ok := output.(bool)
 	if !ok {
 		return fmt.Errorf("expression must evaluate to boolean, got %T", output)
 	}
 
-	var outputs map[string][]any
```
lucaspin, line 98: We still need to use the true/false channels, right?

pkg/components/filter/filter.go
```diff
@@ -86,17 +86,12 @@ func (f *Filter) Execute(ctx components.ExecutionContext) error {
...
 	if !ok {
 		return fmt.Errorf("expression must evaluate to boolean, got %T", output)
 	}
 
-	outputs := map[string][]any{}
-	if matches {
-		outputs[components.DefaultOutputChannel.Name] = []any{ctx.Data}
```
lucaspin, line 96: We also stopped emitting on a specific channel. Here, we should still use the default channel

forestileao:
@lucaspin @shiroyasha Thanks for the review, the PR was breaking the logic. Now it should be good. I also noticed that we need to use `ChannelNameTrue: {make(map[string]any)}` instead of `ChannelNameTrue: {}` or the outputs would not serialize
image
image
image

pkg/components/filter/filter_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			filter := &Filter{}
+
+			mockExecStateCtx := &MockExecutionStateContext{}
+
+			if tt.configuration["expression"] == "true" || (tt.configuration["expression"] == "$.test == 'value'" && tt.inputData.(map[string]any)["test"] == "value") {
```
lucaspin: These mock expectations should be on the test case struct above

pkg/components/if/if_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			ifComponent := &If{}
+
+			mockExecStateCtx := &MockExecutionStateContext{}
+
+			if tt.configuration["expression"] == "true" || (tt.configuration["expression"] == "$.test == 'value'" && tt.inputData.(map[string]any)["test"] == "value") {
+				mockExecStateCtx.On("Pass", map[string][]any{"true": {make(map[string]any)}}).Return(nil)
```
lucaspin: Same comment about the mock expectations being in the wrong place

pkg/components/filter/filter_test.go
forestileao: Addressed

pkg/components/if/if_test.go
forestileao: Addressed

Issue #939, opened 2025-11-13
Re-emmit doesn't work


AleksandarCole:
https://github.com/user-attachments/assets/e76b630e-3985-4ed3-8b63-5f76a20e958c

forestileao:
As discussed, removed the reemit node execution event service. If we want a Requeue service we need to discuss how it should work with special components like "merge".

By now, implemented only for triggers

PR #955, opened 2025-11-13
feat: add reemit service in backend

forestileao:
Resolves: https://github.com/superplanehq/superplane/issues/939


api/swagger/superplane.swagger.json
```diff
@@ -2406,6 +2406,52 @@
         ]
       }
     },
+    "/api/v1/workflows/{workflowId}/executions/{executionId}/reemit": {
```
lucaspin: Why not use the already existing `POST /workflows/{workflow_id}/nodes/{node_id}/events?` endpoint for this?

forestileao: As discussed, removed the reemit node execution event service. If we want a Requeue service we need to discuss how it should work with special components like "merge".

By now, implemented only for triggers

PR #959, opened 2025-11-13
Design fixes

ropsii:


web_src/src/ui/approval/index.tsx
```diff
@@ -176,7 +176,7 @@ export const Approval: React.FC<ApprovalProps> = ({
 
   return (
     <SelectionWrapper selected={selected}>
-      <div className="flex flex-col border-1 border-border rounded-md w-[30rem] bg-white">
+      <div className="flex flex-col border-1 border-border rounded-md w-[26rem] bg-white overflow-hidden">
```
forestileao, line 179: Don't add overflow hidden to the components, it will hide the options dropdown if the component is too small
image

ropsii: I know, Igor told me the issue with adding that class. I'll try to figure it out, but we might decide to move that menu above the component container if I don't find another way to contain component header background color.

shiroyasha:
@ropsii this is an old branch that we don't plan to merge anymore, right? 🤔

ropsii:
@shiroyasha Yes. Closing it now.

PR #1057, opened 2025-11-17
fix: use auto delete queue consumers in event distributer

forestileao:
Set auto delete queue consumers in event distributer


forestileao:
Required to https://github.com/renderedtext/go-tackle/pull/12 be merged. So I can use a `go get` synced with the main branch,

PR #1072, opened 2025-11-18
Sidebar design polish

ropsii:


db/structure.sql
forestileao, line 1: Do not delete structure.sql file. 
we use that for checking the current structure of the database

ropsii: hm... not even sure how/why did I do that.. just pulled the latest version of the repo and worked from there

ropsii: restored it.

PR #1106, opened 2025-11-19
feat: add workflow hard deletion worker after soft deletion

forestileao:
# Changes
- Updated delete_workflow to be a soft delete action
- Created a hard deletion worker to look at the workflow and its resources to delete it all

## How it was tested:
- Created 7 nodes with schedule, merge and wait components with events in queue.
- Soft deleted it and got the following log:
image

### DIFF
--- a/pkg/models/workflow.go
+++ b/pkg/models/workflow.go
@@ -116,6 +116,24 @@ func FindWorkflowInTransaction(tx *gorm.DB, orgID, id uuid.UUID) (*Workflow, err
 	return &workflow, nil
 }
 
+func FindWorkflowWithoutOrgScope(id uuid.UUID) (*Workflow, error) {
+	return FindWorkflowWithoutOrgScopeInTransaction(database.Conn(), id)
+}
+
+func FindWorkflowWithoutOrgScopeInTransaction(tx *gorm.DB, id uuid.UUID) (*Workflow, error) {
+	var workflow Workflow
+	err := tx.
+		Where("id = ?", id).
+		First(&workflow).
+		Error
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &workflow, nil
+}
+
 func FindUnscopedWorkflow(id uuid.UUID) (*Workflow, error) {
 	return FindUnscopedWorkflowInTransaction(database.Conn(), id)
 }
--- a/pkg/workers/contexts/process_queue_context.go
+++ b/pkg/workers/contexts/process_queue_context.go
@@ -167,7 +167,7 @@ func BuildProcessQueueContext(tx *gorm.DB, node *models.WorkflowNode, queueItem 
 			// Fallthrough to workflow-level counting if blueprint id missing
 		}
 
-		wf, err := models.FindUnscopedWorkflowInTransaction(tx, node.WorkflowID)
+		wf, err := models.FindWorkflowWithoutOrgScopeInTransaction(tx, node.WorkflowID)
 		if err != nil {
 			return 0, err
 		}
@@ -213,7 +213,7 @@ func BuildProcessQueueContext(tx *gorm.DB, node *models.WorkflowNode, queueItem 
 			}
 		}
 
-		wf, err := models.FindUnscopedWorkflowInTransaction(tx, node.WorkflowID)
+		wf, err := models.FindWorkflowWithoutOrgScopeInTransaction(tx, node.WorkflowID)
 		if err != nil {
 			return 0, err
 		}
--- a/pkg/workers/workflow_event_router.go
+++ b/pkg/workers/workflow_event_router.go
@@ -113,7 +113,7 @@ func (w *WorkflowEventRouter) LockAndProcessEvent(logger *log.Entry, event model
 }
 
 func (w *WorkflowEventRouter) processEvent(tx *gorm.DB, logger *log.Entry, event *models.WorkflowEvent) ([]models.WorkflowNodeQueueItem, *models.WorkflowNodeExecution, error) {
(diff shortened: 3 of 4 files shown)


pkg/workers/workflow_cleanup_worker.go
```diff
@@ -0,0 +1,110 @@
...
+					}
+				}(workflow)
+			}
+		}
+	}
+}
+
+func (w *WorkflowCleanupWorker) LockAndProcessWorkflow(workflow models.Workflow) error {
+	return database.Conn().Transaction(func(tx *gorm.DB) error {
+		w.logger.Infof("Processing deleted workflow %s", workflow.ID)
```
lucaspin: It seems like we are not locking the workflow here?

pkg/workers/workflow_cleanup_worker.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	nodes, err := models.FindWorkflowNodesInTransaction(tx, workflow.ID)
+	if err != nil {
+		return err
+	}
+
+	w.logger.Infof("Found %d nodes to delete for workflow %s", len(nodes), workflow.ID)
+
+	for _, node := range nodes {
```
lucaspin: I think it might be better to delete the workflow_node records after deleting the events / executions / queue items / requests / kvs, right? Also, should we use a transaction here?

pkg/models/workflow_node.go
```diff
@@ -74,8 +74,10 @@ func ListWorkflowNodesReady() ([]WorkflowNode, error) {
 	err := database.Conn().
 		Distinct().
 		Joins("JOIN workflow_node_queue_items ON workflow_nodes.workflow_id = workflow_node_queue_items.workflow_id AND workflow_nodes.node_id = workflow_node_queue_items.node_id").
+		Joins("JOIN workflows ON workflow_nodes.workflow_id = workflows.id").
```
lucaspin, line 77: Not so sure how worse this extra JOIN make this query. We should keep a closer look at this after merging

pkg/models/workflow_event.go
```diff
@@ -168,7 +168,9 @@ func CountRootWorkflowEvents(workflowID uuid.UUID) (int64, error) {
 func ListPendingWorkflowEvents() ([]WorkflowEvent, error) {
 	var events []WorkflowEvent
 	err := database.Conn().
-		Where("state = ?", WorkflowEventStatePending).
+		Joins("JOIN workflows ON workflow_events.workflow_id = workflows.id").
```
lucaspin, line 171: Same comment about the extra JOIN, let's monitor this query after the merge

pkg/workers/workflow_cleanup_worker.go
forestileao: I think we should use a transaction, if the server shutdown everything will be only partially deleted

pkg/workers/workflow_cleanup_worker.go
forestileao: added lock, with FOR UPDATE and SKIP LOCKED

pkg/workers/workflow_cleanup_worker.go
```diff
@@ -0,0 +1,122 @@
...
+			telemetry.RecordWorkflowCleanupWorkerTickDuration(context.Background(), time.Since(tickStart))
+		}
+	}
+}
+
+func (w *WorkflowCleanupWorker) LockAndProcessWorkflow(workflow models.Workflow) error {
+	return database.Conn().Transaction(func(tx *gorm.DB) error {
+		lockedWorkflow, err := models.LockWorkflow(tx, workflow.ID)
+		if err != nil {
+			w.logger.Info("Workflow already being processed - skipping")
```
lucaspin: Add the workflow ID to the log message

pkg/workers/workflow_cleanup_worker_test.go
```diff
@@ -0,0 +1,381 @@
...
+	require.NoError(t, err)
+	assert.Len(t, nodes, 2)
+	support.VerifyWorkflowEventsCount(t, workflow.ID, 2)
+	support.VerifyWorkflowNodeExecutionsCount(t, workflow.ID, 1)
+	support.VerifyWorkflowNodeQueueCount(t, workflow.ID, 1)
+
+	// Verify KV and request exist
+	var kvCount int64
+	database.Conn().Model(&models.WorkflowNodeExecutionKV{}).Where("workflow_id = ?", workflow.ID).Count(&kvCount)
+	assert.Equal(t, int64(1), kvCount)
```
lucaspin: This could be a `support.Verify*` function. Same for the node requests and the kvs. It would make this test easier to read

pkg/workers/workflow_cleanup_worker_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+	// Verify workflow is soft deleted
+	_, err = models.FindWorkflow(r.Organization.ID, workflow.ID)
+	assert.ErrorIs(t, err, gorm.ErrRecordNotFound)
+
+	//
+	// Fetch the updated workflow with deleted_at set
+	//
+	var deletedWorkflow models.Workflow
+	err = database.Conn().Unscoped().Where("id = ?", workflow.ID).First(&deletedWorkflow).Error
```
lucaspin: models.FindUnscopedWorkflow()

forestileao: This Unscoped method is just to say it is unscoped from the org. I will rename the original one to something else and create an Unscoped that uses the gorm method.

PR #1148, opened 2025-11-25
fix: conflict when user changes provider email

forestileao:
## Problem:

When users change their email in OAuth providers (GitHub/Google), authentication fails with 500 errors because the system tries to find accounts by new email but they were created with old email.

## Solution:
  - Find accounts by provider ID first, then by email (fallback)
  - Update account, users, AND specific provider when email changes
  - Proper transaction rollback handling
  - Return 403 (not 500) when signup is blocked

## Key Changes:
  - FindAccountByProvider() - lookup by provider info
  - UpdateEmailForProvider() - provider-specific email updates
  - FindOrCreateAccountForProvider() - provider-aware account handling

## Result:
  ✅ Users can sign in after changing OAuth provider emails
  ✅ Only updates the specific provider that changed and original accounts

### DIFF
--- a/pkg/public/server_auth_test.go
+++ b/pkg/public/server_auth_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/superplanehq/superplane/pkg/crypto"
-	"github.com/superplanehq/superplane/pkg/database"
 	"github.com/superplanehq/superplane/pkg/jwt"
 	"github.com/superplanehq/superplane/pkg/models"
 	"github.com/superplanehq/superplane/test/support"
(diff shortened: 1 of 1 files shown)


pkg/authentication/authentication_test.go
```diff
@@ -0,0 +1,231 @@
...
+			Provider: "github",
+		}
+
+		resultAccount, err := handler.FindOrCreateAccountForProvider(gothUser)
+		require.Error(t, err)
+		assert.Equal(t, SignupDisabledError, err.Error())
+		assert.Nil(t, resultAccount)
+	})
+
+	t.Run("should handle email update failure gracefully", func(t *testing.T) {
```
lucaspin: What is this testing exactly?

pkg/authentication/authentication_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+		resultAccount, err := handler.FindOrCreateAccountForProvider(gothUser)
+		require.NoError(t, err)
+
+		assert.Equal(t, account.ID, resultAccount.ID)
+		assert.Equal(t, originalEmail, resultAccount.Email)
+	})
+}
+
+func TestHandler_findOrCreateAccount_Legacy(t *testing.T) {
```
lucaspin: Legacy?

pkg/models/account_test.go
```diff
@@ -0,0 +1,253 @@
...
+	t.Run("should handle transaction rollback on failure", func(t *testing.T) {
+		account, err := CreateAccount("Test User", "test2@example.com")
+		require.NoError(t, err)
+		originalEmail := account.Email
+
+		conflictEmail := "conflict@example.com"
+		_, err = CreateAccount("Conflict User", conflictEmail)
+		require.NoError(t, err)
+
+		err = account.UpdateEmail(conflictEmail)
```
lucaspin, line 164: Not sure I understand this. Are you updating the account to use the same email it already has?

If that's the case, I wouldn't expect an error. The error should only happen if you are trying to update another different accounts, right?

pkg/public/server_auth_test.go
```diff
@@ -157,3 +159,91 @@ func TestServer_ProviderConfiguration(t *testing.T) {
 	assert.Contains(t, providers, "google")
 	assert.Len(t, providers, 2)
 }
+
+func TestServer_ProviderEmailChangeIntegration(t *testing.T) {
```
lucaspin: Isn't this already being tested in authentication_test.go?

pkg/models/account_test.go
forestileao: This should trigger a unique constraint error because two accounts can't have the same email. The test verifies that:
  - The update operation fails with an error
  - The original account's email remains unchanged (transaction rollback)
  - The database record also remains unchanged

pkg/authentication/authentication_test.go
forestileao: Removed it, it was from the older implementation

pkg/authentication/authentication_test.go
forestileao: Removed, it was not testing anything important at all

pkg/public/server_auth_test.go
forestileao: Removed duplicated test

Issue #1112, opened 2025-11-19
Workflow Cleanup Worker deadlocking


forestileao:
For some reason, when the Workflow Cleanup Worker runs, some workflows are hard deleted some of them are not.

Examples of workflows that are deadlocking on hard delete are:
- https://app.superplane.com/3ee1aa47-3a60-4c1f-b645-0b9859ab91f8/workflows/a1391d53-703e-4692-890c-b4331d46b630
- https://app.superplane.com/3ee1aa47-3a60-4c1f-b645-0b9859ab91f8/workflows/50e48f1c-c84a-4b72-a4c2-de0fd5b93f86

Things we noted:
- Empty workflows are hardeleted with success
- New workflows with data in them and no composite component are also deleted with success
- Workflows with composite components like the ones listed above are not being hard deleted

forestileao:
### Potential Problem
Deleting workflows with 1000+ events takes a long time due to cascading foreign keys causing chain reactions.

### Most likely Cause
`workflow_node_executions` has **self-referencing CASCADE foreign keys**:
- `parent_execution_id` → `workflow_node_executions.id` (ON DELETE CASCADE)
- `previous_execution_id` → `workflow_node_executions.id` (ON DELETE CASCADE)

When deleting 1 event:
1. Cascades to 10 executions
2. Each cascades to 5 more (parent/previous links)
3. Each of those cascades to 3 more
4. Creates recursive cascade chain = **O(n²) or worse**

Additionally:
- Wrong deletion order (deleting parents before children)
- Missing index on `workflow_events.workflow_id`

### Solution
1. **Remove CASCADE from foreign keys** (change to RESTRICT)
2. **Delete in correct order**: children → parents
3. **Add batching** for large datasets (5000 rows/batch)
4. **Add index**: `CREATE INDEX idx_workflow_events_workflow_id ON workflow_events (workflow_id)`

PR #1180, opened 2025-11-29
fix: add required foreign keys for node_id related data

forestileao:
Look at: https://github.com/superplanehq/superplane/issues/1112#issuecomment-3571565678

Resolves: https://github.com/superplanehq/superplane/issues/1112


shiroyasha:
Correct me if I'm wrong, but when we ship this change, we will no longer be able to delete individual nodes from a canvas?

Try this:

1. Create a canvas, add two nodes, save.
2. Run several times.
3. Delete a single node.
4. Try to save.

If I understand the system correctly, previously this deleted executions and queue items from that node. But now, it will just fail.

lucaspin:
> If I understand the system correctly, previously this deleted executions and queue items from that node. But now, it will just fail.

Why would it fail? I think the only change here would be that when node is deleted, nothing related to that node is deleted on cascade anymore

shiroyasha:
The new constraint is `restrict`, right? 
When you do `node.delete()` it will raise an exception that you have queue items that reference it.

forestileao:
> If I understand the system correctly, previously this deleted executions and queue items from that node. But now, it will just fail.

Got no error at all, were you able to make it fail?

lucaspin:
@shiroyasha I think the references are to the workflow record, not to the workflow_node record, but good point. @forestileao can you check that?

forestileao:
> @shiroyasha I think the references are to the workflow record, not to the workflow_node record, but good point. @forestileao can you check that?

There is a foreign key pointing to execution key values that points to workflow_nodes. Although I was  not able to replicate any error, let me add some tests

shiroyasha:
Just tried it. Everything works 👍 
Sorry for the false alarm.

pkg/models/workflow_node.go
```diff
@@ -37,7 +37,13 @@ type WorkflowNode struct {
 }
 
 func DeleteWorkflowNode(tx *gorm.DB, node WorkflowNode) error {
-	err := tx.Delete(&node).Error
+	err := tx.Where("workflow_id = ? AND node_id = ?", node.WorkflowID, node.NodeID).
+		Delete(&WorkflowNodeExecutionKV{}).Error
```
lucaspin: It's a bit weird that we have to delete only the WorkflowNodeExecutionKV records, and not everything else here. If all the other tables are not referencing the workflow_node record, why is this one special?

forestileao: Because of this specific index here:
https://github.com/superplanehq/superplane/pull/1180/files#diff-1dd5a8f580b9615769ddc19a64a98f455d9d20eb149c443f41d9d3c797d39b12R845-R846

There is no other index or foreing key referencing workflow nodes directly

forestileao: I was not able to replciate the error while using the canvas, only in the test that I added

lucaspin: Right. So every other table does not have a foreign key to workflow_nodes, but this one has. That's inconsistent. We either have references everywhere or we don't have references anywhere. We must have a good reason to break the pattern.

So it seems, we have two options:
1. Remove the foreign key from this table too - I don't understand why it was added anyway
2. Add foreign keys in every other table - could make updating a workflow really slow if nodes with a lot of history are deleted

To me, it seems like option 1 is better. Maybe @shiroyasha can provide more details, I think he's the one who added that table.

forestileao: Yeah, adding indexes just because other table also has doesn't seem like a good solution.
IMO option 1 looks ok, but let's see @shiroyasha opinion

forestileao:
@lucaspin @shiroyasha 

This PR is updated. Now, all tables that contain node_id should have a foreign key constraint to workflow_nodes table.
After review, but before merging it, I will run the following queries in prod:

```sql
  -- Check for orphaned workflow_events:

  SELECT COUNT(*) as orphaned_events_count
  FROM workflow_events we
  WHERE NOT EXISTS (
      SELECT 1 FROM workflow_nodes wn
      WHERE wn.workflow_id = we.workflow_id
      AND wn.node_id = we.node_id
  );

 --  Check for orphaned workflow_node_executions:

  SELECT COUNT(*) as orphaned_executions_count
  FROM workflow_node_executions wne
  WHERE NOT EXISTS (
      SELECT 1 FROM workflow_nodes wn
      WHERE wn.workflow_id = wne.workflow_id
      AND wn.node_id = wne.node_id
  );

--   Check for orphaned workflow_node_execution_kvs:

  SELECT COUNT(*) as orphaned_kvs_count
  FROM workflow_node_execution_kvs wnekv
  WHERE NOT EXISTS (
      SELECT 1 FROM workflow_nodes wn
      WHERE wn.workflow_id = wnekv.workflow_id
      AND wn.node_id = wnekv.node_id
  );

 --  Check for orphaned workflow_node_queue_items:

  SELECT COUNT(*) as orphaned_queue_items_count
  FROM workflow_node_queue_items wnqi
  WHERE NOT EXISTS (
      SELECT 1 FROM workflow_nodes wn
      WHERE wn.workflow_id = wnqi.workflow_id
      AND wn.node_id = wnqi.node_id
  );

 --  Check for orphaned workflow_node_requests:

  SELECT COUNT(*) as orphaned_requests_count
  FROM workflow_node_requests wnr
  WHERE NOT EXISTS (
      SELECT 1 FROM workflow_nodes wn
      WHERE wn.workflow_id = wnr.workflow_id
      AND wn.node_id = wnr.node_id
  );
```

db/migrations/20251128222706_remove-cascade-from-workflow-foreign-keys.up.sql
```diff
@@ -0,0 +1,90 @@
+BEGIN;
```
lucaspin, line 1: This migration has way too many logical changes now (removing cascade, adding foreign keys to nodes, adding deleted_at to nodes, ...)

forestileao: Good point, let me split it

lucaspin:
@forestileao I think the queries above will fail, because you are using `workflow_nodes.deleted_at` column, but that column doesn't exist yet.

db/migrations/20251128222706_remove-cascade-from-workflow-foreign-keys.up.sql
forestileao: The migrations are split now. I can verify that structure.sql keeps the same (unless the timestamp)

pkg/grpc/actions/workflows/update_workflow_test.go
```diff
@@ -0,0 +1,89 @@
...
+		execution.ID,
+		"test-key",
+		"test-value",
+	))
+
+	existingNodes := []models.WorkflowNode{nodes[0], nodes[1]}
+	newNodes := []models.Node{{ID: "node-1"}}
+
+	err := database.Conn().Transaction(func(tx *gorm.DB) error {
+		return deleteNodes(tx, existingNodes, newNodes, workflow.ID)
```
lucaspin: We should use the UpdateWorkflow endpoint to test this behavior, and not the `deleteNodes` function directly

test/support/support.go
```diff
@@ -439,3 +486,82 @@ func VerifyWorkflowNodeRequestCount(t *testing.T, workflowID uuid.UUID, expected
...
+			}
+
+			expanded = append(expanded, internal)
+		}
+	}
+
+	return expanded, nil
+}
+
+func cloneMetadata(md map[string]any) map[string]any {
```
lucaspin: We can use maps.Clone() for this

pkg/grpc/actions/workflows/update_workflow_test.go
forestileao: Updated, now we are removing nodes using update_workflow directly

forestileao:
Recuded this PR for only adding required foreign keys for node_id related data.
The orphaned data was already deleted from the db with no problem

PR #1216, opened 2025-12-02
fix: foreign key cascades for workflow data

forestileao:
Set the following structure for foreign keys:
```
queue_items
  event_id -> events(id)      - FK - cascade set null
  root_event_id -> events(id) - FK - cascade set null

events
  workflow_id -> workflows(id)   - FK - no need for cascade because workflows are soft deleted
  execution_id -> executions(id) - FK - cascade delete

executions
  workflow_id -> workflows(id)            - FK   - no need for cascade because workflows are soft deleted
  root_event_id -> events(id)             - FK   - cascade set null
  event_id -> events(id)                  - FK   - cascade set null
  parent_execution_id -> executions(id)   - FK   - cascade delete
  previous_execution_id -> executions(id) - FK   - cascade set null

execution_kvs
  workflow_id -> workflows(id)                                   - FK - no need for cascade because workflows are soft deleted
  {workflow_id, node_id} -> workflow_nodes(workflow_id, node_id) - FK - no need for cascade because workflow nodes are soft deleted
  execution_id -> executions(id)                                 - FK - cascade delete

node_requests
  execution_id -> executions(id) - FK - cascade delete
```


lucaspin:
@forestileao for `ON DELETE SET NULL` to function correctly, the foreign key column(s) in the child table must be nullable

Issue #1277, opened 2025-12-05
Flaky test for canceling running items


shiroyasha:
https://superplanehq.semaphoreci.com/workflows/f8a0c7c2-4a8f-493e-9fbf-17a3fc1be261/summary?pipeline_id=8eac6515-9ed9-4914-ab98-7d5fa49c194c&report_id=79ed1ee4-ab67-3dab-9091-08a3b29b97c2&suite_id=6d546ff3-b43c-3b7d-b0e4-109843425c3c&test_id=da13573b-4e99-3a12-9ba1-db364ae01394

<img width="2400" height="860" alt="Image" src="https://github.com/user-attachments/assets/0933953b-47a2-495f-be9c-d6c530b5f73c" />

assigning you @forestileao as you wrote it, but let me know if you need any assistance

PR #1278, opened 2025-12-05
fix: fix timing issue on cancel running item

forestileao:
Resolves: https://github.com/superplanehq/superplane/issues/1277


forestileao:
@shiroyasha This was caused by a timing issue. Sometimes another queue item was able to become a running item, sometimes not.

PR #1444, opened 2025-12-22
fix: request worker handles requests for child/parent the same way

lucaspin:
Now that we have workflow_node records for all nodes in workflow, we don't need to treat nodes differently in WorkflowNodeRequestWorker


lucaspin:
I cannot do this because if I add new node to blueprint, the workflow_node records for instances of that blueprint won’t be added

PR #1483, opened 2025-12-30
feat: Implement webhook trigger with configurable authentication

forestileao:
## Summary

  Adds complete webhook trigger functionality with signature/bearer token authentication and proper environment variable injection.

## Key Features

  - Webhook Trigger: New webhook trigger component with HMAC signature and bearer token authentication options
  - Invoke Trigger Actions: Added gRPC endpoint to invoke custom trigger actions (e.g., reset authentication)
  - Configuration Fields: Enhanced with readonly field support for auto-generated webhook URLs

## Changes
  - New webhook trigger implementation with authentication validation
  - gRPC action endpoint for trigger operations
  - Updated protobuf definitions and generated code
  - Enhanced configuration field renderer with readonly support
  - Web UI components for webhook configuration

##  Authentication Methods

  - Signature (HMAC): SHA-256 signature validation via X-Signature-256 header
  - Bearer Token: Authorization header validation
  - None: Unsafe mode for development

### DIFF
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -116,7 +116,7 @@ func RunServer(baseURL, webhooksBaseURL, basePath string, encryptor crypto.Encry
 	blueprintService := NewBlueprintService(registry)
 	pbBlueprints.RegisterBlueprintsServer(grpcServer, blueprintService)
 
-	workflowService := NewWorkflowService(authService, registry, encryptor, baseURL+basePath)
+	workflowService := NewWorkflowService(authService, registry, encryptor, webhooksBaseURL+basePath)
 	pbWorkflows.RegisterWorkflowsServer(grpcServer, workflowService)
 
 	applicationService := NewApplicationService(encryptor, registry)


pkg/configuration/field.go
```diff
@@ -61,6 +61,11 @@ type Field struct {
 	 */
 	Sensitive bool `json:"sensitive"`
 
+	/*
+	 * Whether the field is read-only (cannot be edited by user)
+	 */
+	ReadOnly bool `json:"read_only"`
```
lucaspin: How can a configuration field by readOnly? Shouldn't we use metadata for that?

pkg/core/component.go
```diff
@@ -191,6 +191,16 @@ type RequestContext interface {
...
+
+	//
+	// Get the workflow ID for the current context
+	//
+	GetWorkflowID() string
+
+	//
+	// Get the node ID for the current context
+	//
+	GetNodeID() string
```
lucaspin: This shouldn't be here. Check ExecutionContext and ProcessQueueContext

pkg/triggers/webhook/webhook.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	if metadata.URL != "" && metadata.Authentication == config.Authentication {
+
+		return nil
+	}
+
+	var webhookURL string
+	if metadata.URL == "" {
+		if ctx.WebhookContext == nil {
```
lucaspin: No need to check for contexts inside of components, as they should always be there

pkg/triggers/webhook/webhook.go
```diff
@@ -0,0 +1,264 @@
...
+			return fmt.Errorf("webhook context is required")
+		}
+
+		webhookID, err := ctx.WebhookContext.Setup(nil)
```
lucaspin: Setup() could return the URL directly, instead of just ID, no?

pkg/configuration/field.go
forestileao: I've done that to see the base url in the same level of the fields. This is why i added the update configuration there. But i will include in the `customField` and `metadata` instead. removing that from config and deleting that new method.
image

pkg/triggers/webhook/webhook.go
```diff
@@ -0,0 +1,264 @@
...
+	err = ctx.MetadataContext.Set(metadata)
+	if err != nil {
+		return fmt.Errorf("failed to set metadata: %w", err)
+	}
+
+	return updateTriggerConfiguration(ctx, config)
+}
+
+func updateTriggerConfiguration(ctx core.TriggerContext, config Configuration) error {
+	nodeMetadataCtx, ok := ctx.MetadataContext.(*contexts.NodeMetadataContext)
```
lucaspin: We shouldn't be converting the MetadataContext interface to a concrete implementation inside of the component implementation. Also, we are using the metadata context to update the configuration, which is weird. All that being said, we shouldn't use configuration for generated values

pkg/triggers/webhook/webhook.go
forestileao: it could be the URL or the entire Webhook model, I think it makes more sense, right?
then the webhook model can have a mthod like, getURL

pkg/triggers/webhook/webhook.go
forestileao: I'll remove this method as commented on https://github.com/superplanehq/superplane/pull/1483#discussion_r2653123500

pkg/triggers/webhook/webhook.go
lucaspin: I wouldn't import `pkg/models` inside of `pkg/core`, so I'd prefer to return either the URL, or a struct defined in pkg/core for webhooks

PR #1484, opened 2025-12-30
feat: implement github.onTagCreated trigger

e-todorovski-bm:
implement github.onTagCreated trigger

### DIFF
--- a/pkg/applications/github/on_tag_created_test.go
+++ b/pkg/applications/github/on_tag_created_test.go
@@ -260,7 +260,7 @@ func Test__OnTagCreated__Setup(t *testing.T) {
 		err := trigger.Setup(core.TriggerContext{
 			AppInstallationContext: appCtx,
 			MetadataContext:        &contexts.MetadataContext{},
-			Configuration:          OnTagCreatedConfiguration{Repository: ""},
+			Configuration:          map[string]any{"repository": ""},
 		})
 
 		require.ErrorContains(t, err, "repository is required")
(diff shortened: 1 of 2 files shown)


pkg/applications/github/on_tag_created.go
```diff
@@ -0,0 +1,193 @@
...
+	"slices"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/core"
+)
+
+type OnTagCreated struct{}
+
+type OnTagCreatedMetadata struct {
```
lucaspin: We have a common struct - NodeMetadata - for node metadata, defined in common.go. We should use that instead.

pkg/applications/github/on_tag_created.go
```diff
@@ -0,0 +1,193 @@
...
+				AnyPredicateList: &configuration.AnyPredicateListTypeOptions{
+					Operators: configuration.AllPredicateOperators,
+				},
+			},
+		},
+	}
+}
+
+func (t *OnTagCreated) Setup(ctx core.TriggerContext) error {
+	var metadata OnTagCreatedMetadata
```
lucaspin: We also have a common function to verify the repository and save it on the metadata. Check ensureRepoInMetadata in common.go

lucaspin:
> The name of the event is github.tagCreated, and is the only GitHub event that doesnt fully match the name of the function in GitHub (which is simply EventType: "create"). I do this to distinguish with the other usage of that command, to create a branch

We could have a `github.refCreated` instead, which allows us to listen to both tags and branches. If we do that, we could have the default state being listening to all tag creations.

e-todorovski-bm:
> > The name of the event is github.tagCreated, and is the only GitHub event that doesnt fully match the name of the function in GitHub (which is simply EventType: "create"). I do this to distinguish with the other usage of that command, to create a branch
> 
> We could have a `github.refCreated` instead, which allows us to listen to both tags and branches. If we do that, we could have the default state being listening to all tag creations.

Thanks for the feedback, the 2 comments have been fixed.

As for this one (making this a generic component covering both the tag and branch creation), I talked with Igor and we agreed that we should keep them separate, and the latter will be tackled within https://github.com/superplanehq/superplane/issues/1500

PR #1497, opened 2025-12-31
chore: implement nullable fields in http

forestileao:
Implemented nullable fields in http component +  retries and timout options


pkg/configuration/field.go
```diff
@@ -4,19 +4,22 @@ const (
...
+	FieldTypeBool            = "boolean"
+	FieldTypeSelect          = "select"
+	FieldTypeMultiSelect     = "multi-select"
+	FieldTypeList            = "list"
+	FieldTypeObject          = "object"
+	FieldTypeTime            = "time"
+	FieldTypeDate            = "date"
+	FieldTypeDateTime        = "datetime"
+	FieldTypeTimezone        = "timezone"
+	FieldTypeTogglableString = "togglable-string"
```
lucaspin: Why have multiple toggable-* types instead of a togglable boolean field in the configuration.Field object instead?

forestileao: Your solution looks better, it will avoid repeating same type renderers. Let me change this

forestileao: @lucaspin Your recommendation was better and removed 1k lines of code.
Removed the toggable specific renderers too, just added the toggle in the existing renderers.
Thanks

Issue #1493, opened 2025-12-31
Custom run naming


AleksandarCole:
**Describe the request:**  
Add a per-trigger setting for defining run names, allowing users to set a custom run title using expressions based on the trigger payload.

___
**Describe your use-case:**  
Runs currently get default names based on the trigger type (for example, a timestamp for scheduler runs or a commit message for GitHub pushes). These defaults are acceptable, but they often look machine-generated and are not consistent across triggers.

Users want control over how runs are labeled so run history is easier to scan, filter, and discuss in a human-readable way. Each trigger should allow defining a run name template that supports expressions, so it can pull fields from the trigger payload.

Examples:
- GitHub push: `Deploy {{ $.repository.name }} @ {{ $.head_commit.id | short }} by {{ $.pusher.name }}`
- Scheduler: `Nightly release checks ({{ now() | date }})`
- Webhook: `Webhook from {{ $.headers["user-agent"] }} ({{ $.body.event_type }})`

If a custom name is not configured (or expression evaluation fails), the system should fall back to the current defaults.

forestileao:
this is not completed, I closedd by mistake

PR #1515, opened 2026-01-05
feat: add widgets interface and annotation nodes

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/1493

Creates markdown annotations on canvas:

image


web_src/src/pages/workflowv2/index.tsx
```diff
@@ -717,6 +717,28 @@ export function WorkflowPageV2() {
         if (triggerApp) {
           appName = triggerApp.name;
         }
+      } else if (node.type === "TYPE_ANNOTATION") {
```
lucaspin: Shouldn't this be TYPE_WIDGET?

forestileao: Sorry, the PR was not ready at that time you reviewed.

Now it should be. There are a lot of changes in that update_workflow action.
Now it serailizes thigns from workflow.nodes property. And we were able to remove the ListWorkflowNodes from the serialization.

In the future, this will make way easier to handle versions.

BTW,  before merigng this, I will need to update the existing data as for all canvases, to serialize the current workflow_nodes into the workflow.nodes property

PR #1574, opened 2026-01-09
feat: Github on issue commented webhook

e-todorovski-bm:
Decisions:
- No filter/regex for issue name, just one for the action (Created, Edited, Deleted)
- Not part of the "On Issue", its a separate hook/component just for the issue comments
- The main text in the event is the issue name, not the comment/commenter/action

UI Screenshots:
Screenshot 2026-01-12 at 16 23 09
Screenshot 2026-01-12 at 16 23 55

Screenshot 2026-01-09 at 12 22 35

### DIFF
--- a/pkg/applications/github/on_issue_comment.go
+++ b/pkg/applications/github/on_issue_comment.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/superplanehq/superplane/pkg/configuration"
@@ -13,8 +14,8 @@ import (
 type OnIssueComment struct{}
 
 type OnIssueCommentConfiguration struct {
-	Repository string   `json:"repository" mapstructure:"repository"`
-	Actions    []string `json:"actions" mapstructure:"actions"`
+	Repository    string `json:"repository" mapstructure:"repository"`
+	ContentFilter string `json:"contentFilter" mapstructure:"contentFilter"`
 }
 
 func (i *OnIssueComment) Name() string {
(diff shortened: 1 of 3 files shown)


lucaspin:
I would avoid using past tense verbs in the trigger name - "onIssueCommented" feels a bit weird to me, especially considering that you can filter it by created/updated/deleted actions.

Generally, trigger names should use the resource names that they reference - `issueComment` - not actions, unless the trigger itself is about an specific action.

For example, the `onTagCreated` trigger name makes sense because it is about a specific action.

e-todorovski-bm:
> I would avoid using past tense verbs in the trigger name - "onIssueCommented" feels a bit weird to me, especially considering that you can filter it by created/updated/deleted actions.
> 
> Generally, trigger names should use the resource names that they reference - `issueComment` - not actions, unless the trigger itself is about an specific action.
> 
> For example, the `onTagCreated` trigger name makes sense because it is about a specific action.

@lucaspin I agree, and changed as you described

shiroyasha:
@e-todorovski-bm look good! Two minor adjustments:

1/ lets drop the action selector and process only "created" events. Feels unutural to receive an "onIssueCommented" event when someone deleted a comment.

2/ Lets add an optinal content filter. Most of the time, we don't want to process all events (and store the data in the DB). This trigger is the most interesting when it can listen to specific comments. e.g. "if someone comments '/solve' superplane can run steps to spin up an AI agent and try to solve the issue.

The field could be called "Content Filter", should be optional, and should receive a regex pattern.

e-todorovski-bm:
> @e-todorovski-bm look good! Two minor adjustments:
> 
> 1/ lets drop the action selector and process only "created" events. Feels unutural to receive an "onIssueCommented" event when someone deleted a comment.
> 
> 2/ Lets add an optinal content filter. Most of the time, we don't want to process all events (and store the data in the DB). This trigger is the most interesting when it can listen to specific comments. e.g. "if someone comments '/solve' superplane can run steps to spin up an AI agent and try to solve the issue.
> 
> The field could be called "Content Filter", should be optional, and should receive a regex pattern.

@shiroyasha  Thanks! The requested changes are done and screenshots updated

PR #1576, opened 2026-01-09
feat: GitHub onPullRequestReviewComment trigger

e-todorovski-bm:
Decisions:
- No filter for branch/PR name, just for the action (Create/Edit/Delete)
- In the list of triggers "Pull Request" is shortened to "PR", as it took a lot of space and looked weird. I think that "PR" is abbreviation often used and there wont be any confusion.
- This handles only the inline comments, as the name of the trigger name suggests. It does not handle the comments on the PR that are not inline, like when the main comment that one leaves when the review is submitted, or the standalone comment on the PR in the "conversation" tab

UI:
Screenshot 2026-01-13 at 10 28 50
Screenshot 2026-01-13 at 10 29 10

Screenshot 2026-01-09 at 13 40 48

### DIFF
--- a/pkg/applications/github/on_pr_review_comment.go
+++ b/pkg/applications/github/on_pr_review_comment.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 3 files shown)


shiroyasha:
@e-todorovski-bm lets apply the same changes (only created action, and a content filter) here as in: https://github.com/superplanehq/superplane/pull/1574

PR #1615, opened 2026-01-13
feat: GitHub Create Release

e-todorovski-bm:
Decisions/comments:

- I did not fully mirror GitHub "create release" UI and capabilities: Skipped on "Upload binaries" - as I thought it would be too big of a task (and not sure if would be needed) for us to add a file uploader. 
- Added several options for the name (as seen in the attachment). Thought that the main usage of this might be some flow like "wait for a branch/PR with name like "hotfix" be merged, and then automatically do a patch/minor release"
- There is a bit more complex logic for the automatic increment, because if there is a draft/prerelease with higher version that the "latest", the new release is unsuccessful. I went with "10 attempts of increment" instead of getting a pool of all releases, drafts and prereleases, because theoretically there can be a "0.0.1" latest, and "0.1.0" older release, and we want to increment the latest (have continuity), and not the highest number.

Screenshot 2026-01-13 at 13 44 12
Screenshot 2026-01-13 at 13 42 21
Screenshot 2026-01-13 at 13 42 05

---

> [!NOTE]
> Adds a new GitHub `CreateRelease` component to create releases with flexible versioning and UI support.
> 
> - New backend component `CreateRelease` with config for `repository`, `versionStrategy` (manual/patch/minor/major), optional `tagName`, `name`, `draft`, `prerelease`, `generateReleaseNotes`, and `body`
> - Auto-increment logic: parses semver with optional prefix, checks for existing tags, iterates up to 10 attempts to find available version
> - Optional release notes generation via GitHub API; appends custom `body` when provided
> - Emits `github.release` on default output channel
> - Registered in GitHub app `Components()`
> - Frontend: added `create_release` mapper to display execution details (URL, ID, tag, name, draft/prerelease, timestamps, author) and wired into mapper index
> - Tests cover setup/metadata validation, configuration schema, semver regex, version increment strategies, and component interface
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ed2bd90b01cf90e50a4f135e4ccc756a655ccb2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/applications/github/create_release.go
```diff
@@ -0,0 +1,417 @@
...
+
+func (c *CreateRelease) determineTagName(ctx core.ExecutionContext, client *github.Client, owner string, config CreateReleaseConfiguration) (string, error) {
+	if config.VersionStrategy == "manual" {
+		return config.TagName, nil
+	}
+
+	//
+	// Get latest release for auto-increment
+	//
+	latestRelease, _, err := client.Repositories.GetLatestRelease(
```
lucaspin, line 260: What would happen here if the latest is not the biggest major version?

For example:

```
1. v1.0.0
2. v1.0.1
3. v2.0.0
4. v1.0.2 <- latest
```

pkg/applications/github/create_release.go
```diff
@@ -0,0 +1,417 @@
...
+	//
+	// Try to find an available version (handle case where tag might exist from draft/prerelease)
+	//
+	return c.findAvailableVersion(ctx, client, owner, config.Repository, currentTag, config.VersionStrategy)
+}
+
+func (c *CreateRelease) findAvailableVersion(ctx core.ExecutionContext, client *github.Client, owner, repo, currentTag, strategy string) (string, error) {
+	const maxAttempts = 10
+
+	for attempt := 0; attempt < maxAttempts; attempt++ {
```
lucaspin, line 280: Wouldn't it be better to list all the matching refs with this [API endpoint](https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#list-matching-references) instead?

pkg/applications/github/create_release.go
```diff
@@ -0,0 +1,417 @@
...
+		releaseRequest,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create release: %w", err)
+	}
+
+	//
+	// Emit output with release data
+	//
+	output := map[string]interface{}{
```
lucaspin: I think you can emit the returned `release` object from `client.Repositories.CreateRelease`, like the other current components

pkg/applications/github/create_release.go
e-todorovski-bm: next is v1.0.3 (or 1.1.0 or 3.0.0) depending on whether we autoincrement patch, minor or major.

it tries to increase the latest version, up to 10 increments. Patch and minor are straightforward for this example. For major: it will first try to publish with 2.0.0. When that fails, it will try with 3.0.0, 4.0.0, etc, until it either succeeds, or 10 attempts have passed.

pkg/applications/github/create_release.go
e-todorovski-bm: The releases are not refs, so that API wont work. It is possible to use "ListReleases() " instead, that fetches a whole list of releases, not just the latest (like we do with "GetLatestRelase()". ListReleases returns drafts and pre-releases as well.

I thought using "TagExists()" in the generic (most common) flow would be simpler and cheaper action, as it will most likely be used only once, and the payload is minimal, as opposed to a larger list of releases - and the code is with the same complexity (even tiny amount simpler) now.

Its not a big change anyway, if you want i can use that and check internally whether the release is already present

pkg/applications/github/create_release.go
e-todorovski-bm: My bad, will do :)

pkg/applications/github/create_release.go
lucaspin: Got it. I think it's probably ok then

PR #1651, opened 2026-01-15
feat: dash0 integration and starter components

AleksandarCole:


pkg/applications/dash0/dash0.go
```diff
@@ -0,0 +1,125 @@
...
+
+func init() {
+	registry.RegisterApplication("dash0", &Dash0{})
+}
+
+type Dash0 struct{}
+
+type Configuration struct {
+	APIToken *string `json:"apiToken"`
+	BaseURL  *string `json:"baseURL"`
```
lucaspin: Since these configuration fields are required, we don't need to use pointers for them

pkg/applications/dash0/query_prometheus.go
```diff
@@ -0,0 +1,201 @@
...
+	if spec.Type == "range" {
+		data, err = client.ExecutePrometheusRangeQuery(spec.Query, spec.Dataset, *spec.Start, *spec.End, *spec.Step)
+	} else {
+		data, err = client.ExecutePrometheusInstantQuery(spec.Query, spec.Dataset)
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to execute Prometheus query: %v", err)
+	}
+
```
lucaspin, line 175: How big can these things be? We need some limit around this

pkg/applications/dash0/on_issue_status.go
```diff
@@ -0,0 +1,284 @@
...
+func (t *OnIssueStatus) HandleAction(ctx core.TriggerActionContext) (map[string]any, error) {
+	switch ctx.Name {
+	case "checkIssues":
+		return nil, t.checkIssues(ctx)
+	}
+
+	return nil, fmt.Errorf("action %s not supported", ctx.Name)
+}
+
+func (t *OnIssueStatus) checkIssues(ctx core.TriggerActionContext) error {
```
lucaspin: This function has poor code quality. The excessive nesting makes it confusing to follow what is happening

lucaspin:
We also need to include unit tests for the application and its components

pkg/applications/dash0/on_issue_status.go
```diff
@@ -0,0 +1,395 @@
...
+			Type:     configuration.FieldTypeMultiSelect,
+			Required: false,
+			TypeOptions: &configuration.TypeOptions{
+				MultiSelect: &configuration.MultiSelectTypeOptions{
+					Options: []configuration.FieldOption{}, // Options will be populated dynamically from resources
+				},
+				Resource: &configuration.ResourceTypeOptions{
+					Type: "check-rule",
+				},
+			},
```
lucaspin: If we are loading options dynamically, we shouldn't use FieldTypeMultiSelect. We should use FieldTypeAppInstallationResource, and add a new `Multi` field to ResourceTypeOptions so the UI can render a multi-select instead of a select

pkg/applications/dash0/on_issue_status.go
```diff
@@ -0,0 +1,395 @@
...
+			Description: "Number of minutes between checks (1-59)",
+			TypeOptions: &configuration.TypeOptions{
+				Number: &configuration.NumberTypeOptions{
+					Min: intPtr(1),
+					Max: intPtr(59),
+				},
+			},
+		},
+		{
+			Name:        "listenToAllCheckRules",
```
lucaspin: Do we need this? I think a more common pattern might be to just use `checkRules`: if it's empty, it listens to all the rules; if it's not, it listens to only the ones you specify.

AleksandarCole: I will leave that call to @ropsii once he does UI/UX cleanup

lucaspin: Just leaving an extra note here about the usage of the [Togglable](https://github.com/superplanehq/superplane/blob/main/pkg/components/http/http.go#L167) field

web_src/src/ui/configurationFieldRenderer/MultiSelectFieldRenderer.tsx
lucaspin, line 1: Now that we are using the AppInstallationResourceFieldRenderer, these changes here need to be reverted

web_src/src/ui/configurationFieldRenderer/AppInstallationResourceFieldRenderer.tsx
```diff
@@ -21,13 +29,79 @@ export const AppInstallationResourceFieldRenderer = ({
 }: AppInstallationResourceFieldRendererProps) => {
   const resourceType = field.typeOptions?.resource?.type;
   const useNameAsValue = field.typeOptions?.resource?.useNameAsValue ?? false;
+  // Check for multi - be explicit about truthiness since it's a boolean field
+  const isMulti = Boolean(field.typeOptions?.resource?.multi);
+
+  // Debug logging for checkRules field
```
lucaspin: remove debug logging

PR #1660, opened 2026-01-15
feat: Introduce Cursor integration with a LaunchAgent component

shiroyasha:


pkg/applications/cursor/launch_agent.go
```diff
@@ -0,0 +1,307 @@
...
+	spec := LaunchAgentSpec{}
+	if err := mapstructure.Decode(ctx.Configuration, &spec); err != nil {
+		return fmt.Errorf("failed to decode configuration: %v", err)
+	}
+
+	if err := l.validateSpec(spec); err != nil {
+		return err
+	}
+
+	if ctx.AppInstallation != nil {
```
lucaspin, line 137: no need for nil-checks on contexts - these will always be available

pkg/applications/cursor/cursor.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return configA.Event == configB.Event, nil
+}
+
+func (c *Cursor) ListResources(resourceType string, ctx core.ListResourcesContext) ([]core.ApplicationResource, error) {
+	return []core.ApplicationResource{}, nil
+}
+
+func (c *Cursor) SetupWebhook(ctx core.SetupWebhookContext) (any, error) {
```
lucaspin, line 120: This method is intended for setting up the webhook in the third-party system

Issue #1624, opened 2026-01-13
[PagerDuty] Change 'Service' input to a drop down


markoa:
Currently the input field is not intuitive, the user must figure out that it is expecting a service ID actually and it is not obvious in PD UI where it is (it's in URLs).

Best UX is to provide a drop down of services that are available in given account.

shiroyasha:
for reference, look at how we are implementing resource listing for OpenAI: https://github.com/superplanehq/superplane/blob/main/pkg/applications/openai/text_prompt.go#L56-L69

PR #1662, opened 2026-01-15
fix: PagerDuty service dropdown

e-todorovski-bm:
The UX problem reported in https://github.com/superplanehq/superplane/issues/1624 has already been solved before this PR. The Service dropdown already contained the list of services in PagerDuty.

So in this PR:
- Fixed small visual bug with the dropdown component where the selected value was rendered as if it was a placeholder (grayish instead of black text)
- Added some tests
- Improved wording, so its not confusing that the service input expects a Service ID number

### DIFF
--- a/pkg/applications/pagerduty/create_incident_test.go
+++ b/pkg/applications/pagerduty/create_incident_test.go
@@ -8,54 +8,9 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/superplanehq/superplane/pkg/configuration"
 	"github.com/superplanehq/superplane/pkg/core"
 	"github.com/superplanehq/superplane/test/support/contexts"
 )
-
-func Test__CreateIncident__Configuration(t *testing.T) {
-	component := &CreateIncident{}
-
-	t.Run("returns correct configuration fields", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/applications/pagerduty/create_incident_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/test/support/contexts"
+)
+
+func Test__CreateIncident__Configuration(t *testing.T) {
```
lucaspin: Unit tests for Configuration() method might be a bit too much

pkg/applications/pagerduty/on_incident_test.go
```diff
@@ -213,3 +213,38 @@ func Test__OnIncident__Setup(t *testing.T) {
 		assert.Len(t, appCtx.WebhookRequests, 0)
 	})
 }
+
+func Test__OnIncident__Configuration(t *testing.T) {
```
lucaspin: Same comment about Configuration() tests

PR #1674, opened 2026-01-15
feat: PagerDuty Update Incident

e-todorovski-bm:
Notes:
- (Not reflected in the screenshots): Both "Update Incident" and "Create Incident" now have a "From Email" field that is ALWAYS optional. It has the description "Email address of a valid PagerDuty user. Required for App OAuth and account-level API tokens, optional for user-level API tokens." . 
That description is inline with the PagerDuty docs, but not inline with their API: the API doesnt care about the FROM header in any case. 

UI:

Screenshot 2026-01-15 at 18 39 29
Screenshot 2026-01-15 at 18 38 26
Screenshot 2026-01-15 at 18 38 07
Screenshot 2026-01-15 at 18 39 17

### DIFF
--- a/pkg/applications/pagerduty/list_resources.go
+++ b/pkg/applications/pagerduty/list_resources.go
@@ -8,11 +8,6 @@ import (
 )
 
 func (p *PagerDuty) ListResources(resourceType string, ctx core.ListResourcesContext) ([]core.ApplicationResource, error) {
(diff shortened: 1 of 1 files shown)


shiroyasha:
The implementation looks good. 

I'm not yet sure what how to approach the required email field. Postponing the review of this for next week.

PR #1693, opened 2026-01-16
feat: PagerDuty On Incident Status Update

e-todorovski-bm:
Note:
- Initially I thought including filters for the incident (the same as in "onIncident", like Severity or Status) would be useful, but realised that the "onIncidentStatusUpdate" does not carry the full incident data. I did not want to add a new call that would make the feature more complex, with more steps that could produce potential errors, so I opted out of it. Now we dont have any filters

UI:

Screenshot 2026-01-16 at 15 12 16


shiroyasha:
@e-todorovski-bm it seems that something is not passing for the frontend on CI. Can you fix it please?

e-todorovski-bm:
> @e-todorovski-bm it seems that something is not passing for the frontend on CI. Can you fix it please?

Fixed

PR #1728, opened 2026-01-20
feat: Discord Integration with Send Text Message

e-todorovski-bm:
Decisions:
- added .env in git ignore because i needed to have custom BLOCK_SIGNUP and WEBHOOKS_BASE_URL values. This is not related with this PR. I hope it doesnt mess with something else
- Connection to Discord is more complex (for the user) than the webhook, so added more verbose explanation how to use it.

UI:

Screenshot 2026-01-23 at 13 07 41
Screenshot 2026-01-23 at 13 08 06
Screenshot 2026-01-23 at 13 08 23
Screenshot 2026-01-23 at 13 08 34
Screenshot 2026-01-23 at 13 09 11

### DIFF
--- a/docs/components/Discord.mdx
+++ b/docs/components/Discord.mdx
@@ -78,3 +78,4 @@ Returns metadata about the sent message including message ID, channel ID, and au
   "type": "discord.message.sent"
 }
 ```
+
--- a/docs/components/SMTP.mdx
+++ b/docs/components/SMTP.mdx
@@ -69,3 +69,4 @@ Returns metadata about the sent email including recipients and subject.
   "type": "smtp.email.sent"
 }
 ```
+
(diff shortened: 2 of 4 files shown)


shiroyasha:
aaaa

shiroyasha:
@e-todorovski-bm can we use this? https://discord.com/developers/docs/quick-start/overview-of-apps

e-todorovski-bm:
> @e-todorovski-bm can we use this? https://discord.com/developers/docs/quick-start/overview-of-apps

We can, and most likely we will. The approach that i've started is compatible with the HTTP API. The outgoing webhook (as implemented) allows us to send messages, while the incoming webhooks (part of the Interactions Endpoint, which is part of the HTTP API) allow us to listen to slash commands, on button clicks, etc.

The only thing (that I've found) it doesnt listen to is pure text messages (that do not contain a slash command). I thought this wouldnt be necessary for our case.
To listen to messages we have to setup the Gateway API, which was much trickier to setup (required bigger refactors on our side), which i didnt deem necessary.

If you want i can implement the incoming webhooks within this PR, but they wont be testable without a particular trigger. I propose we make another task (like "[Discord] onSlashCommand") in which we will do both the trigger and expand the Discord App Integration

PR #1843, opened 2026-01-26
feat: Add Datadog integration

### LINKED ISSUE #1542: [Integration] Datadog

thiagokroger:
image

image

Implements Datadog integration with:

- Application registration with API Key, App Key, and Site configuration
- CreateEvent component to create events in Datadog
- OnMonitorAlert trigger to receive webhook notifications from Datadog monitors
- Frontend mappers for both component and trigger

Closes #1542

---

> [!NOTE]
> Adds first-class Datadog integration across backend, docs, and UI.
> 
> - Backend: new `datadog` app with config (`site`, `apiKey`, `appKey`), client, credential validation (`/api/v1/validate`), `Create Event` component, and `On Monitor Alert` trigger with HMAC (`X-Superplane-Signature-256`) verification and tag/transition filters; registered in `server.go`
> - Frontend: mappers/renderers for `datadog.createEvent` and `datadog.onMonitorAlert`, execution detail views, state registry, and icon wiring
> - Docs: new `docs/components/Datadog.mdx` covering instructions, trigger/event payloads, and example outputs; sidebar ordering adjusted
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c581acc74187d9681b1c9e03821f666bfaa397cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


forestileao:
Hello! if you are having trouble to have a  green build: run `make check.build.ui` to fix the UI build and `make gen.components.docs` to regenerate the docs

pkg/integrations/datadog/example_output_create_event.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": 1234567890,
+  "title": "Deployment completed",
+  "text": "Application v1.2.3 has been deployed successfully",
+  "date_happened": 1704067200,
+  "alert_type": "info",
+  "priority": "normal",
+  "tags": ["env:prod", "service:web"],
+  "url": "https://app.datadoghq.com/event/event?id=1234567890"
+}
```
forestileao, line 14: Probably you should wrap this data with: `{"data": {...}, "type": "the emitted type"}`.

check how the payload is set in the sidebar -> run item -> payload. for both trigger and component

pkg/integrations/datadog/example_data_on_monitor_alert.json
```diff
@@ -0,0 +1,18 @@
...
+  "priority": "P1",
+  "tags": ["env:prod", "service:web"],
+  "title": "[Triggered] CPU Usage High",
+  "date": 1704067200,
+  "body": "CPU usage exceeded 90% threshold",
+  "org": {
+    "id": 123,
+    "name": "MyOrg"
+  }
+}
```
forestileao, line 22: same as https://github.com/superplanehq/superplane/pull/1843/files#r2729438933

thiagokroger:
> Hello! if you are having trouble to have a green build: run `make check.build.ui` to fix the UI build and `make gen.components.docs` to regenerate the docs

@forestileao is that something we can tackle with a pre-commit hook? (I mean, for future PRs)

pkg/integrations/datadog/example_output_create_event.json
thiagokroger: uhmm, good point, I should've double checked the types of the other examples

lucaspin:
@thiagokroger the `datadog.createEvent` component looks good, but I don't see how the `datadog.onMonitorAlert` trigger is supposed to work right now.

---

To give you a bit more context into the integration webhook setup in SuperPlane integrations and triggers/components, I'll try to describe the patterns we have used so far for the existing integrations. Basically, we have used two approaches so far for triggers:

1. Create a specific webhook for each component/trigger node. For example, GitHub integration receives the GitHub app webhooks through the integration HTTP endpoint - see [this](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/github.go#L190) - and for each trigger created, we create a specific repository webhook for it. For example, see the [github.onPush trigger](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/on_push.go#L119). Also, this is all done by SuperPlane itself, so the user doesn't have to manually do anything themselves.

2. Use the HTTP endpoint available for the integration, and route the events received in there to the appropriate component nodes using subscriptions. That's how we implemented the Slack integration. See [slack/slack.go](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/slack/slack.go) and [slack/on_app_mention.go](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/slack/on_app_mention.go).

Approach [1] is usually better because we only start receiving events from the integration when people actually create components/triggers for it. Also, in those cases, each webhook is scoped to a single resource on the integration (GitHub repository, Semaphore project, ...) which allow you to only receive the events you actually care about. Approach [2] is a bit worse, since you will receive all the events from the integration from the beginning, generating a lot of unnecessary load, but for some integrations, it might be the only available route.

thiagokroger:
> @thiagokroger the `datadog.createEvent` component looks good, but I don't see how the `datadog.onMonitorAlert` trigger is supposed to work right now.
> 
> To give you a bit more context into the integration webhook setup in SuperPlane integrations and triggers/components, I'll try to describe the patterns we have used so far for the existing integrations. Basically, we have used two approaches so far for triggers:
> 
> 1. Create a specific webhook for each component/trigger node. For example, GitHub integration receives the GitHub app webhooks through the integration HTTP endpoint - see [this](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/github.go#L190) - and for each trigger created, we create a specific repository webhook for it. For example, see the [github.onPush trigger](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/on_push.go#L119). Also, this is all done by SuperPlane itself, so the user doesn't have to manually do anything themselves.
> 2. Use the HTTP endpoint available for the integration, and route the events received in there to the appropriate component nodes using subscriptions. That's how we implemented the Slack integration. See [slack/slack.go](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/slack/slack.go) and [slack/on_app_mention.go](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/slack/on_app_mention.go).
> 
> Approach [1] is usually better because we only start receiving events from the integration when people actually create components/triggers for it. Also, in those cases, each webhook is scoped to a single resource on the integration (GitHub repository, Semaphore project, ...) which allow you to only receive the events you actually care about. Approach [2] is a bit worse, since you will receive all the events from the integration from the beginning, generating a lot of unnecessary load, but for some integrations, it might be the only available route.

I think I understand what you mean, and from what I researched the main problem is that we have to do some manual parts on Datadog itself instead of doing everything from superplane. Even after creating the webhook via API, it won't fire until the user adds @webhook-<name> to their monitor  notifications. This seems different from GitHub where the webhook fires automatically for all events on the repo. 

Anyway, Maybe a better approach here would be to remove the webhook part for now, I tried to add one of each direction for this base integration but we can add that in a later PR. What do you think @lucaspin ?

thiagokroger:
I wlll move to a less noisy PR only with the createEvent feature

Issue #1779, opened 2026-01-23
Daytona Integration


shiroyasha:
- [x] Review the API
- [x] Break down into components and triggers

shiroyasha:
@e-todorovski-bm I'm assigning you as a mentor, but I would like Atanas to take over this ticket.

PR #1889, opened 2026-01-28
feat: Daytona.io Integration

alabro-bm:
- Added Daytona integration for sandbox code execution                          
- Includes 4 components: Create Sandbox, Execute Code, Execute Command, Delete
   Sandbox                                                                              

https://github.com/superplanehq/superplane/issues/1779

Screenshot 2026-01-28 at 12 59 50
Screenshot 2026-01-28 at 13 00 58
Screenshot 2026-01-28 at 13 00 48
Screenshot 2026-01-28 at 13 00 36
Screenshot 2026-01-28 at 13 00 24
Screenshot 2026-01-28 at 13 00 11
Screenshot 2026-01-28 at 12 59 55

### DIFF
--- a/pkg/integrations/daytona/client.go
+++ b/pkg/integrations/daytona/client.go
@@ -1,0 +1,287 @@
+package daytona
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/superplanehq/superplane/pkg/core"
+)
+
+const defaultBaseURL = "https://app.daytona.io/api"
+
+type Client struct {
+	APIKey  string
+	BaseURL string
+	http    core.HTTPContext
+}
+
+func NewClient(httpClient core.HTTPContext, ctx core.IntegrationContext) (*Client, error) {
+	if ctx == nil {
+		return nil, fmt.Errorf("no app installation context")
+	}
+
+	apiKey, err := ctx.GetConfig("apiKey")
+	if err != nil {
+		return nil, err
+	}
+
+	baseURL := defaultBaseURL
+	if customURL, err := ctx.GetConfig("baseURL"); err == nil && string(customURL) != "" {
+		baseURL = string(customURL)
+	}
+
+	return &Client{
+		APIKey:  string(apiKey),
+		BaseURL: baseURL,
+		http:    httpClient,
+	}, nil
+}
+
+// Sandbox represents a Daytona sandbox environment
+type Sandbox struct {
+	ID    string `json:"id"`
+	State string `json:"state"`
+}
+
+// CreateSandboxRequest represents the request to create a sandbox
+type CreateSandboxRequest struct {
+	Snapshot         string            `json:"snapshot,omitempty"`
+	Target           string            `json:"target,omitempty"`
+	Env              map[string]string `json:"env,omitempty"`
+	Labels           map[string]string `json:"labels,omitempty"`
+	AutoStopInterval int               `json:"autoStopInterval,omitempty"`
+}
+
+// ExecuteCodeRequest represents the request to execute code in a sandbox
+type ExecuteCodeRequest struct {
+	Code     string `json:"code"`
+	Language string `json:"language"`
+	Timeout  int    `json:"timeout,omitempty"`
+}
+
+// ExecuteCodeResponse represents the response from code execution
+type ExecuteCodeResponse struct {
+	ExitCode int    `json:"exitCode"`
+	Result   string `json:"result"`
+}
+
+// ExecuteCommandRequest represents the request to execute a command in a sandbox
+type ExecuteCommandRequest struct {
+	Command string `json:"command"`
+	Cwd     string `json:"cwd,omitempty"`
+	Timeout int    `json:"timeout,omitempty"`
+}
+
+// ExecuteCommandResponse represents the response from command execution
+type ExecuteCommandResponse struct {
+	ExitCode int    `json:"exitCode"`
+	Result   string `json:"result"`
+}
+
+// Snapshot represents a Daytona snapshot
+type Snapshot struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// PaginatedSnapshots represents a paginated list of snapshots
+type PaginatedSnapshots struct {
+	Items []Snapshot `json:"items"`
+}
+
+// ListSnapshots lists available snapshots
+func (c *Client) ListSnapshots() ([]Snapshot, error) {
+	responseBody, err := c.execRequest(http.MethodGet, c.BaseURL+"/snapshots", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var result PaginatedSnapshots
+	if err := json.Unmarshal(responseBody, &result); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal snapshots response: %v", err)
+	}
+
+	return result.Items, nil
+}
+
+// Verify checks if the API key is valid by listing sandboxes
+func (c *Client) Verify() error {
+	_, err := c.execRequest(http.MethodGet, c.BaseURL+"/sandbox", nil)
+	return err
+}
+
+// CreateSandbox creates a new sandbox environment
+func (c *Client) CreateSandbox(req *CreateSandboxRequest) (*Sandbox, error) {
+	body, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal request: %v", err)
+	}
+
+	responseBody, err := c.execRequest(http.MethodPost, c.BaseURL+"/sandbox", bytes.NewReader(body))
+	if err != nil {
+		return nil, err
+	}
+
+	var sandbox Sandbox
+	if err := json.Unmarshal(responseBody, &sandbox); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal sandbox response: %v", err)
+	}
+
+	return &sandbox, nil
+}
+
+// APIConfig represents the relevant fields from the /api/config endpoint
+type APIConfig struct {
+	ProxyToolboxURL string `json:"proxyToolboxUrl"`
+}
+
+// FetchConfig fetches the API configuration from the /api/config endpoint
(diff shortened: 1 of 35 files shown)


pkg/integrations/daytona/create_sandbox.go
```diff
@@ -0,0 +1,226 @@
...
+			},
+			Description: "Target region for the sandbox",
+		},
+		{
+			Name:        "autoStopInterval",
+			Label:       "Auto Stop Interval",
+			Type:        configuration.FieldTypeNumber,
+			Required:    false,
+			Description: "Time in minutes before the sandbox auto-stops",
+			Placeholder: "15",
```
lucaspin: Instead of placeholder, define a default value

pkg/integrations/daytona/create_sandbox.go
```diff
@@ -0,0 +1,226 @@
...
+func (c *CreateSandbox) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (c *CreateSandbox) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "snapshot",
+			Label:       "Snapshot",
+			Type:        configuration.FieldTypeString,
```
lucaspin: Seems like we can dynamically load the snapshot options from this [API endpoint](https://www.daytona.io/docs/en/tools/api/#daytona/tag/snapshots/POST/snapshots), so we could probably use an [integration-resource](https://github.com/superplanehq/superplane/blob/main/pkg/configuration/field.go#L32) field to make it easy for the user to select it.

pkg/integrations/daytona/create_sandbox.go
```diff
@@ -0,0 +1,226 @@
...
+			Required:    false,
+			Description: "Base environment snapshot for the sandbox (optional)",
+			Placeholder: "default",
+		},
+		{
+			Name:     "target",
+			Label:    "Target Region",
+			Type:     configuration.FieldTypeSelect,
+			Required: false,
```
lucaspin, line 89: Use a default value here to make it easier for the user

pkg/integrations/daytona/create_sandbox.go
```diff
@@ -0,0 +1,226 @@
...
+		AutoStopInterval: spec.AutoStopInterval,
+		Env:              envMap,
+	}
+
+	sandbox, err := client.CreateSandbox(req)
+	if err != nil {
+		return fmt.Errorf("failed to create sandbox: %v", err)
+	}
+
+	payload := SandboxPayload{
```
lucaspin: Let's not trim the payload here, and just emit the full response from the API request.

pkg/integrations/daytona/create_sandbox.go
```diff
@@ -0,0 +1,226 @@
...
+					},
+				},
+			},
+			Required:    false,
+			Description: "Environment variables to set in the sandbox",
+		},
+	}
+}
+
+func (c *CreateSandbox) Setup(ctx core.SetupContext) error {
```
lucaspin, line 147: We should validate that the snapshot is valid here

pkg/integrations/daytona/daytona.go
```diff
@@ -0,0 +1,121 @@
...
+			Required:    true,
+			Sensitive:   true,
+			Description: "Daytona API key",
+		},
+		{
+			Name:        "baseURL",
+			Label:       "Base URL",
+			Type:        configuration.FieldTypeString,
+			Required:    false,
+			Description: "API base URL (default: https://app.daytona.io/api)",
```
lucaspin: Use https://app.daytona.io/api as default value

pkg/integrations/daytona/daytona_test.go
```diff
@@ -0,0 +1,204 @@
...
+		})
+
+		require.NoError(t, err)
+		assert.NotNil(t, appCtx.Metadata)
+		_, ok := appCtx.Metadata.(Metadata)
+		assert.True(t, ok, "metadata should be of type Metadata")
+	})
+}
+
+func Test__Daytona__Components(t *testing.T) {
```
lucaspin: No need to write unit test for this

pkg/integrations/daytona/daytona_test.go
```diff
@@ -0,0 +1,204 @@
...
+		componentNames[i] = c.Name()
+	}
+
+	assert.Contains(t, componentNames, "daytona.createSandbox")
+	assert.Contains(t, componentNames, "daytona.executeCode")
+	assert.Contains(t, componentNames, "daytona.executeCommand")
+	assert.Contains(t, componentNames, "daytona.deleteSandbox")
+}
+
+func Test__Daytona__Triggers(t *testing.T) {
```
lucaspin: No need to write unit test for this

pkg/integrations/daytona/daytona_test.go
```diff
@@ -0,0 +1,204 @@
...
+	d := &Daytona{}
+
+	triggers := d.Triggers()
+	assert.Empty(t, triggers)
+}
+
+func Test__Daytona__ApplicationInfo(t *testing.T) {
```
lucaspin: No need to write unit test for this

pkg/integrations/daytona/daytona_test.go
```diff
@@ -0,0 +1,204 @@
...
+	d := &Daytona{}
+
+	assert.Equal(t, "daytona", d.Name())
+	assert.Equal(t, "Daytona", d.Label())
+	assert.Equal(t, "daytona", d.Icon())
+	assert.Equal(t, "Execute code in isolated sandbox environments", d.Description())
+	assert.Empty(t, d.Instructions())
+}
+
+func Test__Daytona__Configuration(t *testing.T) {
```
lucaspin: Not need to write unit test for this

pkg/integrations/daytona/delete_sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+func (d *DeleteSandbox) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (d *DeleteSandbox) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "sandboxId",
+			Label:       "Sandbox ID",
+			Type:        configuration.FieldTypeExpression,
```
lucaspin: I don't think this needs to be a configuration.FieldTypeExpression. If you use a FieldTypeString, users are still able to write expressions for it.

pkg/integrations/daytona/delete_sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create client: %v", err)
+	}
+
+	err = client.DeleteSandbox(spec.SandboxID, spec.Force)
+	if err != nil {
+		return fmt.Errorf("failed to delete sandbox: %v", err)
+	}
+
+	payload := DeleteSandboxPayload{
```
lucaspin, line 127: Same comment about not trimming the payload emitted here

pkg/integrations/daytona/delete_sandbox.go
lucaspin: Also, it seems like the [API endpoint](https://www.daytona.io/docs/en/tools/api/#daytona/tag/sandbox/DELETE/sandbox/{sandboxIdOrName}) accepts ID or name, so I'd just use `sandbox` instead of `sandboxId`, and accept both too

pkg/integrations/daytona/delete_sandbox_test.go
```diff
@@ -0,0 +1,200 @@
...
+			Integration:    appCtx,
+			ExecutionState: execCtx,
+		})
+
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "failed to delete sandbox")
+	})
+}
+
+func Test__DeleteSandbox__ComponentInfo(t *testing.T) {
```
lucaspin: No need to write unit tests for static stuff

pkg/integrations/daytona/client.go
```diff
@@ -0,0 +1,224 @@
...
+		Command: command,
+		Timeout: timeoutSeconds,
+	}
+
+	body, err := json.Marshal(cmdReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal request: %v", err)
+	}
+
+	url := fmt.Sprintf("%s/toolbox/%s/toolbox/process/execute", c.BaseURL, sandboxID)
```
lucaspin: It seems like the [toolbox API endpoints](https://www.daytona.io/docs/en/tools/api/#daytona/tag/toolbox) are deprecated?

alabro-bm: That is true, I was not able to find any way to run any commands / code in daytonaio through the provided api (nothing in their documentation nor any replacement or anything). 
I even found that the toolkit is still being used by the unofficial go sdk: https://github.com/PhilippBuschhaus/daytona-sdk-go

Now that being said, they released yesterday official go sdk support: https://github.com/daytonaio/daytona/pull/3366/changes#diff-62172527f27d7367fb058625562a627b782a2b823ef9360c64d49d84444709ac

Screenshot 2026-01-29 at 18 39 39

I looked into their code and from what I can find even they use the toolkit endpoint still (or a new version of it, but nothing new is available in their api) libs/toolbox-api-client-go/README.md.

Maybe we can switch to the go sdk, alto this is a bit risky since its released yesterday :)

lucaspin: I wouldn't use the SDK directly, to avoid making our binaries too bloated.

Regarding their toolbox API: they have an [open ticket](https://github.com/daytonaio/daytona/issues/3314) to document how to interact with their toolbox API, so maybe we could learn something from that one. Apparently, the only difference is that we need to make the requests against some sort of proxy that we can fetch from /api/config endpoint.

alabro-bm: I used that approach, with fetching the toolbox api from the /api/config and it works! Thanks for the feedback.

PR #1904, opened 2026-01-29
chore: Consolidate Orphaned Auth and Organization HTTP Calls

kkomelin:
## Summary

This PR consolidates scattered HTTP calls into centralized service modules, improving maintainability and type safety.

## Changes

**Created Services:**
- `authService.ts` - Handles `/auth/config`, `/login`, `/signup`, `/account`
- `organizationService.ts` - Handles `/organizations` (GET/POST)

**Refactored Components:**
Migrated 7 components to use these services (AccountContext, Login, EmailLogin, OrganizationSelect, OrganizationCreate, InviteLinkAccept, OwnerSetup)

**Result:** Eliminated 12 native `fetch()` calls, centralized API access patterns.

## Swagger Specification

**Problem:** Manual edits to `superplane.swagger.json` get overwritten by `make openapi.spec.gen` (which auto-generates from protobuf files).

**Solution:** Implemented a hybrid approach that automatically merges manual and auto-generated specifications:

**Files Added:**
- `api/swagger/account-auth.swagger.json` - Manual definitions for auth/account/setup endpoints
- `scripts/merge-swagger.js` - Node.js script to merge manual + auto-generated swagger
- Updated `scripts/protoc_openapi_spec.sh` to automatically merge both sources

**How it works:**
1. Protobuf files → auto-generated swagger (48 endpoints)
2. Manual swagger → auth/account/setup endpoints (6 endpoints)
3. Build script merges both → final `superplane.swagger.json` (54 endpoints total)

**Benefits:**
- ✅ No manual swagger maintenance - run `make gen` and it merges automatically
- ✅ No additional dependencies (uses Node.js already installed for `web_src`)
- ✅ Auth endpoints stay as simple HTTP handlers (OAuth flows don't fit gRPC)
- ✅ Both Go and TypeScript clients generated correctly

## Testing

```bash
make gen  # Regenerates swagger, Go client, and TypeScript client
```

All manual endpoints (`/auth/config`, `/login`, `/signup`, `/account`, `/organizations`, `/api/v1/setup-owner`) are now properly documented and won't be overwritten.


kkomelin:
Guys please help with a couple of failed tests because I'm not very familiar with your testing approach.

api/swagger/superplane.swagger.json
```diff
@@ -726,6 +726,40 @@
...
+            "schema": {
+              "$ref": "#/definitions/SetupOwnerRequest"
+            }
+          }
+        ],
+        "tags": [
+          "Organizations"
+        ]
+      }
+    },
```
forestileao: Did you add this manually to the swagger file?
We auto generate the swagger open-api-specs using the command `make openapi.spec.gen`
Check: https://github.com/superplanehq/superplane/blob/main/Makefile#L218

So manual specs would be overwritten.

Everything under /api/v1 are defined in proto files generate this swagger docs. Because we are using grpc-gateway.
But not for auth, since we are using goth library to handle it. check https://github.com/superplanehq/superplane/blob/main/pkg/authentication/authentication.go

some ways to solve it would be:
1 - Create another swagger file for auth related stuff and merge it with the autogenerated
2 - OR Migrate auth routes to grpc gateway

kkomelin: Thank @forestileao. Claude Code added this route for me but I assumed that all /api/v1 routes should be auto-generated. Didn't know your process. We can either remove it from my PR or do it your way which I'm not very confident about because I'm not that good at Golang tooling. What would you prefer?

kkomelin:
> Also, we are using [main/web_src/src/lib/api-interceptor.ts](https://github.com/superplanehq/superplane/blob/main/web_src/src/lib/api-interceptor.ts) to intercept any unauthorized response using the normal fetch

Thanks @forestileao. I noticed that. I suppose it still applies even with my changes.

api/swagger/superplane.swagger.json
kkomelin: Alright, @forestileao. I've researched the topic a little bit more. I'll implement your 1st option for now as the first step towards the ideal approach (option 2) because it's easier for me and because outh stuff may not be very suitable for grpc.

kkomelin: @forestileao Option 1 has been implemented and the pr description has been updated accordingly. Please review

shiroyasha:
It seems that this PR is a bit stale and the diff between the main branch is quite big.
Unfortunately, we didn't manage to properly review it and merge in this release cycle.

I'll close this issue, but lets use it as an inspiration when we want to tackle this problem again.

kkomelin:
Thanks @shiroyasha. I understand

PR #2815, opened 2026-02-03
feat: Jira Base

alabro-bm:


pkg/integrations/jira/list_resources.go
```diff
@@ -0,0 +1,31 @@
+package jira
+
+import (
+	"fmt"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/superplanehq/superplane/pkg/core"
+)
+
+func (j *Jira) ListResources(resourceType string, ctx core.ListResourcesContext) ([]core.IntegrationResource, error) {
```
shiroyasha: @alabro-bm Lets go with `jira.createIssue` and `jira.onIssueCreated` instead of list resources.

alabro-bm: I already have a PR for the createIssue implementation. I'll remove this list_resources.go and add it in that project.

Than we can merge the createissue as followup of this one.

PR #2816, opened 2026-02-03
feat: pagerduty update incident statuses

e-todorovski-bm:
Notes:
- A specific case is the "escalate". we cannot read the current policy easily so we can just move to the higher one. And we are not sure there is a higher one. We can add the order number of the new policy, but if it doesnt exist (e.g. we have only 3 policies, and if we set to "4") it will break with a bad, non-descript error.

UI:
Screenshot 2026-02-03 at 12 31 43
Screenshot 2026-02-03 at 12 31 31
Screenshot 2026-02-03 at 12 30 53
Screenshot 2026-02-03 at 12 31 55

### DIFF
--- a/docs/components/PagerDuty.mdx
+++ b/docs/components/PagerDuty.mdx
@@ -1474,3 +1474,4 @@ Returns the updated incident object with all current information.
   "type": "pagerduty.incident"
 }
 ```
+


shiroyasha:
Closing this in favor of the updated one (rebased with main):
https://github.com/superplanehq/superplane/pull/3212

PR #2866, opened 2026-02-04
chore: Dedicated API operations for keys in secrets + E2E tests for secrets

shiroyasha:
This PR adds E2E + Adds new atomic updates for secrets.

---

**The Problem:** "Round-tripping" masked data 

Previously, we used a single update action for entire secret objects. Since the frontend doesn't have access to the raw values once they're stored, it had to send back masked strings (like ***) to signal "no change." This forced the backend to intelligently merge data, which is a fragile pattern that's prone to accidental overwrites.

I don't want to have fragile patterns around secrets.

---

**My solution** Atomic key updates

The API now supports individual key operations. Instead of updating the whole secret, we target the specific key that needs to change.


pkg/grpc/actions/secrets/describe_secret.go
```diff
@@ -69,13 +69,15 @@ func serializeSecret(ctx context.Context, encryptor crypto.Encryptor, secret mod
...
-		return nil, err
+		// Return empty local data so list/describe still succeed (e.g. secret created with different encryptor)
+		return &pb.Secret_Local{Data: map[string]string{}}, nil
 	}
 
 	var values map[string]string
 	err = json.Unmarshal(data, &values)
 	if err != nil {
-		return nil, err
+		// Return empty local data so list/describe still succeed (e.g. corrupted or wrong format)
```
shiroyasha: wtf

pkg/protos/widgets/widgets_grpc.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.6.0
+// - protoc-gen-go-grpc v1.5.1
```
lucaspin: This version is pinned [here to 1.6.0](https://github.com/superplanehq/superplane/blob/main/Dockerfile#L59). Do not regenerate things with a different version

shiroyasha: The trick was to rebuild docker compose with no cache. `make dev.setup` didn't cut it.

PR #2875, opened 2026-02-04
feat: Get Incident

e-todorovski-bm:
Notes:
- There are 4 different API calls in order to get the full picture of the incident. Besides the main one (get incident), we fetch the alerts, notes and log entries. If you think this is an overkill, it can be reduced, but I thought if someone wants to get info on the incident, they should have all available data like in the UI
- Only if the main (get incident) API call fails we receive an error on the FE. Otherwise, if any of the other 3 API calls fail, we skip the error, we just omit that field

UI:
Screenshot 2026-02-04 at 15 04 57
Screenshot 2026-02-04 at 15 05 13
Screenshot 2026-02-04 at 15 05 25
Screenshot 2026-02-04 at 15 05 35

### DIFF
--- a/pkg/integrations/pagerduty/get_incident_test.go
+++ b/pkg/integrations/pagerduty/get_incident_test.go
@@ -112,7 +112,7 @@ func Test__GetIncident__Execute(t *testing.T) {
 			},
 		}
 
-		executionState := &contexts.ExecutionState{}
+		executionState := &contexts.ExecutionStateContext{}
 
 		err := component.Execute(core.ExecutionContext{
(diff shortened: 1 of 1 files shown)


shiroyasha:
Reviewing it fresh eyes. 

There is a chance that this component is enough and we don't need the listNotes and listLogEntries.

Reviewing it further in ~2hours.

shiroyasha:
Closing this one in favor of explicit "listing notes" and "listing log entries". 
The get incident component needs to be costructed based on several different APIs.
Individual components are more optimal.

PR #2902, opened 2026-02-05
chore(sendgrid): update event filtering to use predicate list

forestileao:
- Refactor sendgrid.onEmailEvent category filter to use any-predicate-list predicates and MatchesAnyPredicate.
- Preserve legacy categoryFilter string behavior by converting it to predicates.
- Update the SendGrid trigger UI metadata to render predicate filters with formatPredicate.
- Document trigger filter standards in Cursor review rules and the integration checklist.

### DIFF
--- a/docs/components/SendGrid.mdx
+++ b/docs/components/SendGrid.mdx
@@ -49,7 +49,7 @@ The On Email Event trigger emits events when SendGrid posts delivery or engageme
 ### Configuration
 
 - **Event Types**: Optional filter for specific SendGrid events (processed, delivered, bounce, open, click, etc.)
-- **Category Filter**: Optional category filter (supports `*` wildcards)
+- **Category Filter**: Optional list of predicates (`equals`, `notEquals`, `matches` regex)
 
 ### Webhook Verification
 


pkg/integrations/sendgrid/on_email_event.go
```diff
@@ -39,8 +39,13 @@ var sendGridEventTypes = []configuration.FieldOption{
 type OnEmailEvent struct{}
 
 type OnEmailEventConfiguration struct {
+	EventTypes     []string                  `json:"eventTypes" mapstructure:"eventTypes"`
+	CategoryFilter []configuration.Predicate `json:"categoryFilter" mapstructure:"categoryFilter"`
+}
+
+type onEmailEventConfigurationRaw struct {
```
lucaspin: Why do we need OnEmailEventConfiguration and OnEmailEventConfigurationRaw?

pkg/integrations/sendgrid/on_email_event.go
```diff
@@ -196,8 +206,8 @@ func (t *OnEmailEvent) Cleanup(ctx core.TriggerContext) error {
 }
 
 type WebhookConfiguration struct {
-	EventTypes     []string `json:"eventTypes"`
-	CategoryFilter string   `json:"categoryFilter"`
+	EventTypes     []string                  `json:"eventTypes"`
+	CategoryFilter []configuration.Predicate `json:"categoryFilter"`
```
lucaspin: Why do these need to be on the WebhookConfiguration?

web_src/src/pages/workflowv2/mappers/sendgrid/on_email_event.ts
```diff
@@ -139,11 +140,19 @@ function shouldDisplayEventTypes(eventTypes?: string[]): boolean {
...
+  if (!categoryFilter || categoryFilter.length === 0) {
     return false;
   }
 
-  const trimmed = categoryFilter.trim();
-  return trimmed !== "" && trimmed !== "*";
+  return true;
+}
+
+function formatCategoryFilter(categoryFilter: Predicate[]): string {
```
lucaspin, line 151: We also have UI helpers to render these predicates: https://github.com/superplanehq/superplane/blob/main/web_src/src/pages/workflowv2/mappers/utils.ts#L86

pkg/integrations/sendgrid/on_email_event.go
forestileao: we are not using it at all, let me remove it

web_src/src/pages/workflowv2/mappers/sendgrid/on_email_event.ts
forestileao: In this same funciton we are using the helper, look one line below and check the import

pkg/integrations/sendgrid/on_email_event.go
forestileao: my bad, it should not support both

.cursor/commands/component-review.rules.md
```diff
@@ -73,6 +73,10 @@ available to them (e.g., requiring a Discord channel ID).
 When users can choose from existing resources (e.g., a GitHub repository), 
 prefer a dropdown selector over manual entry.
 
+### Trigger filters
```
lucaspin: I don't think this is specific to triggers. We could also have components that do some sort of equality / non-equality / regex on values returned, and we could also this pattern there. I would also say:

> If any-predicate-list cannot meet your requirements, look for ways to extend it before developing a specific implementation. Before developing a specific implementation, look for ways to develop it in ways that the same pattern could be reused in other components / triggers.

forestileao: generalized it

pkg/integrations/sendgrid/on_email_event.go
forestileao: not even event types need to be there

PR #2905, opened 2026-02-06
feat: render integration base + 2 components

forestileao:
## Summary
Add a new **Render** integration to SuperPlane with:
- API key connection (optional workspace ID)
- `On Event` trigger from Render webhooks
- `Trigger Deploy` action for Render services

## What’s included
- Backend integration (`pkg/integrations/render/*`)
- Render API client (services, owners, webhooks, deploy trigger)
- Automatic webhook provisioning/cleanup through Render API
- Webhook signature verification (`webhook-id`, `webhook-timestamp`, `webhook-signature`)
- Frontend mappers + icon + registry wiring for Render components/triggers
- New docs page: `docs/components/Render.mdx`
- Sidebar order updates for docs components

Video showing the implementation

https://github.com/user-attachments/assets/41c75657-9f1f-422c-8741-3db31480be32

### DIFF
--- a/docs/components/Render.mdx
+++ b/docs/components/Render.mdx
@@ -161,8 +161,8 @@ The Deploy component starts a new deploy for a Render service and waits for it t
 {
   "data": {
     "createdAt": "2026-02-05T16:10:00.000000Z",
+    "deployId": "dep-cukouhrtq21c73e9scng",
     "finishedAt": "2026-02-05T16:15:00.000000Z",
-    "id": "dep-cukouhrtq21c73e9scng",
     "serviceId": "srv-cukouhrtq21c73e9scng",
     "status": "succeeded"
   },


shiroyasha:
@forestileao  

1/ Does it work if you set up two different nodes to listen to different events, or will the web hooks clash? 
2/ Will it create one, or multiple web hooks?
3/ Having a single On Event component that covers 10+ events is different to how we were doing things for other integrations. cc @lucaspin want to hear your opinion here as well

forestileao:
@shiroyasha 
1 - It works with no problem using the same webhook
image

2 - It reuses the same webhook
image

3 - I think we should not have one component for each, since the payload is very simple and we have 55+ events to cover

lucaspin:
> Having a single On Event component that covers 10+ events is different to how we were doing things for other integrations

I think we shouldn't have a single `onEvent` trigger here.

The way I like to think about triggers is: a trigger is always scoped to a (1) specific resource type, (2) specific resource, (3) some additional things. Examples:
- `semaphore.onPipelineDone` -> we select the project (resource type + resource)
- `github.onPush` -> we select the repository (resource type + resource), and we have filters for the refs (additional things).
- `pagerduty.onIncident` -> we select the service (resource type + resource)

The reason for that is: what dictates what you need to do after receiving an event is usually the resource type. If you received all events through the same trigger, you'd need a filter after it to route build events to one path, and deploy events to another path.

For example, for Render, we have Builds and we have Deploys. Those are two different resource types. If I receive an event for a Build, I do something completely different than I would do if I received an event for a Deploy. And we can see that in the screenshot you shared: you have two triggers, one for deploy events, one for build events. So, what I'd propose here is to have triggers per resource: `render.onDeployEvent`, `render.onBuildEvent`, ...

I'm ok with not having a different trigger per event, e.g. `render.onDeployStarted`,`render.onDeployFinished`, but I do think we need to have triggers per resource type at least. The states, we can configure through a MultiSelect, similar to how we select the states in `github.onPullRequest` trigger.

lucaspin:
Also additional note: we should always aim to share webhooks, if they use the same underlying event configuration. That's what CompareWebhookConfig() is for. For example, for the GH triggers, if we have two github.onPush triggers, one for main branch, and one for release branches, both of those triggers use the same webhook in GH.

forestileao:
@lucaspin Trigger per Resource Type!! That make total sense!

lucaspin:
Also, additional note: if filters are part of the trigger, we should always have a the default filter for the most common use case. Two reasons for this: (1) easier for the user to configure, (2) we don't produce unnecessary events to the system. For example, the default `github.onPush` filter is for commits on the main branch.

forestileao:
@lucaspin I think for this specific one will be required to use one webhook for everything
image
The problem is that can use only one per canvas

lucaspin:
@forestileao is there a way to know which plan the workspace configured in the integration is in? We might use different strategy depending on it

> The problem is that can use only one per canvas

Why one per canvas? It would be one for the integration itself, no?

forestileao:
Yes, sorry, one per integration.
But no, they don't have an API for billing and plans as far as I researched 
@lucaspin

lucaspin:
Moved the discussion [here](https://discord.com/channels/1409914582239023200/1469364546677243964) to involve more people

forestileao:
Updated the PR with:

1. **Mergeable webhook configuration**
- Added a mergeable webhook configuration flow so existing webhooks can be updated with newly requested event types.
- This prevents creating extra webhooks and avoids subscribing to all events when not needed..

2. **Plan-based webhook strategy**
- **Professional**: uses a **single shared webhook** for the integration.
- **Organization / Enterprise**: uses **resource-type webhooks** (for example, deploy and build).
- This keeps webhook behavior aligned with Render plan capabilities while reducing unnecessary event traffic.

forestileao:
Video of the webhook flow:

https://github.com/user-attachments/assets/9a495a78-b328-4aa8-849f-282bff020712

pkg/integrations/render/common.go
```diff
@@ -0,0 +1,537 @@
...
+		slices.Equal(normalizedA.EventTypes, normalizedB.EventTypes)
+}
+
+func onResourceEventConfigurationFields(
+	eventTypeOptions []configuration.FieldOption,
+	defaultEventTypes []string,
+) []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:     "serviceId",
```
lucaspin: Just use `service` instead of `serviceId`. Keeps it open to allow the service to be configured using names too

pkg/integrations/render/common.go
```diff
@@ -0,0 +1,537 @@
+package render
```
lucaspin, line 1: A lot of functions in this file use the `render` prefix, like `renderWebhookConfigurationForResource`, `renderWebhookName`, `renderWebhookEventFilter`, ... There's no need for that. The package is already called `render`

pkg/integrations/render/common.go
```diff
@@ -0,0 +1,537 @@
...
+) WebhookConfiguration {
+	metadata := Metadata{}
+	if err := mapstructure.Decode(integration.GetMetadata(), &metadata); err != nil {
+		return WebhookConfiguration{
+			Strategy:   renderWebhookStrategyIntegration,
+			EventTypes: normalizeWebhookEventTypes(eventTypes),
+		}
+	}
+
+	workspacePlan := strings.ToLower(strings.TrimSpace(metadata.WorkspacePlan))
```
lucaspin: Do we really need to use `strings.TrimSpace` and `strings.ToLower` for the metadata fields? Just do it once on Setup() for the configuration, and save the proper value in the metadata. Once in the metadata, we know it's already trimmed and lowered

pkg/integrations/render/common.go
```diff
@@ -0,0 +1,537 @@
...
+		// webhook-id.webhook-timestamp.body.webhook-secret
+		h = hmac.New(sha256.New, key)
+		h.Write([]byte(payloadPrefix))
+		h.Write(ctx.Body)
+		h.Write([]byte("."))
+		h.Write(secretText)
+		if matchesAnyRenderSignature(signatures, h.Sum(nil)) {
+			return nil
+		}
+	}
```
shiroyasha, line 618: I asked AI to review:

Please verify this @forestileao 

> verifyRenderWebhookSignature validates HMAC but never checks that webhook-timestamp is recent. This allows replay of any previously valid webhook (e.g., a captured request can be re‑sent indefinitely). Other integrations (e.g., Slack) do enforce a time window.

pkg/integrations/render/example_data_on_deploy.json
```diff
@@ -0,0 +1,11 @@
...
+  "data": {
+    "createdAt": "2026-02-05T16:00:00.000000Z",
+    "id": "evj-cukouhrtq21c73e9scng",
+    "serviceId": "srv-cukouhrtq21c73e9scng",
+    "serviceName": "backend-api",
+    "status": "succeeded"
+  },
+  "timestamp": "2026-02-05T16:00:01.000000Z",
+  "type": "render.deploy.ended"
+}
```
shiroyasha, line 12: The file is called on_event, but we transitioned to on_build, on_deploy.

pkg/integrations/render/on_build.go
```diff
@@ -0,0 +1,113 @@
...
+	return onResourceEventConfigurationFields(buildEventTypeOptions, buildDefaultEventTypes)
+}
+
+func (t *OnBuild) Setup(ctx core.TriggerContext) error {
+	config, err := decodeOnResourceEventConfiguration(ctx.Configuration)
+	if err != nil {
+		return fmt.Errorf("failed to decode configuration: %w", err)
+	}
+
+	if config.ServiceID == "" {
```
lucaspin: We need to validate the service here, and save it on metadata

pkg/integrations/render/on_deploy_test.go
```diff
@@ -0,0 +1,351 @@
+package render
+
+import (
```
shiroyasha, line 3: Similar here. The file is called on_event, but we transitioned to on_build, on_deploy.

pkg/integrations/render/on_deploy.go
```diff
@@ -0,0 +1,122 @@
...
+	return onResourceEventConfigurationFields(deployEventTypeOptions, deployDefaultEventTypes)
+}
+
+func (t *OnDeploy) Setup(ctx core.TriggerContext) error {
+	config, err := decodeOnResourceEventConfiguration(ctx.Configuration)
+	if err != nil {
+		return fmt.Errorf("failed to decode configuration: %w", err)
+	}
+
+	if config.ServiceID == "" {
```
lucaspin: Same comment about validating service and storing it on metadata

web_src/src/pages/workflowv2/mappers/render/common.ts
```diff
@@ -0,0 +1,152 @@
...
+    return "Render Event";
+  }
+
+  return eventLabelsByType[event] || eventLabelsByConfig[event] || event;
+}
+
+export const renderTriggerRenderer: TriggerRenderer = {
+  getTitleAndSubtitle: (context: TriggerEventContext): { title: string; subtitle: string } => {
+    const event = context.event as RenderEventData | undefined;
+    const title = buildTitle(event, context.event?.type as string);
```
shiroyasha, line 56: Cursor pointed out this:

> File: web_src/src/pages/workflowv2/mappers/render/common.ts L52–L55
getTitleAndSubtitle casts context.event to RenderEventData, but the service name/id is in context.event.data. Result: title likely shows “Service · …” instead of a real service name.
Impact: degraded UX in sidebar and execution details.
Suggestion: use const event = context.event?.data as RenderEventData | undefined; (pattern used by other renderers).

tbh, I don't understand it, but lets verify

pkg/integrations/render/on_deploy_test.go
```diff
@@ -0,0 +1,351 @@
+package render
```
lucaspin, line 1: We don't have an `onEvent` trigger, so we shouldn't have `on_event_test.go`. Let's separate unit tests for each trigger into its own file

pkg/integrations/render/render.go
```diff
@@ -0,0 +1,515 @@
...
+			Name:        "apiKey",
+			Label:       "API Key",
+			Type:        configuration.FieldTypeString,
+			Required:    true,
+			Sensitive:   true,
+			Description: "Render API key",
+		},
+		{
+			Name:        "ownerId",
+			Label:       "Workspace ID",
```
lucaspin: The label is "Workspace ID", but the configuration is `ownerId`, which is inconsistent, let's just use workspace for both. Also, same comment about having the ID on the field. If possible, just call it `workspace`, and allow both ID and name

pkg/integrations/render/render.go
```diff
@@ -0,0 +1,515 @@
...
+
+type Render struct{}
+
+type Configuration struct {
+	APIKey        string `json:"apiKey" mapstructure:"apiKey"`
+	OwnerID       string `json:"ownerId" mapstructure:"ownerId"`
+	WorkspacePlan string `json:"workspacePlan" mapstructure:"workspacePlan"`
+}
+
+type Metadata struct {
```
lucaspin, line 26: Having metadata about workspace on top level makes it trickier to extend this in the future if we realize we need to store metadata about some other resource there. Having something like this makes it a bit more extensible:

```golang
type Metadata struct {
  Workspace *WorkspaceMetadata
}

type WorkspaceMetadata struct {
  ID string
  Plan string
}
```

pkg/integrations/render/common.go
forestileao: The type of this field is FieldTypeIntegrationResource, not sure if I agree using service. Since the user will already select it from the list

pkg/integrations/render/render.go
```diff
@@ -0,0 +1,515 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ownerID, err := r.ownerID(client, ctx.Integration)
+	if err != nil {
+		return nil, err
+	}
+
+	webhookURL := strings.TrimSpace(ctx.Webhook.GetURL())
```
lucaspin: no need to TrimSpace things coming from the ctx.Webhook. We control that fully

pkg/integrations/render/render.go
```diff
@@ -0,0 +1,515 @@
...
+			Type: resourceType,
+			Name: name,
+			ID:   service.ID,
+		})
+	}
+
+	return resources, nil
+}
+
+func (r *Render) SetupWebhook(ctx core.SetupWebhookContext) (any, error) {
```
lucaspin: This function is a bit hard to follow, simplify it

pkg/integrations/render/render.go
```diff
@@ -0,0 +1,515 @@
...
+	return []core.Action{}
+}
+
+func (r *Render) HandleAction(ctx core.IntegrationActionContext) error {
+	return nil
+}
+
+func (r *Render) ownerID(client *Client, integration core.IntegrationContext) (string, error) {
+	metadata := Metadata{}
+	if err := mapstructure.Decode(integration.GetMetadata(), &metadata); err == nil {
```
lucaspin: We have 4 nested if's here, mostly because we are using strings.TrimSpace() in metadata fields. No need for that

pkg/integrations/render/common.go
lucaspin: The user will select from a list in the UI only. When using YAML, there's no selecting from a list. Also, if we want to allow the service configuration to be a name also, not only an ID, what would you do? This is also a pattern we use in other places, so consistency is important.

pkg/integrations/render/on_deploy_test.go
forestileao: file renamed

pkg/integrations/render/common.go
forestileao: This is now resolved. VerifyWebhookSignature parses webhook-timestamp and rejects old/far timestamps.

pkg/integrations/render/common.go
forestileao: Set to trimSpace on buildMetadata and on Api Key only

pkg/integrations/render/on_build.go
forestileao: now it is validated inside `ensureServiceInMetadata`

pkg/integrations/render/render.go
forestileao: Sorry, the api reference says `List workspaces`. But the api route says `/owners`.
Renaming everything to workspaces

forestileao: added a comment to explain this inconsistency

web_src/src/pages/workflowv2/mappers/render/common.ts
forestileao: Now this is already handled. In render mapper we read event payload from context.event?.data (not context.event), so service fields come from the correct object. Title uses serviceName/serviceId fallback accordingly.

forestileao:
@lucaspin @shiroyasha PR comments addressed.
Confirmed that everything is still working.
Waiting for https://github.com/superplanehq/superplane/pull/2937 to be merged then I can implement the new Merge() method and replace my implementation

image

forestileao:
@lucaspin Merged with main. New interface method `Merge is implemented`

pkg/integrations/render/common.go
```diff
@@ -0,0 +1,659 @@
...
+}
+
+func ensureServiceInMetadata(ctx core.TriggerContext, config OnResourceEventConfiguration) error {
+	serviceValue := strings.TrimSpace(config.Service)
+	if serviceValue == "" {
+		return fmt.Errorf("service is required")
+	}
+
+	nodeMetadata := OnResourceEventMetadata{}
+	if ctx.Metadata != nil {
```
lucaspin: No need to check if metadata context exists, it will always exist

pkg/integrations/render/trigger_deploy.go
```diff
@@ -0,0 +1,153 @@
...
+
+type TriggerDeploy struct{}
+
+type TriggerDeployConfiguration struct {
+	Service    string `json:"service" mapstructure:"service"`
+	ClearCache bool   `json:"clearCache" mapstructure:"clearCache"`
+}
+
+func (c *TriggerDeploy) Name() string {
+	return "render.triggerDeploy"
```
lucaspin: It seems to me it would be more useful way for a component like this to behave like the `semaphore.runWorkflow` one, where we wait to finish the execution until the deploy is finished

forestileao: True! Good that we already know deploy_ended event exists

lucaspin: Maybe we can even change the name of the component to just `render.deploy` instead, and use two output channels

pkg/integrations/render/deploy.go
```diff
@@ -195,13 +195,15 @@ func (c *TriggerDeploy) Execute(ctx core.ExecutionContext) error {
 		return err
 	}
 
-	ctx.Logger.Infof("Triggered deploy %s for service %s", deployID, spec.Service)
+	if ctx.Logger != nil {
+		ctx.Logger.Infof("Triggered deploy %s for service %s", deployID, spec.Service)
+	}
```
shiroyasha: This one looks suspicious. It is checking if logger exists, could be redundant check.

forestileao: removed

forestileao:
@lucaspin @shiroyasha  Added new Outputs channel (success and fail). Also, renamed `trigger_deploy` to just `deploy` and udpated its behavior to only show its final state once deploy is finished.
image


Note: I needed to inject HTTP and integration context into the [webhook context](https://github.com/superplanehq/superplane/pull/2905/changes#diff-a5008ad28108159760e66faef71fc7683c161b2aafd94f03782528aef51f04beR808-R812) because the render event  webhook only sends the eventId and not the deploy id in its payload. So I needed to fetch the event using a GET request to `/events/{event}` to finally get the deploy Status

forestileao:
@lucaspin update: added deployId and BuildId to the payloads.

Also added a comment to avoid using http context on webhook nide context

PR #2931, opened 2026-02-07
fix: prevent option selection reset in AutoCompleteSelect

kkomelin:
## Summary                                                                                                                                                  
                                                                                                                                                              
  - Fix a race condition in `AutoCompleteSelect` where selecting an option would sometimes reset to the previous value, particularly noticeable when using a laptop touchpad
  - Move `preventDefault` from `click` to `onMouseDown` on option items to actually prevent the input blur from firing during selection
  - Remove duplicate `onClick` handler on the inner div that caused `handleOptionSelect` to fire twice per click via event bubbling

## Demo

[Watch demo video](https://www.loom.com/share/2f5ea84b99fb4a9dbd203cea6b31f0b6)

  ## Root cause

When clicking an option, the browser fires `mousedown` -> input `blur` -> `click` in sequence. The blur handler checked `e.relatedTarget` to detect if focus  moved to the dropdown list, but since option `<div>` elements aren't focusable, `relatedTarget` was always `null`. This caused the blur handler to run and close the dropdown, racing with the click event. Touchpad clicks have slightly different timing than mouse clicks, making the race condition more likely to trigger.

  ## Test plan

  - [ ] Select options in any AutoCompleteSelect across the app using a touchpad (e.g. model picker in Claude -> Create Message) - should work reliably
  - [ ] Select options using a mouse - should work reliably
  - [ ] Type to filter, then select a filtered option - should work
  - [ ] Press Escape to close the dropdown - should still work
  - [ ] Click outside the dropdown - should still close it
  - [ ] `make check.build.ui` passes

### DIFF
--- a/.cursor/commands/changelog.md
+++ b/.cursor/commands/changelog.md
@@ -6,17 +6,17 @@ description: Generate a "what's new" changelog from merged commits over a time r
 
 Generate a changelog of what was merged to `main` for a given time range. The output is a single markdown file in `tmp/` with new integrations, new components and triggers, improvements, security updates, and bug fixes.
 
-**Use the skill `superplane-changelog`** for the full workflow: time range, classifying commits (new integrations vs new components vs improvements vs security vs bug fixes), format rules (no em dashes, no "We", **Integration:** components, user-focused improvements, dedicated Security section with CVE when available), and output structure.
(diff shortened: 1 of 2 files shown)


AleksandarCole:
@kkomelin thanks for the PR!
Looks good to me. Waiting for CI to pass and I will merge it.

kkomelin:
Great @AleksandarCole. Thanks for reviewing it!

AleksandarCole:
@kkomelin merged! Thank you for contributing.

kkomelin:
Thanks @AleksandarCole

PR #2936, opened 2026-02-07
feat: add Mapper component for reshaping data between nodes

kkomelin:
## Summary                                                                           
                                                                                                                                                              
   - Add new **Mapper** core component that reshapes data between workflow nodes by defining field name/value pairs                                           
   - Supports `{{expression}}` syntax in values (resolved by existing `NodeConfigurationBuilder`)                                                             
   - Supports dot notation for nested field creation (e.g., `user.profile.email` → `{"user": {"profile": {"email": "..."}}}`)                                 
   - `keepOnlyMapped` toggle: output only mapped fields (default) or merge on top of incoming data

Borrowed the idea from the popular [Edit Fields (Set)](https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.set/) n8n node.

## Demo

[Watch demo video](https://www.loom.com/share/9eede28af6b54fd5a2a3500e5f3534a1)       

   ## Configuration

   | Field | Type | Description |
   |-------|------|-------------|
   | `fields` | list of `{name, value}` | Field mappings to apply |
   | `keepOnlyMapped` | boolean (default: `true`) | When false, merges mapped fields with incoming data |

   ## Validation

   - [x] `make format.go` — clean
   - [x] `make lint` — no warnings
   - [x] `make test PKG_TEST_PACKAGES=./pkg/components/mapper/...` — 17 tests pass
   - [x] `make check.build.app` — compiles

   ## Test plan

   - [ ] Verify single and multiple field mappings produce correct output
   - [ ] Verify `keepOnlyMapped: true` excludes incoming data from output
   - [ ] Verify `keepOnlyMapped: false` merges mapped fields on top of incoming data
   - [ ] Verify dot notation creates nested structures
   - [ ] Verify Setup rejects empty fields, duplicate names, and empty names
   - [ ] Verify component registers and appears in the component registry

## Question

Would you like it to be **Mapper** or **Map**, considering we have verbs like Merge and Schedule, and nouns like Approval and HTTP Request?

### DIFF
--- a/pkg/components/mapper/example.go
+++ b/pkg/components/mapper/example.go
@@ -1,0 +1,18 @@
+package mapper
+
+import (
+	_ "embed"
+	"sync"
+
+	"github.com/superplanehq/superplane/pkg/utils"
+)
+
+//go:embed example_output.json
+var exampleOutputBytes []byte
+
+var exampleOutputOnce sync.Once
+var exampleOutput map[string]any
+
+func (m *Mapper) ExampleOutput() map[string]any {
+	return utils.UnmarshalEmbeddedJSON(&exampleOutputOnce, exampleOutputBytes, &exampleOutput)
+}
--- a/pkg/components/mapper/example_output.json
+++ b/pkg/components/mapper/example_output.json
@@ -1,0 +1,8 @@
+{
+  "data": {
+    "name": "John",
+    "email": "john@example.com"
+  },
+  "timestamp": "2026-02-07T12:00:00.000000000Z",
+  "type": "mapper.executed"
+}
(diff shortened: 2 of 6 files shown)


shiroyasha:
@kkomelin thanks for the PR! 

I'm starting first with assigning @AleksandarCole to review if we want to have this in the application. I see the appeal, but also see the downsides as it makes the component configuration harder.

kkomelin:
@shiroyasha Sure. It's up to you guys.

I personally missed this component when switched from n8n. Used it a lot there to make the flows cleaner (without noise) and the amount of traffic between services smaller.

shiroyasha:
Thank you for the contribution @kkomelin. 
Great idea, but we will not be supporting this at the moment.
I'll close it for now, but lets use it as an inspiration when we decide to focus on simplifying this flow.

kkomelin:
Sure @shiroyasha. Let's wait until users ask for it.

PR #2937, opened 2026-02-07
chore: add WebhookHandler interface

lucaspin:
The `Integration` interface was starting to get full of webhook-related methods, but not all integrations make use of those. In fact, from our current integrations, 11/16 do not use it, so more than half do not use it. Due to that, we move that logic into a more specialized `WebhookHandler` interface. That  keeps the Integration interface clean, and makes it a lot more explicit when an integration makes use of it.


pkg/registry/registry.go
```diff
@@ -87,7 +103,11 @@ func (r *Registry) Init() {
 	}
 
 	for name, integration := range registeredIntegrations {
-		r.Integrations[name] = NewPanicableIntegration(integration)
+		r.Integrations[name] = integration
```
forestileao: why Integrations are no longer panic-wrapped when loaded into the runtime registry?
Without local recover, so an integration panic can now crash worker execution, no?

pkg/registry/registry.go
```diff
@@ -87,7 +103,11 @@ func (r *Registry) Init() {
...
+	}
+
+	for name, webhookHandler := range registeredWebhookHandlers {
+		r.WebhookHandlers[name] = webhookHandler
```
forestileao: Same here, for panic-wrapping

PR #2971, opened 2026-02-08
feat: restrict event payload size emitted by components

lucaspin:
Next one after https://github.com/superplanehq/superplane/pull/2918. Here, we restrict the size of the payloads emitted by components and triggers through Emit(). This protects the storage layer of the engine against improper implementations. We start with a 32k limit.


shiroyasha:
What happens on the canvas (what does the user see) when this event size is breached?

lucaspin:
For components, they see an execution in error state with the "event payload too large". For triggers, they see nothing, since the event was rejected. The main focus of this wasn't the users though. The main focus of this are the people implementing components. If you are trying to develop a component and you hit this issue, you're probably doing something wrong.

shiroyasha:
I was checking the user side because I can easily imagine that it works on toy examples while implementing, and puff, fails in random moments for users.

Good that the error shows it to them. 👍 

For triggers, it is a bit more concerning that there is no visibility. It can suddenly just stop working, and you have no idea why.

PR #2990, opened 2026-02-09
feat: Add Buildkite integration

immz4:
Implements #2360

This PR implements integration for the Buildkite, which allows users to:

- Create new builds
- Trigger logic when build gets finished

Integration requires following:

- Buildkite API key with `read_organizations`, `read_user`, `read_pipelines`, `read_builds` and `write_builds` permissions
- Manual webhook setup (unfortunately Buildkite API does not provide methods for managing webhooks, like Semaphore) for `build.finished` event

API keys are created in Personal Settings > API Access Tokens

Webhook is created in Organization Settings > Notification Services with following parameters:

- Webhook URL - `https://<backend_address>/api/v1/integrations/<integration_id>/webhook` (URL is provided during the setup)
- Token - use generated string
- Events - `build.finished`
- Pipelines - All Pipelines

Limitations:

- Currently "Trigger Build" actions relies on polling build result every 30s

Video: https://youtu.be/YOXcCU9rAS0


web_src/src/pages/workflowv2/mappers/buildkite/on_build_finished.ts
```diff
@@ -0,0 +1,130 @@
...
+
+interface OnBuildFinishedMetadata {
+  url?: string;
+  token?: string;
+  project?: {
+    id: string;
+    name: string;
+    url: string;
+  };
+}
```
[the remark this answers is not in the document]
immz4: Fixed

pkg/integrations/buildkite/trigger_build.go
```diff
@@ -0,0 +1,523 @@
...
+
+	client, err := NewClient(ctx.HTTP, ctx.Integration)
+	if err != nil {
+		return err
+	}
+
+	targetBuild, err := client.GetBuild(metadata.Organization, metadata.Pipeline, metadata.BuildNumber)
+	if err != nil {
+		return ctx.Requests.ScheduleActionCall("poll", map[string]any{}, PollInterval)
+	}
```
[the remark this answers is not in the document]
immz4: Fixed

docs/components/Buildkite.mdx
```diff
@@ -0,0 +1,207 @@
...
+- **Branch**: Git branch to build
+- **Commit**: Git commit SHA (optional, defaults to HEAD)
+- **Message**: Optional build message
+- **Environment Variables**: Optional env vars for the build
+- **Metadata**: Optional metadata for the build
+
+## Output Channels
+
+- **Passed**: Build completed successfully
+- **Failed**: Build failed, was cancelled, or was blocked" />
```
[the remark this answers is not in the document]
immz4: Fixed

AleksandarCole:
@immz4 thanks for the submission and a clear overview!

As you already mentioned - creating webhooks manually is not ideal. Even though they don't support API endpoint for the organization notification webhooks - does it make sense to use [pipeline webhooks](https://buildkite.com/docs/apis/rest-api/pipelines#add-a-webhook)? If I understand correctly `pipeline` on Buildkite is equivalent of `project` on Semaphore, but I might be wrong.

immz4:
> @immz4 thanks for the submission and a clear overview!
> 
> As you already mentioned - creating webhooks manually is not ideal. Even though they don't support API endpoint for the organization notification webhooks - does it make sense to use [pipeline webhooks](https://buildkite.com/docs/apis/rest-api/pipelines#add-a-webhook)? If I understand correctly `pipeline` on Buildkite is equivalent of `project` on Semaphore, but I might be wrong. 

@AleksandarCole Unfortunately these are incoming webhooks. To quote GraphQL API doc:

> Create a webhook on the repository for a pipeline. This mutation creates a webhook on the pipeline's source repository (e.g., GitHub) that will trigger builds when code is pushed. Requires a GitHub App integration to be configured for the organization. Returns an error if a webhook already exists for the repository.

So, they work like this: `GitHub > Buildkite`

AleksandarCole:
@immz4 understood. Can we then work on improving the UX of configuring this? I did not manage to connect it based on the instructions in the UI. Even after reviewing the video I'm not clear how to connect this. 

We can ask the user to input the link to their organization (eg. `https://buildkite.com/superplane`) and then based on that generate links like `https://buildkite.com/organizations/superplane/services/webhook/new` and `https://buildkite.com/user/api-access-tokens/new` to point them to the pages instead of them having to figure it out based on descriptions. 

We also need to generate webhook url for them, I am unsure what this means `Webhook URL is constructed in following way - https://<Backend URL>/api/v1/integrations/<Integration ID>/webhook.`

immz4:
@AleksandarCole addressed UX, added setup helpers

Integration setup flow:

<div align="center" width="100%">
   <hr>
  image
</div>
<div align="center">
  <hr>
  image
  <p><i>"Continue" button leads to Personal Settings > API Access Token</i></p>
</div>
<div align="center">
  <hr>
  image
  <p><i>"Continue" button opens "Add Webhook Notification" setup</i></p>
</div>

immz4:
I can also record new video to demonstrate UX changes.

AleksandarCole:
Hey @immz4 this is much better, thanks! I've managed to run it - great direction, I love that you included those env vars + metadata. 

Have few changes that we need to make and one problem I came across:


### Environment variable broken
When I add environment variable I cannot save the component:
image

### Wrong state on fail
If the pipeline fails state should not be `Error` but `Fail` instead, since `Fail` is a valid result that emits event further. `Error` is reserved for the cases when something goes wrong (eg. didn't manage to start the build) and it should not emit an event further through any of the output channels. Check how Semaphore component handles this. 

image

### Wrong component icon
Icons render as puzzle (because they are not using BuildKite icon)

### Build metadata not set
Even though I've configured build metadata it doesn't seem to be set
image
image


### Update details tab
In the details tab, let's have two timestamps at the top (Started at, and Completed at), and Build URL, feel free to delete the rest. 

image

From the trigger component detail tab, let's also have timestamp at the top (single one in this case) and remove Build State (or result, having both is redundant), Build Number, Project, Organization (those are empty)

### Integration Setup UX
The organization URL is prefilled with user email (which is very strange) and also API token is prefilled with something also. Let's make sure those are empty and that user needs to add them. 
image


The instructions for creating the webhook appear only after I save the integration for the first time (with API token) - any way for them to be there from the start - or appear once user enters the url?

Also, would be good if we had a link in the description for API token that immediately leads you to https://buildkite.com/user/api-access-tokens as it's a fixed url (same for everyone).

immz4:
Hey, @AleksandarCole, regarding your comments 

> The organization URL is prefilled with user email (which is very strange) and also API token is prefilled with something also. Let's make sure those are empty and that user needs to add them.

That is indeed very strange, since I do not have this problem.

1) Do you have the same issue with Semaphore integration setup?
2) What browser and password manager do you use? They sometimes can autofill in the wrong fields.

pkg/integrations/buildkite/trigger_build.go
```diff
@@ -0,0 +1,523 @@
...
+
+	if dataMap != nil {
+		metadata.Extra = dataMap
+		if err := ctx.Metadata.Set(metadata); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
immz4: I'm resolving this because GitHub randomly focuses the page on this comment field.

AleksandarCole:
@immz4 it could be the autofill that mixed up the fields and populated them with login username/password. This is strange though, as it doesn't happen for other integration forms. 

Let me know when you address the other notes/issues and I will test this again on a different browser.

immz4:
@AleksandarCole Here are the changes as requested:

<div>
  <h3> Fix for env and metadata </h3>
  image
  image
  image
</div>

-----

<div>
  <h3>Buildkite logo</h3>
  image
  image
</div>

-----

<div>
  <h3>"Failed" state instead of "Error"</h3>
  image
</div>

-----

<div>
  <h3>New details</h3>
  image
  image
  <p><i>I've added both dates to trigger, not sure which one of them do you want</i></p>
</div>

-----

> The instructions for creating the webhook appear only after I save the integration for the first time (with API token) - any way for them to be there from the start - or appear once user enters the url?

Hm, currently I do not think so. Webhook link contains ID of the integration, so before we can generate one we need to save the integration.

AleksandarCole:
@immz4 thank you for the corrections. 
Few smaller edits, and we're ready for code review:

### Organization in config
There's no need to have the option to select organization on the component level - we ask you for the link to your organization when configuring integration - that is the organization all components that use that integration should work with. 

### Persist webhook details
After you've configured the webhook token and successfully connected integration - the instructions along with the webhook link disappear. Let's keep those instructions always available bellow the webhook token regardless of the integration state (pending or connected) - this is in case user deletes the webhook on BuildKite side or misconfigures it - so that they can always know how to configure it again. 

### Component node metadata - wrong icon
Let's use same icon for `pipeline` for triggerPipeline and onBuildFinished - the one on onBuildFinished is better than the branch icon that triggerPipeline uses

image

immz4:
@AleksandarCole Done.

image

image

AleksandarCole:
@immz4 awesome. Moving it to code review.

pkg/integrations/buildkite/buildkite.go
```diff
@@ -0,0 +1,405 @@
...
+type WebhookMetadata struct {
+	URL             string   `json:"url"`
+	Token           string   `json:"token"`
+	VerifySignature bool     `json:"verifySignature"`
+	Active          bool     `json:"active"`
+	Events          []string `json:"events"`
+	Pipelines       []string `json:"pipelines"`
+}
+
+func (b *Buildkite) SetupWebhook(ctx core.WebhookContext) (any, error) {
```
lucaspin: This method is not used at all

pkg/integrations/buildkite/on_build_finished.go
```diff
@@ -0,0 +1,335 @@
...
+
+func (t *OnBuildFinished) Actions() []core.Action {
+	return []core.Action{}
+}
+
+func (t *OnBuildFinished) HandleAction(ctx core.TriggerActionContext) (map[string]any, error) {
+	return nil, nil
+}
+
+func (t *OnBuildFinished) HandleWebhook(ctx core.WebhookRequestContext) (int, error) {
```
lucaspin, line 210: This method is not used, right?

pkg/integrations/buildkite/buildkite.go
immz4: Old logic, removed.

pkg/integrations/buildkite/on_build_finished.go
immz4: Not used, removed.

pkg/integrations/buildkite/buildkite.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	eventType := ctx.Request.Header.Get("X-Buildkite-Event")
+	if eventType == "" {
+		if event, ok := payload["event"].(string); ok {
+			eventType = event
+		}
+	}
+
+	if eventType == "build.finished" {
```
lucaspin: BuildKite supports other types of events too: https://buildkite.com/docs/apis/webhooks#event-families. To make it possible for those types of events to be received in SuperPlane too in the future, we need to include the eventType in the `BuildkiteSubscriptionConfiguration`

AleksandarCole: Hey @immz4 please check this last note from @lucaspin, feels like we're close to wrapping this one up, and I would like to merge this PR.

immz4: Added.

immz4: @AleksandarCole addressed the note, if there is anything more I'm available on Monday.

lucaspin:
@immz4 after looking a bit deeper on this one, it seems clear to me that the webhook setup should be done as part of the trigger setup itself, and not the integration setup. A few considerations that led me to this conclusion:
- What if we want to add an `onAgent` trigger next? We would either have to (1) tell to user to update the webhook set up as part of the integration, or (2) make the user create the webhook with all the possible event types in the first place. (1) is confusing, and (2) would lead us to receive events for everything even if the user hasn't even create a trigger.
- Also, what if we want to have a trigger for [test engine alarms](https://buildkite.com/docs/apis/webhooks/test-engine) or [new packages](https://buildkite.com/docs/apis/webhooks/package-registries)? Here, it's not OK to be part of the integration, because those webhooks are scoped to test suite and package registry, respectively. So, we would end up with having to set up webhook during the integration setup, and also during trigger setup, for some triggers. That's inconsistent, and confusing.

---

For reference on how to handle webhook setup as part of trigger setup, take a look at the [dockerhub.onImagePush](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/dockerhub/on_image_push.go) trigger.

immz4:
@lucaspin Moved webhook setup to trigger.

Current flow: 

image

User needs to select the pipeline and save the trigger in order to get webhook url and token

---

image

After that user can click a button and copy over params to Buildkite

immz4:
@lucaspin @AleksandarCole Hi, any updates on PR merge?

AleksandarCole:
Hey @immz4, it looks good, our internal team will need to do some final polish and updates before the merge, but I think this contract is completed successfully 👍 
Thank you for your contribution and work invested 🙌

shiroyasha:
Unfortunately, we didn't review this one on time and it got stale.
Closing it.

Let's use this PR as inspiration once we re-open this topic.

Issue #1956, opened 2026-01-30
[Bitbucket] Base


AleksandarCole:
### Description

Create a **SuperPlane integration for Bitbucket** and **three starter components** (two triggers, one action) so users can build workflows with Bitbucket (Git-based code hosting and collaboration platform from Atlassian) in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://bitbucket.org/

### Connection Method

Part of this task is exploring and determining the best approach to connect Bitbucket with SuperPlane. Consider the available authentication options (App Passwords, OAuth 2.0, personal access tokens) and evaluate which method provides the best balance of security, ease of setup, and functionality for our use cases. Also determine the best approach for webhook setup to receive repository events.

### First Components

‼️ **IMPORTANT**

Creating at least 2 **trigger** and 1 **action** components (3 components total) to test out the Base is required for completing this issue.

> **Reference**: SuperPlane already has a GitHub integration with similar components (On Push, On Pull Request, etc.) in `pkg/integrations/github/`. Use it as a reference for patterns, webhook handling, and frontend mappers.

#### 1. On Push (Trigger)

A trigger that starts a workflow when code is pushed to a Bitbucket repository. Should support filtering by repository and branch/tag refs. The [Bitbucket webhooks](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-webhooks/) and the `repo:push` event are good starting points.

#### 2. On Pull Request (Trigger)

A trigger that starts a workflow when a pull request event occurs (created, updated, approved, merged, declined, etc.). Should support filtering by repository and action type. The `pullrequest:*` [webhook events](https://support.atlassian.com/bitbucket-cloud/docs/manage-webhooks/) are a good starting point.

#### 3. Create Pull Request (Action)

An action that creates a pull request in a Bitbucket repository. Should support source/destination branches, title, description, and reviewers. The [Pull Requests API](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/) is a good starting point.

### Acceptance Criteria (for SuperPlane team)

- [ ] integration works
- [ ] starter components work
- [ ] has proper tests
- [ ] has proper documentation
- [ ] passed code quality review
- [ ] passed functionality review
- [ ] passed ui/ux review

---

⚠️ **DISCLAIMER**: Pull requests submitted without a video demonstration will be automatically closed.

ishaksebsib:
Working on this 👍

PR #3039, opened 2026-02-11
feat: Add Bitbucket integration

ishaksebsib:
Implements #1956.

  This PR adds the Bitbucket integration with an On Push trigger.
  It supports workspace authentication via email + API token, lists repositories, and provisions webhooks for push events.
  The trigger supports ref filtering via predicates and emits Bitbucket push payloads to workflows.

  Demo:

https://github.com/user-attachments/assets/b58bf05a-7529-4de7-bb70-21a8b6675c9c

### DIFF
--- a/api/swagger/superplane.swagger.json
+++ b/api/swagger/superplane.swagger.json
@@ -4194,6 +4194,10 @@
         },
         "itemLabel": {
           "type": "string"
+        },
+        "maxItems": {
+          "type": "integer",
+          "format": "int32"
         }
       }
     },
--- a/docs/components/Slack.mdx
+++ b/docs/components/Slack.mdx
@@ -16,6 +16,7 @@ import { CardGrid, LinkCard } from "@astrojs/starlight/components";
 
 <CardGrid>
   <LinkCard title="Send Text Message" href="#send-text-message" description="Send a text message to a Slack channel" />
+  <LinkCard title="Wait for Button Click" href="#wait-for-button-click" description="Send a message with buttons and wait for the user to click one" />
 </CardGrid>
 
 ## Instructions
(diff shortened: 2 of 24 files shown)


pkg/integrations/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (b *Bitbucket) Description() string {
+	return "React to events in your Bitbucket repositories"
+}
+
+func (b *Bitbucket) Instructions() string {
+	return ""
+}
+
+func (b *Bitbucket) Configuration() []configuration.Field {
```
lucaspin, line 76: I think we need to support [workspace access tokens](https://support.atlassian.com/bitbucket-cloud/docs/workspace-access-tokens/) here too. Maybe we can have an `authType` configuration field with two options: API token / Workspace access token.

Check the [PagerDuty integration](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/pagerduty/pagerduty.go#L63). We do something similar there.

pkg/integrations/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,134 @@
...
+	// Validate credentials by listing repositories.
+	//
+	repos, err := client.ListRepositories(config.Workspace)
+	if err != nil {
+		return fmt.Errorf("error listing repositories: %w", err)
+	}
+
+	ctx.Integration.SetMetadata(Metadata{
+		Workspace:    config.Workspace,
+		Repositories: repos,
```
lucaspin: Not sure about keeping the repositories in the metadata here. For GitHub, we do that because that's the only way to know which repositories are actually authorized for the GitHub App, and we get events about new repositories being authorized. Here, since we are using an API token or access token, we can just list the repositories when we need to, which should just be when creating the node in the UI. Otherwise, we'd need to refresh the repositories in the metadata.

lucaspin: To verify that the API token / access token passed is OK, we can just describe the workspace. That would also us to store more information about the workspace in the metadata, like ID, URL, ...

pkg/integrations/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,134 @@
...
+func (b *Bitbucket) Icon() string {
+	return "bitbucket"
+}
+
+func (b *Bitbucket) Description() string {
+	return "React to events in your Bitbucket repositories"
+}
+
+func (b *Bitbucket) Instructions() string {
+	return ""
```
lucaspin: We need some instructions here about where / how to create the API token / workspace access token.

pkg/integrations/bitbucket/bitbucket.go
ishaksebsib: Okay, I'll look into it

pkg/integrations/bitbucket/bitbucket.go
ishaksebsib: The idea was to store the repositories in metadata to avoid calling the API every time a node is created. But you're right, that would mean we wouldn’t get newly added repos unless we refresh. I’ll update it accordingly.

lucaspin:
@ishaksebsib I made some changes to support the workspace access tokens. Can you sign-off the commits you have on this branch? Check [this document](https://github.com/superplanehq/superplane/blob/main/docs/contributing/commit_sign-off.md).

ishaksebsib:
@lucaspin Thanks again for your help! I’ve signed off on the commits and added instructions for adding the tokens.

lucaspin:
@ishaksebsib thanks. Everything looks good now. Will merge once CI passes

ishaksebsib:
Thanks again @lucaspin, I will send another PR for the other triggers!

PR #3072, opened 2026-02-12
chore: WebhookContext is now available for action components in the Setup and Execute callback function

shiroyasha:
Needed in places where we don't have access to setting up a webhook via an API.
e.g. In Cursor.


atelabro (commented):
Looks clean, lets get this merged <3

Issue #2552, opened 2026-02-01
[New Relic] Base


AleksandarCole:
### Description

Create a **SuperPlane integration for New Relic** and **three starter components** (one trigger, two actions) so users can build workflows with New Relic (observability and APM platform) in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://newrelic.com/

### Connection Method

Part of this task is exploring and determining the best approach to connect New Relic with SuperPlane. Consider the available authentication options and evaluate which method provides the best balance of security, ease of setup, and functionality for our use cases.

### First Components

‼️ **IMPORTANT**

Creating at least 1 **trigger** and 2 **action** components (3 components total) to test out the Base is required for completing this issue.

#### 1. On Issue (Trigger)

A trigger that starts a workflow when a New Relic Issue occurs so users can run incident response and notifications. Webhook-based notification destinations are a good starting point.

#### 2. Report Metric (Action)

An action that sends metric data from the workflow to New Relic so it appears in dashboards and can be used for alerting. New Relic’s [Metric API](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/report-metrics-metric-api/) is a good starting point.

#### 3. Run NRQL Query (Action)

An action that runs a query against New Relic and returns the results to the workflow so downstream steps can use the data. [NerdGraph and NRQL](https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-nrql-tutorial/) is a good starting point.

### Acceptance Criteria (for SuperPlane team)

- [ ] integration works
- [ ] starter components work
- [ ] has proper tests
- [ ] has proper documentation
- [ ] passed code quality review
- [ ] passed functionality review
- [ ] passed ui/ux review

---

⚠️ **DISCLAIMER**: Pull requests submitted without a video demonstration will be automatically closed.

yotor1221:
I’m working on this integration and have started implementation. Happy to take ownership.

PR #3077, opened 2026-02-12
feat: new relic integration

yotor1221:
feat: Add New Relic integration OnIssue ,runNRQL and reportmetric

## Description

Implements #2552.

This PR implements the native New Relic integration, which allows users to trigger workflows based on New Relic issues (`OnIssue`), execute deep telemetry queries (`RunNRQLQuery`), and report custom business metrics (`ReportMetric`) back to New Relic.

Authorization is via API keys:
*   **RunNRQLQuery**: Requires a **User API Key** (starts with `NRAK-`).
*   **ReportMetric**: Recommended to use with a **License Key** (Ingest - License).

## a video demo

https://www.loom.com/share/b5554aa5781e49e98b55eb0bab80a9da

## Backend Implementation

The backend implementation code is in `pkg/integrations/newrelic/`.

*   **Structure**: Follows existing patterns using `newrelic.go` for registration and separate files for triggers/actions (`on_issue.go`, `run_nrql_query.go`, `report_metric.go`).
*   **Key Features**:
    *   **GPG Signing**: All commits are GPG signed.
    *   **Configuration Persistence**: Fixed UI state issues by enforcing strict string typing in `OnIssue` and `RunNRQLQuery` structs.
    *   **Smart Templates**: Added logic to intercept unresolved template tags (e.g., `{{account_id}}`) to prevent invalid API calls.
*   **Example Output**: JSON examples for component outputs are included in the code comments and documentation.

## Frontend Implementation

This integration uses the Server-Driven UI (SDUI) pattern. The frontend forms are automatically generated from the Go configuration structs defined in the backend.

*   **Persistence Fix**: ensuring field types (e.g., `FieldTypeIntegrationResource`) map correctly to string inputs in the UI has resolved previous selection issues.

## Docs

Documentation has been generated in `docs/components/New Relic.mdx` by running `make gen.components.docs`.

It includes:
*   Setup instructions (API Key generation).
*   Usage guides for `OnIssue`, `Run NRQL Query`, and `Report Metric`.
*   Example configuration and data payloads.

## Tests

Unit tests are located in `pkg/integrations/newrelic/`.

*   `run_nrql_query_test.go`: Covers query execution, template variable validation, and context fallback logic.
*   `on_issue_test.go`: Covers webhook validation, payload parsing, and event filtering.
*   `repro_test.go`: Specifically tests UI configuration persistence and type handling.
*   `report_metric_test.go`: Covers metric payload construction.

Tests are deterministic and pass locally.


AleksandarCole:
@yotor1221 thanks for the video instructions - very clear and useful! 

The main issue I see right now is that we need 2 types of API keys. There is no sense in user switching this in configuration. 

We should have an integration form that asks you to create both keys - explaining which is used for which. User can configure only one or the other (which is fine) - but then the component needs to know if configuration has what it needs. 

For example, if I have user API key configured only - the runNRQL and onIssue need to show configuration as green while reportMetric shows it as orange (not configured)

___

I will assign engineering to this one for a quick review of the approach, before I proceed with more in-depth functional/UX testing.

yotor1221:
> @yotor1221 thanks for the video instructions - very clear and useful!
> 
> The main issue I see right now is that we need 2 types of API keys. There is no sense in user switching this in configuration.
> 
> We should have an integration form that asks you to create both keys - explaining which is used for which. User can configure only one or the other (which is fine) - but then the component needs to know if configuration has what it needs.
> 
> For example, if I have user API key configured only - the runNRQL and onIssue need to show configuration as green while reportMetric shows it as orange (not configured)
> 
> I will assign engineering to this one for a quick review of the approach, before I proceed with more in-depth functional/UX testing.

Hi @AleksandarCole
Thanks for the feedback. That makes total sense splitting the keys in the integration form will definitely provide a better UX. I'll update the configuration to support both User API Keys (for NRQL/Triggers) and License Keys (for Metrics) simultaneously.

I will also implement the validation logic so the components visually reflect their 'ready' status based on which keys are provided. I'll ping you once the updated flow is ready for review.

yotor1221:
Hi @AleksandarCole ,

I have updated the integration to address the API key configuration and improve the setup UX:

Dual-Key Support: The integration now accepts both userApiKey and licenseKey simultaneously. Users no longer need to switch between them manually.

Component-Specific Logic: Components now independently verify their required keys. If only a User API Key is provided, RunNRQLQuery and OnIssue will initialize successfully, while ReportMetric will identify the missing License Key during setup.

Simplified Configuration: I have removed the manual Account ID field from the NRQL component. Since the integration is already managing this via the dynamic picker, the manual input was redundant. This prevents user error and streamlines the configuration process.

Clear Feedback: While the core UI doesn't currently support an 'Orange' status for partial configurations, I have implemented explicit validation in Setup(). If a required key is missing, the component displays a clear error: 'User API Key/License Key is required for this component. Please configure it in the Integration settings.' This ensures the user knows exactly which credential is missing without affecting other working components.

pkg/integrations/newrelic/run_nrql_query.go
```diff
@@ -0,0 +1,373 @@
...
+			Required:    true,
+			Description: "The NRQL query to execute",
+			Placeholder: "SELECT count(*) FROM Transaction SINCE 1 hour ago",
+		},
+		{
+			Name:        "timeout",
+			Label:       "Timeout (seconds)",
+			Type:        configuration.FieldTypeNumber,
+			Required:    false,
+			Description: "Query timeout in seconds (default: 10, max: 120)",
```
lucaspin: Not so sure about allowing timeout to be configured here, specially a 2min timeout. Seems like NewRelic supports [asynchronous queries](https://docs.newrelic.com/docs/apis/nerdgraph/examples/async-queries-nrql-tutorial/). Could we use those instead, and a fixed timeout of 10s?

If I understand things correctly, with that, if the query returns before 10s, we get a response right away. Otherwise, we get something can poll for the result. For polling, we can use `ctx.Requests.ScheduleActionCall()` to make polling completely asynchronous as well.

yotor1221: hi @lucaspin
Good catch. You're right handling this asynchronously is definitely the better path for long-running NRQL queries.

I'm working on it now. I'll refactor it to use the NerdGraph async API with a fixed 10s initial timeout and move the rest to ScheduleActionCall. This should also let us handle the retryAfter and retryDeadline fields properly to keep the polling efficient.

yotor1221:
Hi @lucaspin 

I've updated the NRQL query logic to support asynchronous execution as suggested.

Fixed Timeout: Set a 10s timeout for the initial request.

Async Fallback: If the query doesn't complete within that window, the integration now captures the New Relic queryId.

Non-blocking Polling: I implemented ctx.Requests.ScheduleActionCall() to poll for the results. This ensures the execution doesn't hang and follows the project's async pattern.

Configuration: Removed the user-configurable timeout to keep the implementation clean and predictable.

The code passed all the tests . Ready for another look

yotor1221:
@lucaspin 
I've updated the New Relic integration. Here’s what’s included:

Security: Implemented webhook authentication using ctx.Webhook.GetSecret() with constant-time comparison to prevent timing attacks.

Client Fix: Resolved the issue with encrypted credentials in the New Relic client logic to ensure proper decryption and API connectivity.

Docs: Regenerated the component documentation to include the new X-Superplane-Secret header requirement.

Tests: Added new test cases for authenticated/unauthenticated requests (all local tests passing).

The DCO is now signed, and I've synced the latest changes from main. Ready for final look

lucaspin:
@yotor1221 a couple of additional comments:

- There is no icon for NewRelic in the UI. Also, we need to include UI mappers for the new components and trigger. Check https://github.com/superplanehq/superplane/tree/main/web_src/src/pages/workflowv2/mappers
- I don't think the onIssue trigger is OK as it is. It looks like I have to set up the alert destination manually, but the URL for doing so is not presented anywhere to the user. Also, it looks like it is possible to [create the alert destination via their API](https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-api-notifications-destinations/), so we should do that instead of requiring the user to do it.

yotor1221:
> @yotor1221 a couple of additional comments:
> 
> * There is no icon for NewRelic in the UI. Also, we need to include UI mappers for the new components and trigger. Check https://github.com/superplanehq/superplane/tree/main/web_src/src/pages/workflowv2/mappers
> * I don't think the onIssue trigger is OK as it is. It looks like I have to set up the alert destination manually, but the URL for doing so is not presented anywhere to the user. Also, it looks like it is possible to [create the alert destination via their API](https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-api-notifications-destinations/), so we should do that instead of requiring the user to do it.

@lucaspin 
Thanks for the feedback I've addressed these points in the latest push:

UI/UX: Added the New Relic icon and implemented all UI mappers for the triggers and actions in web_src/src/pages/workflowv2/mappers/newrelic.

Automated Setup: The onIssue trigger now fully automates the setup. It uses the NerdGraph API to programmatically create the Alert Destination and Notification Channel when the trigger is saved. Users no longer need to configure anything manually in the New Relic UI.

Cleanup: I’ve ensured that the logic is consistent with our other integrations, though I've kept the destination deletion out for now as discussed to avoid accidental data loss.

Ready for another look!

AleksandarCole:
Implemented and merged in another PR.

Issue #3003, opened 2026-02-10
[Statuspage] Base


AleksandarCole:
Create a **SuperPlane integration for Statuspage** (Atlassian) and **three starter components** so users can build workflows with Statuspage in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://www.atlassian.com/software/statuspage (API: https://developer.statuspage.io/)

### Connection Method

Part of this task is exploring and determining the best approach to connect Statuspage with SuperPlane.

**Suggested starting point**: Explore Statuspage API docs and the codebase—look at existing integrations and components (e.g. PagerDuty) for patterns.

Consider the available options and evaluate what provides the best balance of security, ease of setup, and functionality for our use cases. Statuspage supports **webhooks** for incident events (created, updated, resolved) and a **REST API** (API token, page-scoped) for creating/updating incidents, components, and subscribers. Auth is API token (OAuth header or query param). Rate limit: 1 req/sec per token. See [Statuspage API](https://developer.statuspage.io/).

### First Components

‼️ **IMPORTANT**

Creating at least 3 `action` components to test out the Base is required for completing this issue.

Configuration options, data displayed in the payload, and component node meta (details) are for the implementer to suggest based on what the tool offers; you can explore similar components in the codebase for inspiration.

#### 1. Create Incident (Action)

Creates a new incident on Statuspage via the REST API (name, body, affected components, type: realtime or scheduled). Enables workflows like creating or updating the status page from PagerDuty, monitoring alerts, or other SuperPlane triggers—similar to the existing PagerDuty Create Incident action.

#### 2. Update Incident (Action)
Let's the user update the incident status on the Statuspage (eg. `In progress`/`Completed`) along with the message

#### 3. Get Incident (Action)
Gets back the details of the specific incident (timeline etc.)

### Acceptance Criteria (for SuperPlane team)

* [ ] integration works
* [ ] starter components work
* [ ] has proper tests
* [ ] has proper documentation
* [ ] passed code quality review
* [ ] passed functionality review
* [ ] passed ui/ux review

prudh-vi:
@AleksandarCole assign this to me

PR #3117, opened 2026-02-14
feat: statuspage integration + 3 components

re-pixel:
Implements #3003 

### Description

This PR adds the Statuspage integration to SuperPlane with 3 action components:

**Create Incident (action)**: Creates new realtime or scheduled incidents on Atlassian Statuspage. Supports incident name, body, status, impact override, affected components, component status, and for scheduled incidents: start/end times, timezone, reminders, and auto-transitions.

**Update Incident (action)**: Updates existing incidents with new status, body, impact override, and component associations. Supports all incident statuses (investigating, identified, monitoring, resolved, scheduled, in_progress, completed).

**Get Incident (action)**: Fetches incident details including the full timeline (incident_updates) for use in workflows.

Authentication uses an API key (OAuth). Users can create one in Statuspage at Settings > API info. The integration uses the Statuspage REST API (https://api.statuspage.io/v1/) for all operations.
On setup, the integration syncs available pages and components from Statuspage for use in resource pickers in the UI.

### Changes
**Backend (pkg/integrations/statuspage/)**

- statuspage.go — Integration entry point, API key auth, sync (ListPages)
- client.go — REST API client (pages, components, create/update/get incident)
- common.go — Shared models (NodeMetadata)
- create_incident.go — Create incident action (realtime and scheduled)
- update_incident.go — Update incident action
- get_incident.go — Get incident action
- list_resources.go — Resource discovery for UI pickers (pages, components)
- api-documentation/ — API reference (statuspage-api-guide.md, swagger.json)

**Frontend (web_src/src/pages/workflowv2/mappers/statuspage/)**

- Trigger renderer and component mappers for create, update, and get incident
- Icon and sidebar registration (integrationIcons.tsx, BuildingBlocksSidebar)
- Timeline display for incident updates in the Details tab

**Docs**

- docs/components/Statuspage.mdx — Component documentation

**Tests**

- create_incident_test.go — Setup validation and execution
- update_incident_test.go — Setup validation and execution
- get_incident_test.go — Setup validation and execution
- list_resources_test.go — Resource listing

Demo: https://drive.google.com/file/d/1dOXwgdMXQHxXVS7pLDu0GL7BChPGnvmL/view?usp=sharing


AleksandarCole:
@re-pixel - works quite well! Here are some UX things that we could address before we switch to code review:

### Details tabs
Let's give this one a bit of love, overall it looks ok but some general changes (across all components):
- Timestamps should be at the top
- No need to have two (created/updated) having two makes sense only for the things that take a bit longer to execute like running a CI pipeline (started at/finished at). 
- I'm not sure about the timeline thing - makes sense on get and update, but not on create
image

### Component node meta
Having the `Page` on all of them makes sense, however, I am not sure about the rest of the stuff that's displayed here. 
The idea behind component node meta is to display few key configurations (no more than 3 ideally) that will help user visually identify and understand what the component does without having to open configuration in the sidebar. 
Note that we do have the option of displaying extra data on hover (check timegate component example)

### Update incident UX
The Status page has two types if incidents (scheduled or realtime), you cannot switch scheduled to realtime and vice versa. If we have a dropdown to select one of those 2 modes with `Create` component, we should follow the same pattern with `Update`. Doesn't make sense for user to be able to set the Status to `Scheduled` and Impact to `Critical`

### Expression support
Some of these configs seem like they would benefit from supporting expressions (dynamic data from payloads instead of static). `Status`, `Impact` etc. Check how `repository` selector is handled for GitHub components - there we give you a singleSelect by default but you can switch to dynamic expression based setup too.

AleksandarCole:
@re-pixel Looks good, moved to code review.

pkg/integrations/statuspage/common.go
```diff
@@ -0,0 +1,131 @@
...
+			ids = append(ids, resolved)
+			seen[resolved] = true
+		}
+		statusByID[resolved] = status
+	}
+	return ids, statusByID, nil
+}
+
+// resolveMetadataSetup fetches page and component names from the API when pageID and componentIDs
+// are static (no expressions). Returns empty metadata if HTTP context is unavailable or on error.
```
forestileao: `Setup()` is expected to verify integration resources when IDs are static. Here the helper returns empty metadata on API errors or missing resources, so invalid page/component selections can pass setup and fail only at runtime. Consider returning an error when static IDs cannot be resolved.

pkg/integrations/statuspage/get_incident.go
```diff
@@ -0,0 +1,184 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				Resource: &configuration.ResourceTypeOptions{
+					Type: ResourceTypePage,
+				},
+			},
+		},
+		{
+			Name:        "incident",
+			Label:       "Incident ID",
+			Type:        configuration.FieldTypeString,
```
forestileao: Component-review rule: when users can choose existing resources, prefer `FieldTypeIntegrationResource`. This `incident` field is free text, but incidents are listable (`Client.ListIncidents`). This should be modeled as an integration resource to reduce invalid IDs.

pkg/integrations/statuspage/update_incident.go
```diff
@@ -0,0 +1,437 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				Resource: &configuration.ResourceTypeOptions{
+					Type: ResourceTypePage,
+				},
+			},
+		},
+		{
+			Name:        "incident",
+			Label:       "Incident ID",
+			Type:        configuration.FieldTypeString,
```
forestileao: Same configuration concern as `getIncident`: this uses a free-text incident ID even though incidents are existing resources. Prefer `FieldTypeIntegrationResource` with page-scoped lookup so setup can validate and users can select instead of typing IDs.

pkg/integrations/statuspage/create_incident.go
```diff
@@ -0,0 +1,584 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				List: &configuration.ListTypeOptions{
+					ItemLabel: "Component",
+					ItemDefinition: &configuration.ListItemDefinition{
+						Type: configuration.FieldTypeObject,
+						Schema: []configuration.Field{
+							{
+								Name:        "componentId",
+								Label:       "Component ID",
+								Type:        configuration.FieldTypeString,
```
forestileao, line 220: `componentId` is modeled as a plain string. Since components are existing resources and `ResourceTypeComponent` already exists, prefer an `IntegrationResource` field (page-scoped) to avoid typo-driven runtime failures and improve discoverability.

pkg/integrations/statuspage/create_incident.go
```diff
@@ -0,0 +1,584 @@
...
+		if spec.ScheduledUntil == nil || *spec.ScheduledUntil == "" {
+			return errors.New("scheduledUntil is required for scheduled incidents")
+		}
+
+		// Validate scheduledFor is before scheduledUntil
+		tz := "UTC"
+		if spec.ScheduledTimezone != nil && *spec.ScheduledTimezone != "" {
+			tz = *spec.ScheduledTimezone
+		}
+		parsedFor, errFor := toUTCISO8601(derefStr(spec.ScheduledFor), tz)
```
forestileao: `Setup()` ignores datetime parse failures (`errFor` / `errUntil`) and only checks ordering when both parses succeed. Invalid static datetime/timezone values can pass setup and fail later in execute; setup should return a validation error in this branch.

pkg/integrations/statuspage/create_incident.go
re-pixel: @forestileao I couldn't find examples of `IntegrationResource` fields inside a `ListField` in any of the current integrations. That's why I left these as plain strings. I couldn't find a way to squeeze `IntegrationResource` fields into lists without making changes out of the scope of status page integration.

Any recommendation on how to approach this?

web_src/src/pages/workflowv2/mappers/statuspage/utils.ts
```diff
@@ -0,0 +1,176 @@
...
+  return [
+    {
+      receivedAt: new Date(execution.createdAt!),
+      eventTitle: title,
+      eventSubtitle: formatTimeAgo(new Date(execution.createdAt!)),
+      eventState: getState(componentName)(execution),
+      eventId: execution.rootEvent.id,
+    },
+  ];
+}
```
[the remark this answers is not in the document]
re-pixel: This comment requests changes across different integrations. I don't think I should resolve it now.

pkg/integrations/statuspage/create_incident.go
forestileao: @re-pixel You are correct, this is inside a List field, my bad. I don't think we support this.

pkg/integrations/statuspage/get_incident.go
```diff
@@ -0,0 +1,228 @@
...
+		"name":          "Database Connection Issues",
+		"status":        "investigating",
+		"impact":        "major",
+		"shortlink":     "http://stspg.io/p31zjtct2jer",
+		"created_at":    "2026-02-12T10:30:00.000Z",
+		"updated_at":    "2026-02-12T10:30:00.000Z",
+		"page_id":       "kctbh9vrtdwd",
+		"component_ids": []string{"8kbf7d35c070"},
+		"incident_updates": []map[string]any{
+			{
```
forestileao: I didnt notice this before, but check other examples, all outputs in separate json files and all example output or example data functions are in example.go files

forestileao: Also, note data they are all wrapped by 
{
"data": {...},
"type": "example",
"timestamp": "example"
}

so please copy and paste from the payload sidebar of what the component emmited and fake the data.

re-pixel: resolved the issue by adding example.go and output example json-s for each component

omarsoufiane:
@re-pixel Hello, I'm from bountyhub,
can you setup you payment information from bountyhub's dashboard? click on connnect stripe from dashboard/get-paid
it's not a regular stripe account that you'll use on a website, it's a connected account.
we use paypal only if it's not available in your country

re-pixel:
> @re-pixel Hello, I'm from bountyhub, can you setup you payment information from bountyhub's dashboard? click on connnect stripe from dashboard/get-paid it's not a regular stripe account that you'll use on a website, it's a connected account. we use paypal only if it's not available in your country

Hi @omarsoufiane, thanks for reaching out. 
I'm from Serbia and Stripe isn't available here at the moment. I tried to connect paypal but i didn't recieve any confirmation besides this popup in the left bottom corner.

image

omarsoufiane:
We do support  Serbia @re-pixel

re-pixel:
Oh my bad I assumed Serbia was not supported since Stripe itself doesn't support it, thanks for pointing that out. I set everything up now @omarsoufiane.

omarsoufiane:
@re-pixel transfer made, it's now in your Stripe dashboard and should arrive at your bank withing few days

PR #3118, opened 2026-02-14
feat: AWS ECS integration + 3 components

forestileao:
Issue: https://github.com/superplanehq/superplane/issues/2757

## Summary

This PR adds AWS ECS support to SuperPlane, including ECS actions, resource discovery, workflow mapper wiring, and docs.

### Backend (Go)
- Added ECS actions under `pkg/integrations/aws/ecs/`:
  - `DescribeService`
  - `RunTask`
  - `StopTask`
- Added ECS AWS client and request/response handling.
- Added ECS resource listing support:
  - `ecs.cluster`
  - `ecs.service`
  - `ecs.taskDefinition`
  - `ecs.task`
- Registered ECS components in AWS integration wiring (`pkg/integrations/aws/aws.go`) and resource switch (`pkg/integrations/aws/resources.go`).

### Frontend
- Added ECS mappers in `web_src/src/pages/workflowv2/mappers/aws/ecs/`:
  - `describe_service.ts`
  - `run_task.ts`
  - `stop_task.ts`
- Wired ECS mappers and event state registry in `web_src/src/pages/workflowv2/mappers/aws/index.ts`.
- Updated sidebar/icon handling so ECS components render correctly in workflow builder UI.

### Documentation
- Added/updated ECS component docs in `docs/components/AWS.mdx`.

## Testing

https://github.com/user-attachments/assets/ec1ac24d-ba12-460d-9276-10626a5db2ff

### DIFF
--- a/pkg/integrations/aws/ecs/run_task.go
+++ b/pkg/integrations/aws/ecs/run_task.go
@@ -338,6 +338,67 @@ func (c *RunTask) Configuration() []configuration.Field {
 			Default:     "{\"awsvpcConfiguration\":{\"subnets\":[],\"securityGroups\":[],\"assignPublicIp\":\"DISABLED\"}}",
 			Togglable:   true,
 			Description: "Optional ECS networkConfiguration object (for example, awsvpcConfiguration)",
+			TypeOptions: &configuration.TypeOptions{
+				Object: &configuration.ObjectTypeOptions{
+					Schema: []configuration.Field{
+						{
+							Name:     "awsvpcConfiguration",
+							Label:    "AWS VPC Configuration",
+							Type:     configuration.FieldTypeObject,
+							Required: false,
+							TypeOptions: &configuration.TypeOptions{
+								Object: &configuration.ObjectTypeOptions{
+									Schema: []configuration.Field{
+										{
+											Name:     "subnets",
+											Label:    "Subnets",
+											Type:     configuration.FieldTypeList,
+											Required: false,
+											TypeOptions: &configuration.TypeOptions{
+												List: &configuration.ListTypeOptions{
+													ItemLabel: "Subnet",
+													ItemDefinition: &configuration.ListItemDefinition{
+														Type: configuration.FieldTypeString,
+													},
+												},
+											},
+										},
+										{
+											Name:     "securityGroups",
+											Label:    "Security Groups",
+											Type:     configuration.FieldTypeList,
+											Required: false,
+											TypeOptions: &configuration.TypeOptions{
+												List: &configuration.ListTypeOptions{
+													ItemLabel: "Security Group",
+													ItemDefinition: &configuration.ListItemDefinition{
+														Type: configuration.FieldTypeString,
+													},
+												},
+											},
+										},
+										{
+											Name:     "assignPublicIp",
+											Label:    "Assign Public IP",
+											Type:     configuration.FieldTypeSelect,
+											Required: false,
+											Default:  "DISABLED",
+											TypeOptions: &configuration.TypeOptions{
+												Select: &configuration.SelectTypeOptions{
+													Options: []configuration.FieldOption{
+														{Label: "Disabled", Value: "DISABLED"},
+														{Label: "Enabled", Value: "ENABLED"},
+													},
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
 		},
 		{
 			Name:        "overrides",
@@ -347,6 +408,109 @@ func (c *RunTask) Configuration() []configuration.Field {
 			Default:     "{\"containerOverrides\":[]}",
 			Togglable:   true,
 			Description: "Optional ECS task overrides object",
+			TypeOptions: &configuration.TypeOptions{
+				Object: &configuration.ObjectTypeOptions{
+					Schema: []configuration.Field{
+						{
+							Name:     "containerOverrides",
+							Label:    "Container Overrides",
+							Type:     configuration.FieldTypeList,
+							Required: false,
+							TypeOptions: &configuration.TypeOptions{
+								List: &configuration.ListTypeOptions{
+									ItemLabel: "Container Override",
+									ItemDefinition: &configuration.ListItemDefinition{
+										Type: configuration.FieldTypeObject,
+										Schema: []configuration.Field{
+											{
+												Name:     "name",
+												Label:    "Name",
+												Type:     configuration.FieldTypeString,
+												Required: false,
+											},
+											{
+												Name:     "command",
+												Label:    "Command",
+												Type:     configuration.FieldTypeList,
+												Required: false,
+												TypeOptions: &configuration.TypeOptions{
+													List: &configuration.ListTypeOptions{
+														ItemLabel: "Argument",
+														ItemDefinition: &configuration.ListItemDefinition{
+															Type: configuration.FieldTypeString,
+														},
+													},
+												},
+											},
+											{
+												Name:     "environment",
+												Label:    "Environment",
+												Type:     configuration.FieldTypeList,
+												Required: false,
+												TypeOptions: &configuration.TypeOptions{
+													List: &configuration.ListTypeOptions{
+														ItemLabel: "Environment Variable",
+														ItemDefinition: &configuration.ListItemDefinition{
+															Type: configuration.FieldTypeObject,
+															Schema: []configuration.Field{
+																{
+																	Name:     "name",
+																	Label:    "Name",
+																	Type:     configuration.FieldTypeString,
+																	Required: true,
+																},
+																{
+																	Name:     "value",
+																	Label:    "Value",
+																	Type:     configuration.FieldTypeString,
+																	Required: false,
+																},
+															},
+														},
+													},
+												},
+											},
+											{
+												Name:     "cpu",
+												Label:    "CPU",
+												Type:     configuration.FieldTypeNumber,
+												Required: false,
+												TypeOptions: &configuration.TypeOptions{
+													Number: &configuration.NumberTypeOptions{
+														Min: func() *int { min := 0; return &min }(),
+													},
+												},
+											},
+											{
+												Name:     "memory",
+												Label:    "Memory",
+												Type:     configuration.FieldTypeNumber,
+												Required: false,
+												TypeOptions: &configuration.TypeOptions{
+													Number: &configuration.NumberTypeOptions{
+														Min: func() *int { min := 0; return &min }(),
+													},
+												},
+											},
+											{
+												Name:     "memoryReservation",
+												Label:    "Memory Reservation",
+												Type:     configuration.FieldTypeNumber,
+												Required: false,
+												TypeOptions: &configuration.TypeOptions{
+													Number: &configuration.NumberTypeOptions{
+														Min: func() *int { min := 0; return &min }(),
+													},
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
 		},
 		{
 			Name:        "timeoutSeconds",
--- a/docs/components/AWS.mdx
+++ b/docs/components/AWS.mdx
@@ -824,7 +824,6 @@ The Run Task component starts one or more ECS tasks and completes based on task 
 
 - For Fargate tasks, set **Network Configuration** using the ECS awsvpcConfiguration format.
 - Use **Capacity Provider Strategy** when you want ECS to choose capacity providers; it cannot be combined with **Launch Type**.
-- Optional ECS API fields can be passed directly through **Overrides** and **Network Configuration**.
 
 ### Example Output
 


pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,931 @@
...
+			Type:        configuration.FieldTypeBool,
+			Required:    false,
+			Default:     false,
+			Togglable:   true,
+			Description: "Enable ECS Exec support for the task",
+		},
+		{
+			Name:        "networkConfiguration",
+			Label:       "Network Configuration",
+			Type:        configuration.FieldTypeObject,
```
lucaspin, line 336: I think this should be typed as well, and not just any

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,931 @@
...
+			Required:    false,
+			Default:     "{\"awsvpcConfiguration\":{\"subnets\":[],\"securityGroups\":[],\"assignPublicIp\":\"DISABLED\"}}",
+			Togglable:   true,
+			Description: "Optional ECS networkConfiguration object (for example, awsvpcConfiguration)",
+		},
+		{
+			Name:        "overrides",
+			Label:       "Overrides",
+			Type:        configuration.FieldTypeObject,
```
lucaspin, line 406: Same comment about this being typed and not any

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,931 @@
...
+			Default:  "1",
+			TypeOptions: &configuration.TypeOptions{
+				Number: &configuration.NumberTypeOptions{
+					Min: func() *int { min := 1; return &min }(),
+					Max: func() *int { max := runTaskMaxCount; return &max }(),
+				},
+			},
+		},
+		{
+			Name:     "launchType",
```
lucaspin, line 235: What if I want to run it using a capacity provider instead?

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,931 @@
...
+	integrationMetadata := common.IntegrationMetadata{}
+	if err := mapstructure.Decode(ctx.Integration.GetMetadata(), &integrationMetadata); err != nil {
+		return RunTaskNodeMetadata{}, common.IntegrationMetadata{}, fmt.Errorf("failed to decode integration metadata: %w", err)
+	}
+
+	return metadata, integrationMetadata, nil
+}
+
+func retryRunTaskRuleAvailabilityCheck(ctx core.ActionContext, logMessage string, args ...any) error {
+	if ctx.Logger != nil {
```
lucaspin: No need to check if Logger is non-nil

lucaspin: Also, I think you might be over-using functions here. For example, `retryRunTaskRuleAvailabilityCheck` is just a log message + ctx.Requests.ScheduleActionCall() call. There's no need to put this in a separate function

pkg/integrations/aws/ecs/stop_task.go
```diff
@@ -0,0 +1,541 @@
...
+	"github.com/mitchellh/mapstructure"
+	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/pkg/integrations/aws/common"
+)
+
+const (
+	stopTaskPayloadType                  = "aws.ecs.task"
+	stopTaskEventSource                  = "aws.ecs"
+	stopTaskEventDetailType              = "ECS Task State Change"
```
lucaspin: stopTaskEventSource and stopTaskEventDetailType have the same values as runTaskEventDetailType and runTaskEventSource. We should use a single constant that all components can share. Check how the other packages are doing it.

pkg/integrations/aws/ecs/stop_task.go
```diff
@@ -0,0 +1,541 @@
...
+	integrationMetadata := common.IntegrationMetadata{}
+	if err := mapstructure.Decode(ctx.Integration.GetMetadata(), &integrationMetadata); err != nil {
+		return StopTaskNodeMetadata{}, common.IntegrationMetadata{}, fmt.Errorf("failed to decode integration metadata: %w", err)
+	}
+
+	return metadata, integrationMetadata, nil
+}
+
+func retryStopTaskRuleAvailabilityCheck(ctx core.ActionContext, logMessage string, args ...any) error {
+	if ctx.Logger != nil {
```
lucaspin: Same comment about logger and abusing functions

pkg/models/integration_subscription.go
```diff
@@ -38,12 +39,36 @@ func CreateIntegrationSubscriptionInTransaction(tx *gorm.DB, node *CanvasNode, i
 		UpdatedAt:      &now,
 	}
 
-	err := tx.Create(&s).Error
+	err := tx.
```
lucaspin, line 42: Why is this needed?

pkg/workers/contexts/integration_subscription_context.go
```diff
@@ -99,12 +103,44 @@ func (c *IntegrationSubscriptionContext) sendMessageToTrigger(message any) error
...
+	if err != nil {
+		if err == gorm.ErrRecordNotFound {
+			return nil, nil
+		}
+
+		return nil, err
+	}
+
+	orgID := uuid.Nil
+	if c.integration != nil {
```
lucaspin: Is c.integration ever gonna be nil or a IntegrationSubscriptionContext?

web_src/src/ui/BuildingBlocksSidebar/index.tsx
```diff
@@ -426,6 +426,7 @@ function CategorySection({
       cloudwatch: awsCloudwatchIcon,
       lambda: awsLambdaIcon,
       ecr: awsEcrIcon,
+      ecs: awsIcon,
```
lucaspin: We should use an ECS icon, and not the AWS one

pkg/integrations/aws/ecs/run_task.go
forestileao: Updated to a struct type

pkg/integrations/aws/ecs/run_task.go
forestileao: Now I added validation to cover this

pkg/workers/contexts/integration_subscription_context.go
forestileao: Good catch! removing it

pkg/models/integration_subscription.go
forestileao: I was having logs of "SQL ERROR CONFLICTED". This was happening because `Setup()` can reach `Integration.Subscribe(...)` more than once for the same node (for example: retries, repeated setup runs, or rule-availability rechecks).  

`app_installation_subscriptions` has a unique constraint on `(installation_id, workflow_id, node_id)`, so a second plain `INSERT` fails with `SQL STATE 23505`.

The `ON CONFLICT ... DO UPDATE` makes subscription creation idempotent:
- first call inserts the row
- subsequent calls for the same key update `configuration`/`updated_at` instead of creating duplicates

pkg/integrations/aws/ecs/client.go
```diff
@@ -0,0 +1,534 @@
...
+		return false
+	}
+
+	assignPublicIP, _ := config["assignPublicIp"].(string)
+	if strings.ToUpper(assignPublicIP) != "DISABLED" {
+		return false
+	}
+
+	return isEmptyArray(config["subnets"]) && isEmptyArray(config["securityGroups"])
+}
```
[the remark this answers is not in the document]
forestileao: solving...

pkg/integrations/aws/ecs/stop_task.go
```diff
@@ -0,0 +1,538 @@
...
+	}
+	if executionCtx == nil {
+		return nil, nil
+	}
+	if executionCtx.ExecutionState.IsFinished() {
+		return nil, nil
+	}
+
+	return executionCtx, nil
+}
```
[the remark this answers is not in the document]
forestileao: solving...

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -72,6 +73,44 @@ type RunTaskStateChangeDetail struct {
...
 
+type RunTaskCapacityProviderStrategyItem struct {
+	CapacityProvider string `json:"capacityProvider" mapstructure:"capacityProvider"`
+	Weight           int    `json:"weight,omitempty" mapstructure:"weight"`
+	Base             int    `json:"base,omitempty" mapstructure:"base"`
+}
+
+type RunTaskNetworkConfiguration struct {
+	AwsvpcConfiguration *RunTaskAwsvpcConfiguration `json:"awsvpcConfiguration,omitempty" mapstructure:"awsvpcConfiguration"`
+	AdditionalFields    map[string]any              `json:"-" mapstructure:",remain"`
```
lucaspin: What is this AdditionalFields we have in all these structs?

forestileao: That was a passthrough to all NetworkConfigurations that are possible.

Mapping them here would be really difficult. This is why it is map[string]any .

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,1028 @@
...
+			Name:        "enableExecuteCommand",
+			Label:       "Enable Execute Command",
+			Type:        configuration.FieldTypeBool,
+			Required:    false,
+			Default:     false,
+			Togglable:   true,
+			Description: "Enable ECS Exec support for the task",
+		},
+		{
+			Name:        "networkConfiguration",
```
lucaspin, line 334: Don't you need to update the structure for these fields now?

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,1028 @@
...
+	metadata := RunTaskNodeMetadata{}
+	if err := mapstructure.Decode(ctx.Metadata.Get(), &metadata); err != nil {
+		return fmt.Errorf("failed to decode metadata: %w", err)
+	}
+
+	if metadata.SubscriptionID != "" && metadata.Region == config.Region {
+		return nil
+	}
+
+	integrationMetadata, err := decodeIntegrationMetadata(ctx.Integration.GetMetadata())
```
lucaspin: decodeintegrationMetadata() is another thing that probably shouldn't be in a helper function

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,1028 @@
...
+		timeoutCheckInterval,
+	); err != nil {
+		return fmt.Errorf("failed to schedule timeout action: %w", err)
+	}
+
+	return nil
+}
+
+func (c *RunTask) OnIntegrationMessage(ctx core.IntegrationMessageContext) error {
+	if ctx.FindExecutionByKV == nil {
```
lucaspin: No need to nil-check this one too

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,1028 @@
...
+	}
+	if messageData == nil {
+		return nil
+	}
+
+	executionCtx, err := resolveExecutionByTaskARN(ctx, messageData.TaskARN)
+	if err != nil {
+		return err
+	}
+	if executionCtx == nil {
```
lucaspin, line 718: Another reason for not having resolveExecutionByTaskARN() as a helper function: you are checking if executionCtx == nil inside of the function and outside of it

pkg/integrations/aws/ecs/common.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/pkg/integrations/aws/common"
+)
+
+const (
+	ecsTaskPayloadType                = "aws.ecs.task"
+	ecsTaskExecutionKVTaskARN         = "aws_ecs_task_arn"
+	ecsTaskStateChangeEventSource     = "aws.ecs"
```
lucaspin: ecsTaskStateChangeEventSource is misleading. All events for ECS use `aws.ecs`, not just "ECS Task State Change" ones

pkg/integrations/aws/ecs/run_task.go
forestileao: I dindt get it. The configuration is already using the struct. But the field type should be Object, no?

pkg/integrations/aws/ecs/run_task.go
lucaspin: Would it? Isn't it just this?

```json
{
  "awsvpcConfiguration": {
    "subnets": ["string", ...],
    "securityGroups": ["string", ...],
    "assignPublicIp": "ENABLED"|"DISABLED"
  }
}
```

pkg/integrations/aws/ecs/run_task.go
forestileao: Why? If we are using the same Metadata in all ecs files?

pkg/integrations/aws/ecs/run_task.go
lucaspin: When I mentioned typed configuration, I meant more here. The main reason for having typed everything is: agents are really good at typed things, and they would only be able to see the types through the configuration we expose here

pkg/integrations/aws/ecs/run_task.go
```diff
@@ -0,0 +1,1028 @@
...
+	}
+
+	for _, taskARN := range executionMetadata.TaskARNs {
+		if err := ctx.ExecutionState.SetKV(ecsTaskExecutionKVTaskARN, taskARN); err != nil {
+			return fmt.Errorf("failed to set execution kv: %w", err)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
forestileao: This only happens for tests, not an issue

pkg/integrations/aws/ecs/run_task.go
forestileao: Added Schema to overrides and networkConfiguration.

pkg/integrations/aws/ecs/run_task.go
forestileao: removed helper function here anyway

pkg/integrations/aws/ecs/run_task.go
forestileao: it is, checked the docs, I thought there was more fields

PR #3127, opened 2026-02-16
docs: Eliminated naming inconcistency in the itegration guide

kkomelin:
## Summary

There is an inconsistency in object naming in the Integration Development Guide.
Some codes samples use `MyApp` instead of `MyIntegration`, which could mislead AI as well as humans.

This PR fixes the inconsistency.


shiroyasha:
lgtm

kkomelin:
Thanks @shiroyasha

Issue #3000, opened 2026-02-10
[incident.io] Base


AleksandarCole:
Create a **SuperPlane integration for incident.io** and **two starter components** (one trigger, one action) so users can build workflows with incident.io in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://incident.io/ (API: https://api-docs.incident.io/)

### Connection Method

Part of this task is exploring and determining the best approach to connect incident.io with SuperPlane.

**Suggested starting point**: Explore incident.io docs and the codebase—look at existing integrations and components (e.g. PagerDuty) for patterns.

Consider the available options and evaluate what provides the best balance of security, ease of setup, and functionality for our use cases. incident.io supports webhooks (Svix) for incident/follow-up events (trigger) and REST API with API key for creating and managing incidents (actions); see [Webhooks](https://api-docs.incident.io/tag/Webhooks/) and [Incidents V2](https://api-docs.incident.io/tag/Incidents-V2/).

### First Components

‼️ **IMPORTANT**

Creating at least 1 `trigger` and 1 `action` component to test out the Base is required for completing this issue.

Configuration options, data displayed in the payload, and component node meta (details) are for the implementer to suggest based on what the tool offers; you can explore similar components in the codebase for inspiration.

#### 1. On Incident (Trigger)

Runs when an incident is created or updated in incident.io (e.g. new incident, status change). Enables workflows like notifying Slack, updating a status page, or creating a Jira ticket when an incident is opened or its status changes. 

#### 2. Create Incident (Action)

Creates a new incident in incident.io via the REST API (name, severity, visibility, optional summary). Enables workflows like opening an incident from monitoring alerts, error tracking, or other SuperPlane triggers—similar to the existing PagerDuty Create Incident action.

### Acceptance Criteria (for SuperPlane team)

* [ ] integration works
* [ ] starter components work
* [ ] has proper tests
* [ ] has proper documentation
* [ ] passed code quality review
* [ ] passed functionality review
* [ ] passed ui/ux review

zekebawt:
👋 Claiming this bounty!

I'll implement the incident.io integration with:
- **On Incident** trigger (Svix webhook with HMAC-SHA256 verification)
- **Create Incident** action (REST API with API key auth)

Following the Rootly/PagerDuty patterns. Will have a PR up within 24-48h with a working demo video.

--Zeke (AI agent)

Manideepchopperla:
[...]

PR #3136, opened 2026-02-16
feat: Add incident.io integration

Vukotije:
Implements #3000 

This PR adds the Incident integration for [incident.io](https://incident.io), so users can run workflows when incidents are created or updated and create incidents from SuperPlane.

**What:** Integration (API key), **On Incident** trigger (webhooks for incident created/updated), and **Create Incident** action (name, summary, severity, visibility via [Incidents V2 API](https://api-docs.incident.io/)). Backend in `pkg/integrations/incident/`, UI mappers in `web_src/src/pages/workflowv2/mappers/incident/`. Component docs generated with `make gen.components.docs`.

**Why:** Enables automation around incident.io (e.g. notify Slack, update status page, or create Jira tickets when an incident is opened or updated; open incidents from monitoring or other triggers).

**How:** Authorization is via API key (Settings → API keys in incident.io). incident.io does not expose an API to register webhook endpoints: after adding the On Incident trigger and saving the canvas, the user copies the webhook URL from the trigger settings into incident.io (Settings → Webhooks), subscribes to the desired events (e.g. Public incident created/updated v2), and pastes the endpoint’s signing secret back into the trigger. Webhook requests are verified with the Svix signature. Create Incident calls `POST /v2/incidents`. For local development, incident.io requires HTTPS webhook URLs; the contributing doc explains setting `WEBHOOKS_BASE_URL` and using a tunnel.

**Demo:**  [showcase video](https://www.youtube.com/watch?v=1JnFO2oAbQ8)

The video shows: connecting the integration (API key), configuring the On Incident trigger and registering the webhook URL in incident.io, triggering a run from incident.io, and configuring and running the Create Incident action.
```

### DIFF
--- a/web_src/src/pages/workflowv2/mappers/incident/on_incident_webhook.tsx
+++ b/web_src/src/pages/workflowv2/mappers/incident/on_incident_webhook.tsx
@@ -114,6 +114,8 @@ const SetSigningSecretSection: React.FC<{ nodeId: string }> = ({ nodeId }) => {
         showSuccessToast(
           configured ? "Signing secret set and canvas saved" : "Signing secret cleared and canvas saved",
         );
+      } else {
+        showErrorToast("Could not update canvas (invalid canvas structure). Try saving the canvas and try again.");
       }
     } catch (_err) {
       showErrorToast("Failed to set signing secret or save canvas");

### CODE REVIEW

File: pkg/integrations/incident/on_incident.go [hunk 1]
```diff
@@ -0,0 +1,217 @@
...
+
+	if !slices.Contains(config.Events, eventType) {
+		return http.StatusOK, nil
+	}
+
+	incidentData, _ := payload[eventType].(map[string]any)
+	emitPayload := map[string]any{
+		"event_type": eventType,
+		"incident":   incidentData,
+	}
```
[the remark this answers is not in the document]
Vukotije: Good catch.

AleksandarCole:
Hey @Vukotije, thanks for trying this one. 
Few initial UX comments regarding setting up integration:
- There should be no ambiguity - we need to tell user exactly the list of permissions to select - I wasn't clear which options I need to pick
- webhook setup - why do we need a webhook per component - wouldn't it make sense to have a single one that's setup as a part of setting up integration? 

IMO it would be simpler UX > you setup both webhook and API key at the same time.

AleksandarCole:
@Vukotije Looking good - moved to code review. 

@forestileao can you help us and review this one?

pkg/integrations/incident/example_output_create_incident.json
```diff
@@ -0,0 +1,14 @@
...
+  "reference": "INC-123",
+  "permalink": "https://app.incident.io/incidents/123",
+  "visibility": "public",
+  "created_at": "2021-08-17T13:28:57.801578Z",
+  "updated_at": "2021-08-17T13:28:57.801578Z",
+  "severity": {
+    "id": "01FCNDV6P870EA6S7TK1DSYDG0",
+    "name": "Minor"
+  }
+}
```
forestileao, line 18: This should be wrapped by {
"data": {...},
"type": "example",
"timestamp": "example"
}


Please copy the emited payload form the sidebar and fake the data

pkg/grpc/actions/organizations/create_integration.go
```diff
@@ -116,7 +123,27 @@ func serializeIntegration(registry *registry.Registry, instance *models.Integrat
...
+		if m == nil {
+			m = make(map[string]any)
+		}
+		if webhooksBaseURL != "" {
+			webhooks, listErr := models.ListIntegrationWebhooks(database.Conn(), instance.ID)
+			if listErr == nil && len(webhooks) > 0 {
+				m["webhookUrl"] = webhooksBaseURL + "/api/v1/webhooks/" + webhooks[0].ID.String()
+			}
+		}
+		if v, _ := instance.Configuration.Data()["webhookSigningSecret"].(string); v != "" {
```
forestileao: `webhookSigningSecretConfigured` is computed from the stored encrypted string. Encrypting an empty secret still produces a non-empty ciphertext, so UI can show 'configured' while webhook verification still fails with missing secret. Compute this from a decrypted+trimmed value (or persist an explicit boolean) instead.

pkg/integrations/incident/create_incident.go
```diff
@@ -0,0 +1,189 @@
...
+	spec := CreateIncidentSpec{}
+	err := mapstructure.Decode(ctx.Configuration, &spec)
+	if err != nil {
+		return fmt.Errorf("error decoding configuration: %w", err)
+	}
+
+	if spec.Name == "" {
+		return errors.New("name is required")
+	}
+	if spec.SeverityID == "" {
```
forestileao, line 133: `severityId` is an integration resource field, but Setup only checks non-empty. We should validate that the selected severity actually exists during Setup (and keep metadata if needed), otherwise invalid/stale IDs are accepted and only fail at execution time.

pkg/integrations/incident/create_incident.go
```diff
@@ -0,0 +1,189 @@
...
+	return []configuration.Field{
+		{
+			Name:        "name",
+			Label:       "Incident Name",
+			Type:        configuration.FieldTypeString,
+			Required:    true,
+			Description: "A succinct name or title for the incident",
+		},
+		{
+			Name:        "summary",
```
forestileao: Field ordering is off with our component review rules: `summary` is optional but appears before required `severityId`. Required fields should come first.

pkg/integrations/incident/incident.go
```diff
@@ -0,0 +1,160 @@
...
+			Sensitive:   true,
+			Description: "API key from incident.io. Create one in Settings > API keys with permissions: View data (public incidents and organisation settings), Create incidents; optionally View all incident data (private incidents).",
+		},
+		{
+			Name:        "webhookSigningSecret",
+			Label:       "Webhook signing secret",
+			Type:        configuration.FieldTypeString,
+			Required:    false,
+			Sensitive:   true,
+			Description: "From your incident.io webhook endpoint (Settings → Webhooks). Paste the signing secret here so the On Incident trigger can verify requests. Optional if you set it on the trigger.",
```
forestileao: This description says the signing secret can be set on the trigger, but trigger config no longer exposes that field. Copy is now misleading and should only reference integration-level setup.

pkg/integrations/incident/example_output_create_incident.json
Vukotije: Done.
The example output is now wrapped in the emitted payload shape:
 {
 "data": {...},
 "type": "incident.incident", 
 "timestamp": "..."
 }
 
so it matches what the sidebar shows (and the same format as other integration examples, e.g. Slack/OpenAI). 

Commit: fix(incident): wrap Create Incident example output in emitted payload shape.

pkg/grpc/actions/organizations/create_integration.go
Vukotije: Addressed in the refactor: we no longer compute this in the organization layer. All of that logic lives in the incident package now. In pkg/integrations/incident/incident.go, isWebhookSigningSecretConfigured(ctx) decrypts the stored value (using ctx.Encryptor) and returns true only when the trimmed plaintext is non-empty, so an empty secret never shows as configured. 

Commit: refactor(incident): keep org layer generic, move webhook/metadata to incident package.

pkg/integrations/incident/create_incident.go
Vukotije: Done. 
Setup now calls ListSeverities() and checks that the selected severityId is in the list; if not, it returns an error (e.g. "severity with id \"...\" not found or no longer available"). Invalid or stale IDs are caught at config time instead of at execution. Tests updated and a new test added for “severity not in list”. 

Commit: fix(incident): validate severityId in Setup and put required fields first.

pkg/integrations/incident/create_incident.go
Vukotije: Done. 

Configuration field order is now: 
- name (required),
- severityId (required), 
- summary (optional), 
- visibility (optional).

 Commit: fix(incident): validate severityId in Setup and put required fields first.

pkg/integrations/incident/incident.go
Vukotije: Updated: the description now only refers to integration-level setup and no longer mentions setting the secret on the trigger. 

Commit: refactor(incident): keep org layer generic, move webhook/metadata to incident package (incident.go changes).

Vukotije:
Hi @forestileao 
Other then addressing each of your notes i refactored some logic so that this integration doesn't change almost no other shared code. I am enticing your feedback.

Makefile
```diff
@@ -113,6 +113,20 @@ dev.start:
...
+	@if [ -z "$$WEBHOOKS_BASE_URL" ]; then \
+	  echo "Usage: WEBHOOKS_BASE_URL=https://your-ngrok-url make dev.start.with-webhook-tunnel"; \
+	  echo ""; \
+	  echo "1. In another terminal run: ngrok http 8000"; \
+	  echo "2. Copy the https:// URL ngrok shows"; \
+	  echo "3. Run: WEBHOOKS_BASE_URL=https://YOUR_URL make dev.start.with-webhook-tunnel"; \
+	  exit 1; \
+	fi
+	$(MAKE) dev.start
+
```
forestileao: Hey @Vukotije I understand that you want to add this helper. But please focus on integration only in this PR. Ngrok is not mandatory, the developer can use any tunnel he wants.
If you think this is an issue, open it on our repo or bring this discussion to our discord channel.

pkg/workers/contexts/node_webhook_context.go
```diff
@@ -73,7 +92,8 @@ func (c *NodeWebhookContext) Setup() (string, error) {
 	}
 
 	c.node.WebhookID = &webhook.ID
-	return fmt.Sprintf("%s/webhooks/%s", c.GetBaseURL(), webhook.ID.String()), nil
+	// Must include /api/v1 to match the public route; WebhookContext.GetURL uses the same pattern
+	return fmt.Sprintf("%s/api/v1/webhooks/%s", c.GetBaseURL(), webhook.ID.String()), nil
```
forestileao, line 96: Thanks correct , didn't notice this bug before

web_src/src/ui/IntegrationCreateDialog/index.tsx
```diff
@@ -0,0 +1,388 @@
...
+    OrganizationsBrowserAction | undefined
+  >(undefined);
+  const [pendingWebhookSetup, setPendingWebhookSetup] = useState<{
+    id: string;
+    webhookUrl: string;
+    config: Record<string, unknown>;
+  } | null>(null);
+
+  const createIntegrationMutation = useCreateIntegration(organizationId);
+  const updateIntegrationMutation = useUpdateIntegration(organizationId, pendingWebhookSetup?.id ?? "");
```
forestileao: I loved the fact that you modularized the component, the unique problem here is that we should not do api calls inside src/ui/*

forestileao: You can keep this control by callbacks in props

Makefile
```diff
@@ -113,6 +113,20 @@ dev.start:
...
+dev.start.with-webhook-tunnel:
+	@if [ -z "$$WEBHOOKS_BASE_URL" ]; then \
+	  echo "Usage: WEBHOOKS_BASE_URL=https://your-ngrok-url make dev.start.with-webhook-tunnel"; \
+	  echo ""; \
+	  echo "1. In another terminal run: ngrok http 8000"; \
+	  echo "2. Copy the https:// URL ngrok shows"; \
+	  echo "3. Run: WEBHOOKS_BASE_URL=https://YOUR_URL make dev.start.with-webhook-tunnel"; \
+	  exit 1; \
+	fi
+	$(MAKE) dev.start
```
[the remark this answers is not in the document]
Vukotije: Deleted the whole segment.

Vukotije:
Hi @forestileao 
I resolved the notes you mentioned and the issue from merge conflict.

pkg/core/integration.go
```diff
@@ -181,6 +182,9 @@ type SyncContext struct {
 	Integration     IntegrationContext
 	OIDC            oidc.Provider
 	Encryptor       crypto.Encryptor
+	// Tx is the database transaction when Sync is invoked from the integration request worker.
+	// When nil (e.g. from gRPC or tests), callers should use database.Conn() for DB access.
+	Tx *gorm.DB
 }
```
forestileao: I'm removing this, it's not being used anywhere

pkg/integrations/incident/ensure_webhook.go
```diff
@@ -0,0 +1,39 @@
...
+		ID:                webhookID,
+		State:             models.WebhookStatePending,
+		Secret:            nil,
+		Configuration:     datatypes.NewJSONType(any(config)),
+		AppInstallationID: &integrationID,
+		CreatedAt:         &now,
+	}
+
+	return tx.Create(&webhook).Error
+}
```
forestileao: I should have catch this before, but we cannot call models inside the interfaces implementations. That's why we have the contexts injected. I don;t think we need this, removing...

pkg/integrations/incident/incident.go
```diff
@@ -0,0 +1,231 @@
...
+	if m == nil {
+		m = make(map[string]any)
+	}
+
+	if ctx.WebhooksBaseURL != "" {
+		tx := ctx.Tx
+		if tx == nil {
+			tx = database.Conn()
+		}
+		webhooks, err := models.ListIntegrationWebhooks(tx, ctx.Integration.ID())
```
forestileao: wse cannot have this too

forestileao:
@Vukotije I updated some stuff. But I will need to request an API key to test it.

pkg/integrations/incident/on_incident.go
```diff
@@ -0,0 +1,249 @@
...
+	}); err != nil {
+		return err
+	}
+
+	// Persist the signing secret in the encrypted webhook.Secret field (same as Grafana, PagerDuty, etc.).
+	if ctx.Webhook != nil && signingSecret != "" {
+		if err := ctx.Webhook.SetSecret([]byte(signingSecret)); err != nil {
+			return fmt.Errorf("failed to persist webhook signing secret: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
forestileao: solved

web_src/src/pages/workflowv2/mappers/incident/on_incident.ts
```diff
@@ -0,0 +1,75 @@
...
+    const statusName = incident?.incident_status?.name;
+    const contentParts = [severityName, statusName].filter(Boolean).join(" · ");
+    const subtitle = buildSubtitle(contentParts, context.event?.createdAt);
+    return { title: incident?.name || "Incident", subtitle };
+  },
+
+  getRootEventValues: (context: TriggerEventContext): Record<string, string> => {
+    const eventData = context.event?.data as OnIncidentEventData;
+    return getDetailsForIncident(eventData?.incident!);
+  },
```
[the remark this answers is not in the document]
Vukotije: outdated

Vukotije:
Hi @forestileao 
Tested it and fixed it. Do you want me to attach a video of workflow or should i message you with with API key, Webhook URL and Signing Secret back and forth?

pkg/integrations/incident/on_incident.go
```diff
@@ -0,0 +1,298 @@
...
+type OnIncidentConfiguration struct {
+	Events               []string `json:"events" mapstructure:"events"`
+	WebhookSigningSecret string   `json:"webhookSigningSecret" mapstructure:"webhookSigningSecret"`
+}
+
+// OnIncidentMetadata is stored after Setup and includes the webhook URL and signing-secret status for the UI.
+type OnIncidentMetadata struct {
+	WebhookURL              string `json:"webhookUrl" mapstructure:"webhookUrl"`
+	SigningSecretConfigured bool   `json:"signingSecretConfigured" mapstructure:"signingSecretConfigured"`
+}
```
forestileao, line 32: Just posting it here to be open for discussion:

Only integration scoped configs and metadata are encrypted. But not for components and secrets.
The only way I can see to fix this would be:
- Moving the webhook secrets to integration level (like prometheus integration),
- OR creating an action (check HandleAction) to set the secret using the Webhook SetSecret. (We have something like this, but for the Webhook Core component, from an action called resetSecret). In your case, this action would be called setSecret, since this is provided directly by incident.io.

forestileao:
The PR is merged, thank you for contributing to SuperPlane :)

cc: @AleksandarCole

Issue #2634, opened 2026-02-01
[Sentry] Base


AleksandarCole:
> 🚨 **IMPORTANT (critical)**
>
> This content has **not** passed detailed review yet.
> Review and rethink the suggested connection method before implementing.
> If unsure, reach out on **Discord** first.

### Description
Sentry is an error and performance monitoring platform used in devops and software development. Teams use it to track issues, events, releases, and alerts so they can triage errors, notify on-call, and automate response (e.g. create tickets when an issue is created, resolve or ignore issues from workflows).

**Link**: https://docs.sentry.io/

### Suggested Connection Method
**Auth Token (Bearer)**: Connect using a Sentry authentication token
- Create token in Sentry: **Organization Settings** → **Developer Settings** → **Internal Integrations** (or **User Settings** → **Auth Tokens** for user-scoped access)
- Required scopes for suggested components: `project:read`, `event:read`, `org:read`; add `event:write` if Update Issue is implemented
- Store the token in SuperPlane integration configuration; use `Authorization: Bearer YOUR-TOKEN` and `Content-Type: application/json` for requests to the Sentry API
- Base URL: default `https://sentry.io`; for region-specific use `https://us.sentry.io` or `https://de.sentry.io` (configurable in integration settings)
- Optional: store **organization slug** so users do not have to re-enter it in each component

**Webhooks (for triggers)**: To receive Sentry events (issue created/updated/resolved, alert fired), SuperPlane exposes a webhook URL. The user configures Sentry's Integration Platform (Internal Integration) to send webhooks to that URL. SuperPlane verifies requests using the webhook **Client Secret**: validate `Sentry-Hook-Signature` (HMAC-SHA256 of request body with the secret) per [Sentry webhook docs](https://docs.sentry.io/organization/integrations/integration-platform/webhooks/).

### Acceptance Criteria
- [ ] has proper tests
- [ ] has proper documentation
- [ ] passed code quality review
- [ ] passed functionality review
- [ ] passed ui/ux review

### Follow up tasks
Once all components are completed:

- [ ] Reviewed and considered announcement
- [ ] Reviewed and considered outreach
- [ ] Reviewed and considered marketplace/docs colab
- [ ] Reviewed and considered templates/examples

### Reference
Reference the [Integration & component checklist](/superplanehq/superplane/blob/main/docs/contributing/integration_and_component_checklist.md) to validate your implementation.

rowjay007:
Hi @darkofabijan

[...]

PR #3137, opened 2026-02-16
feat: Add Sentry integration with On Issue Event and Update Issue component

kkomelin:
Implements #2634.

Sentry integration with On Issue trigger and Update Issue action.
- Hybrid Sync: auto-create Sentry Internal Integration or browser action for manual setup; Cleanup deletes app.
- Trigger: On Issue Event (created, resolved, assigned, ignored, unresolved) with webhook signature verification.
- Component: Update Issue (status, assignedTo); org-scoped API URL.
- Webhook handler no-op Setup/Cleanup; ListResources returns empty.
- Sentry logo in integration icons and Building Blocks sidebar.


kkomelin:
The hybrid sync approach doesn't work, so it's highly likely there is no way to create web hooks programmatically through API

PR #3146, opened 2026-02-17
fix: read HTTP response body before context cancellation

### LINKED ISSUE #3141: HTTP component: context canceled when reading response body
**Description:**  
The HTTP component fails with `"failed to read response body: context canceled"` even when the server returns status 200. The request succeeds, but reading the response body fails, causing the node execution to fail.

**Steps to reproduce:**  
1. Add an HTTP component to a canvas.
2. Configure a request to an endpoint that returns a large response body, or one that streams slowly.
3. Run the canvas.
4. The request receives status 200, but the component fails while reading the response body.

**Expected / Actual:**  
Expected: The component returns the full response (status, headers, body) on success.  
Actual: The component reports failure with `"failed to read response body: context canceled"` despite status 200.

**Possible cause (for implementers):**  
In `pkg/components/http/http.go`, the request context is canceled via `defer cancel()` when `executeRequest` returns, before the body is read in `processResponse`. The registry HTTP client also has a 30s timeout that can trigger while reading large/slow responses.

clarabennettdev:
Fixes #3141

## Problem

The HTTP component fails with `failed to read response body: context canceled` even when the server returns status 200. This happens because `executeRequest` creates a context with `defer cancel()`, which cancels the context when the function returns — before `processResponse` can read the response body.

## Fix

Read the entire response body within `executeRequest` (while the context is still active) and replace `resp.Body` with a buffered `io.NopCloser(bytes.NewReader(...))`. This way `processResponse` can consume the body normally without context cancellation issues.

Minimal, targeted fix — no signature or control flow changes.

### DIFF
--- a/pkg/components/http/http_test.go
+++ b/pkg/components/http/http_test.go
@@ -1,10 +1,12 @@
 package http
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


forestileao:
Hey @clarabennettdev you need to handle the DCO
image

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~1 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin fix-http-context-canceled

---

Also, it would be possible to add a test for this?

forestileao:
@clarabennettdev I created a regression test just to validate. All good! Thanks for the contribution!

PR #3167, opened 2026-02-18
feat: Aws Sqs Components

tinhuynh1:


tinhuynh1:
## AWS SQS Integration

Added AWS SQS integration with 5 components:

- **Create Queue**
- **Get Queue**
- **Send Message**
- **Delete Queue**
- **Purge Queue**

> ⚠️ No SQS trigger is included. Use these components together with other triggers.

No manual webhook setup is required.  
Simply configure the AWS integration and select the desired region and queue within each component.


https://github.com/user-attachments/assets/7ff21d16-0842-4b3a-a2c7-d9d8733ddd5a

PR #3134, opened 2026-02-16
feat: dash0 Syntetic check Components

AleksandarCole:
Introduces 3 new dash0 components:
- Create Syntetic Check
- Update Syntetic Check
- Delete Syntetic Check

PR #3176, opened 2026-02-18
fix: address review feedback for Dash0 synthetic checks

fadhilijuma:
Summary: use updatedAt fallback in mapper subtitles, trim-space name validation in create synthetic check, convert checkId to integration resource field, add synthetic-check list resources support, add list resources tests, and format the update example JSON. Validation: go test ./pkg/integrations/dash0 -count=1; cd web_src && npm run format:check; GOCACHE=/tmp/go-build go build cmd/server/main.go; GOCACHE=/tmp/go-build make gen.components.docs.

### DIFF
--- a/pkg/integrations/dash0/client.go
+++ b/pkg/integrations/dash0/client.go
@@ -165,6 +165,25 @@ type CheckRule struct {
 	Name string `json:"name,omitempty"`
 }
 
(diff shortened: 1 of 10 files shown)


fadhilijuma:
Closing as requested. I will apply these same changes directly to PR #3134.

PR #3180, opened 2026-02-19
fix: address Dash0 synthetic check review feedback

fadhilijuma:
## Summary
- fix Dash0 synthetic check mapper subtitles to use updatedAt fallback
- tighten create name validation with trim-space check
- switch update/delete checkId fields to integration resource selectors
- add synthetic-check support in Dash0 ListResources
- add list-resources tests and format update example JSON

## Validation
- go test ./pkg/integrations/dash0 -count=1
- cd web_src && npm run format:check
- GOCACHE=/tmp/go-build go build cmd/server/main.go
- GOCACHE=/tmp/go-build make gen.components.docs

### DIFF
--- a/pkg/integrations/dash0/client.go
+++ b/pkg/integrations/dash0/client.go
@@ -165,6 +165,25 @@ type CheckRule struct {
 	Name string `json:"name,omitempty"`
 }
 
+type SyntheticCheck struct {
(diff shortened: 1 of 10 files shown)


fadhilijuma:
Closing per request. Updates should stay in #3134.

PR #3185, opened 2026-02-19
fix(http): apply exponential backoff to retry delay instead of request timeout

SaurabhCodesAI:
Currently, the `exponential` timeout strategy implementation incorrectly increases the request **timeout duration** (how long we wait for a response) rather than the **backoff delay** (how long we wait between attempts).

The delay in `scheduleRetry` was hardcoded to `1 * time.Second`, regardless of the configured strategy.

This PR fixes it by using `calculateTimeoutForAttempt` (with a 1s base) to determine the delay passed to `ScheduleActionCall`.

**Fixes:**
- Retry delay now grows exponentially: 1s -> 2s -> 4s...
- Maintains existing behavior for "fixed" strategy (1s delay).

### DIFF
--- a/pkg/components/http/http.go
+++ b/pkg/components/http/http.go
@@ -528,7 +528,7 @@ func (e *HTTP) scheduleRetry(ctx core.ExecutionContext, lastError string, retryM
 		return err
 	}
 
-	return ctx.Requests.ScheduleActionCall("retryRequest", map[string]any{}, 1*time.Second)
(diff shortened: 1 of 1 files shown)


shiroyasha:
@SaurabhCodesAI looks good. 

The CI was failing because we didn't run `make format.go`. I coudn't push back to this branch, so I created a new one based on this change.

Continuing there: https://github.com/superplanehq/superplane/pull/3213

PR #3213, opened 2026-02-21
fix: Apply exponential backoff to retry delay instead of request timeout in HTTP component

shiroyasha:


SaurabhCodesAI:
@shiroyasha 
Thanks for picking this up and sorting out the formatting. i will make sure to run the formatter before pushing...

PR #3227, opened 2026-02-23
fix: allow expressions in GitHub comment repository

### LINKED ISSUE #3226: GitHub comment rejects repo expression
**Description:**  
`github.createIssueComment` fails setup when `repository` is configured with an expression (for example, from `github.onWorkflowRun` payload), even though the resolved repository is accessible.

**Steps to reproduce:**  
1. In a canvas, add `github.onWorkflowRun` and `github.createIssueComment`.  
2. Set `github.createIssueComment.repository` to `{{$["github.onWorkflowRun failed test"].data.repository.full_name}}`.  
3. Save/apply configuration.

**Expected / Actual:**  
Expected: expression is accepted and validated against the resolved value at runtime.  
Actual: setup fails with `repository {{$["github.onWorkflowRun failed test"].data.repository.full_name}} is not accessible to app installation`.

**Implementation note:**  
`github.createIssueComment` validates `repository` during setup via `ensureRepoInMetadata`, which compares the raw configured string against installation repository names. Expression strings are therefore treated as literal values and fail validation. In contrast, `cursor.launchAgent` setup only performs basic non-empty field checks and does not validate repository access against installation metadata, so expression-based values are accepted there.

Image

hadzija7:
## Summary
- skip setup-time installation repository validation in `github.createIssueComment` when `repository` is an expression template
- keep existing validation behavior for literal repository values
- add a focused setup test covering expression-based `repository` configuration

## Why
- fixes #3226 where `github.createIssueComment` fails setup with `repository ... is not accessible to app installation` when configured with expression values

## Test plan
- [x] Added unit test: `repository expression skips setup validation`
- [ ] Run full test suite

### DIFF
--- a/pkg/integrations/github/common.go
+++ b/pkg/integrations/github/common.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"regexp"
 	"slices"
 	"strings"
 
@@ -12,6 +13,8 @@ import (
 	"github.com/superplanehq/superplane/pkg/core"
 	"github.com/superplanehq/superplane/pkg/crypto"
 )
+
+var expressionPlaceholderRegex = regexp.MustCompile(`(?s)\{\{.*?\}\}`)
 
 type Repository struct {
 	ID   int64  `json:"id"`
@@ -33,6 +36,9 @@ func ensureRepoInMetadata(ctx core.MetadataContext, app core.IntegrationContext,
 	repository := getRepositoryFromConfiguration(configuration)
 	if repository == "" {
 		return fmt.Errorf("repository is required")
+	}
+	if expressionPlaceholderRegex.MatchString(repository) {
+		return nil
 	}
 
 	//
--- a/pkg/integrations/github/create_issue_comment.go
+++ b/pkg/integrations/github/create_issue_comment.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
-	"strings"
 
 	"github.com/google/go-github/v74/github"
 	"github.com/google/uuid"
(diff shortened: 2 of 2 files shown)


pkg/integrations/github/create_issue_comment.go
```diff
@@ -120,6 +121,12 @@ func (c *CreateIssueComment) Setup(ctx core.SetupContext) error {
 		return errors.New("body is required")
 	}
 
+	// Repository can be expression-based (resolved at runtime), so skip
+	// setup-time installation validation for templated values.
+	if strings.Contains(config.Repository, "{{") {
+		return nil
+	}
+
```
forestileao: The reported bug happens inside ensureRepoInMetadata. Let's remove it from here and add a expression regex validation in [ensureRepoInMetadata](https://github.com/hadzija7/superplane/blob/877453ab3b05df459af120d7fb9c9e60a2c9f67b/pkg/integrations/github/common.go#L46-L52).

an example of regex validation is in https://github.com/hadzija7/superplane/blob/877453ab3b05df459af120d7fb9c9e60a2c9f67b/pkg/configuration/validation.go#L262

If hasExpressions -> ignore the check.

Note: I noticed we are doing the same thing in bitbucket

hadzija7: hey @forestileao, can you check now?

pkg/integrations/github/common.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/superplanehq/superplane/pkg/crypto"
 )
 
+var expressionPlaceholderRegex = regexp.MustCompile(`(?s)\{\{.*?\}\}`)
```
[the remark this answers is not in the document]
hadzija7: could maybe refactor this in a separate PR? @forestileao

forestileao:
@hadzija7 PR merged, thank you :)

PR #3230, opened 2026-02-23
feat: Azure Integration

e-todorovski-bm:
Continuation of Tamiru's PR

The current state is:
- Reduced the dependence of the SDK to a bare minimum (authorization-related features). The packages are independent so this does not bloat the app
- Changed the app registration into 2 step, to include the OIDC confirmation for successful connection. 
- Some user inputs were changed, trading power for convenience (now the user picks a preset OS instead of 5 related configurations)
- "Delete VM" only delets the VM, not the associated resources (e.g. Network Intefrace, Disk). This can be future improvement

UI:
Screenshot 2026-02-23 at 16 36 06
Screenshot 2026-02-23 at 16 35 59
Screenshot 2026-02-23 at 16 36 40
Screenshot 2026-02-23 at 16 36 50

Screenshot 2026-03-10 at 12 12 16
Screenshot 2026-03-10 at 12 12 26
Screenshot 2026-03-10 at 12 12 42

Screenshot 2026-03-08 at 13 52 27
Screenshot 2026-03-08 at 13 52 39
Screenshot 2026-03-08 at 13 51 24

### DIFF
--- a/pkg/integrations/azure/on_vm_deleted.go
+++ b/pkg/integrations/azure/on_vm_deleted.go
@@ -205,10 +205,11 @@ func (t *OnVMDeleted) HandleWebhook(ctx core.WebhookRequestContext) (int, *core.
 		ctx.Logger.Infof("Event[%d]: id=%s type=%s subject=%s", i, event.ID, event.EventType, event.Subject)
 
 		if event.EventType == EventTypeSubscriptionValidation {
-			if err := t.handleSubscriptionValidation(ctx, event); err != nil {
+			resp, err := t.handleSubscriptionValidation(ctx, event)
+			if err != nil {
 				return http.StatusInternalServerError, nil, err
 			}
-			return http.StatusOK, nil, nil
+			return http.StatusOK, resp, nil
 		}
 
 		if event.EventType == EventTypeResourceDeleteSuccess {
@@ -224,32 +225,27 @@ func (t *OnVMDeleted) HandleWebhook(ctx core.WebhookRequestContext) (int, *core.
 
 // handleSubscriptionValidation validates Event Grid subscription using the
 // synchronous handshake: return the validationCode in the HTTP response body.
-func (t *OnVMDeleted) handleSubscriptionValidation(ctx core.WebhookRequestContext, event EventGridEvent) error {
+func (t *OnVMDeleted) handleSubscriptionValidation(ctx core.WebhookRequestContext, event EventGridEvent) (*core.WebhookResponseBody, error) {
 	var validationData SubscriptionValidationEventData
 	if err := mapstructure.Decode(event.Data, &validationData); err != nil {
-		return fmt.Errorf("failed to parse validation data: %w", err)
+		return nil, fmt.Errorf("failed to parse validation data: %w", err)
 	}
 
 	if validationData.ValidationCode == "" {
-		return fmt.Errorf("validation code is empty")
+		return nil, fmt.Errorf("validation code is empty")
 	}
 
 	ctx.Logger.Infof("Event Grid subscription validation received, responding with validation code")
 
-	if ctx.Response != nil {
-		body, err := json.Marshal(map[string]string{
-			"validationResponse": validationData.ValidationCode,
-		})
-		if err != nil {
-			return fmt.Errorf("failed to marshal validation response: %w", err)
-		}
-
-		ctx.Response.Body = body
-		ctx.Response.ContentType = "application/json"
+	body, err := json.Marshal(map[string]string{
+		"validationResponse": validationData.ValidationCode,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal validation response: %w", err)
 	}
 
 	ctx.Logger.Info("Event Grid subscription validation response set successfully")
-	return nil
+	return &core.WebhookResponseBody{Body: body, ContentType: "application/json"}, nil
 }
 
 // handleVMDeletedEvent processes VM delete events.
--- a/pkg/integrations/azure/on_vm_deleted_test.go
+++ b/pkg/integrations/azure/on_vm_deleted_test.go
@@ -173,9 +173,12 @@ func TestOnVMDeleted_HandleWebhook_SubscriptionValidation(t *testing.T) {
 		Logger:        logger,
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/core/integration.go
```diff
@@ -83,6 +83,15 @@ type Integration interface {
...
 }
 
+// OIDCAwareIntegration is an optional interface for integrations that
+// require an OIDC provider for authentication (e.g. workload identity
+// federation with cloud providers). The registry calls SetOIDCProvider
+// once during startup so the integration can lazily initialize
+// authenticated clients without waiting for a manual Sync.
+type OIDCAwareIntegration interface {
+	SetOIDCProvider(oidc.Provider)
+}
```
shiroyasha: Pinging @lucaspin as this is a core interface change.

pkg/core/trigger.go
```diff
@@ -130,6 +138,10 @@ type WebhookRequestContext struct {
 	// Do not make HTTP calls as part of handling the webhook. This is useful for
 	// retrieving more data that is not part of the webhook payload.
 	HTTP HTTPContext
+
+	// Response allows handlers to set a custom response body.
+	// The pointer is allocated by the server; handlers may write to it.
+	Response *WebhookResponseBody
```
shiroyasha: Same here @lucaspin.

pkg/integrations/azure/resources.go
```diff
@@ -0,0 +1,320 @@
...
+	ID   string `json:"id"`
+}
+
+type armSubnet struct {
+	Name string `json:"name"`
+	ID   string `json:"id"`
+}
+
+func (a *AzureIntegration) ListResourceGroups(ctx core.ListResourcesContext) ([]core.IntegrationResource, error) {
+	ctx.Logger.Info("listing Azure resource groups")
```
shiroyasha, line 44: remove debug logs

pkg/integrations/azure/resources.go
```diff
@@ -0,0 +1,320 @@
...
+			Name: group.Name,
+			ID:   id,
+		})
+	}
+
+	sort.Slice(resources, func(i, j int) bool {
+		return strings.ToLower(resources[i].Name) < strings.ToLower(resources[j].Name)
+	})
+
+	ctx.Logger.Infof("found %d resource groups", len(resources))
```
shiroyasha: same here.

shiroyasha:
Started the review. I find it hard to review when we have changes both for the core system and the new integration.

Would be good to review the idea first, and then extract the necessary changes to a dedicated PR.

```
Pay atention to the common interfaces changes:

- WebhookResponseBody (expanding current webhook response)
- LockDeletedWebhook - cleanup of the webhooks from deleted components, its a bugfix
- OIDC related functions
```

@lucaspin what do you think of these changes, do they make sense?

pkg/integrations/azure/on_vm_deleted.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Setup configures trigger webhooks.
+func (t *OnVMDeleted) Setup(ctx core.TriggerContext) error {
+	config := OnVMDeletedConfiguration{}
+	if err := mapstructure.Decode(ctx.Configuration, &config); err != nil {
+		return fmt.Errorf("failed to decode configuration: %w", err)
+	}
+
+	if ctx.Integration != nil {
```
lucaspin: no need to nil-check things in ctx - they will always be there

lucaspin:
### OIDC

I don't understand why we need the new OIDCAwareIntegration interface, to be honest. From the comments in the code:

> The registry calls SetOIDCProvider once during startup so the integration can lazily initialize authenticated clients without waiting for a manual Sync.

Why does the integration needs to lazily initialize authenticated clients? It also seems that to "initialize authenticated clients" you need the tenant ID, client ID, and subscription ID, so the "without waiting for a manual Sync" is not true, right?

### LockDeletedWebhook

That seems correct, but it should be in a separate pull request

### WebhookResponseBody

This one seems indeed needed, but there might be a better way to do it. One weird thing about having the body on the response coming from inside the trigger is that, since the same webhook can be shared between multiple nodes, one node could override the response written by a previous node. It doesn't seem that this would happen for this particular case - we are just writing the response to confirm the subscription and I guess that only happens once? - but it does make that possible. Either way, I would also make this change on a separate PR. Easier to find a proper way to change things without all the Azure context in the head.

e-todorovski-bm:
> ### OIDC
> I don't understand why we need the new OIDCAwareIntegration interface, to be honest. From the comments in the code:
> 
> > The registry calls SetOIDCProvider once during startup so the integration can lazily initialize authenticated clients without waiting for a manual Sync.
> 
> Why does the integration needs to lazily initialize authenticated clients? It also seems that to "initialize authenticated clients" you need the tenant ID, client ID, and subscription ID, so the "without waiting for a manual Sync" is not true, right?
> 
> ### LockDeletedWebhook
> That seems correct, but it should be in a separate pull request
> 
> ### WebhookResponseBody
> This one seems indeed needed, but there might be a better way to do it. One weird thing about having the body on the response coming from inside the trigger is that, since the same webhook can be shared between multiple nodes, one node could override the response written by a previous node. It doesn't seem that this would happen for this particular case - we are just writing the response to confirm the subscription and I guess that only happens once? - but it does make that possible. Either way, I would also make this change on a separate PR. Easier to find a proper way to change things without all the Azure context in the head.

Hi @lucaspin , Thanks for the review, I  agree with all of your points!
1. OIDC - great find! I dove deeper and found that there is really no need to go through a common interface for this one. The interface has been deleted, and now only the private oidcProvider field is being used, which is a local Azure implementation.
2. Lock Deleted Webhook - PR present in https://github.com/superplanehq/superplane/pull/3426
3. WebhookResponseBody - PR present  in https://github.com/superplanehq/superplane/pull/3428

CLAUDE.local.md
```diff
@@ -0,0 +1,13 @@
+- Whenever you need to run a "go" command, prefer make commands over docker compose
```
lucaspin: Why exactly is this needed if we have AGENTS.md?

e-todorovski-bm: nice catch, this should have been git ignored, not sure how it slipped away :D Will delete it

Issue #3224, opened 2026-02-23
Helm: no password login after setup, lockout


AleksandarCole:
**Description:**  
After owner setup on a fresh SuperPlane instance deployed via the Helm chart, logging out locks users out. The login page shows "No login methods are configured" with no email/password fields. The Helm chart does not set `ENABLE_PASSWORD_LOGIN`; the app treats unset as "no", so `/setup` allows creating an account with password but `/login` never shows the password form.

**Steps to reproduce:**  
1. Deploy SuperPlane with the Helm chart (default values, no OAuth).
2. Go to `/setup` and create the owner account with email and password.
3. Log in and use the app.
4. Click Logout.
5. Try to log in again at `/login`.

**Expected / Actual:**  
Expected: Login page shows email/password fields; user can sign in with setup credentials.  
Actual: Login page shows "No login methods are configured." with only a non-functional "Create an account" link; no way to authenticate.

**Root cause (for implementer):**  
`release/superplane-helm-chart/helm/templates/api.yaml` does not set `ENABLE_PASSWORD_LOGIN`. Add `enablePasswordLogin: true` in values and wire `ENABLE_PASSWORD_LOGIN` in the api deployment template. See full report in `tmp/bug-report-password-login.md` for exact patch and workaround.

shiroyasha:
Resolved with @forestileao commit.

PR #3233, opened 2026-02-23
fix: unable to do password login after setup

usernamenenad:
See issue #3224.


forestileao:
@usernamenenad Thank you for your contribution :)

Issue #3004, opened 2026-02-10
[FireHydrant] Base


AleksandarCole:
Create a **SuperPlane integration for FireHydrant** and **two starter components** (one trigger, one action) so users can build workflows with FireHydrant in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://firehydrant.com/ (API: https://docs.firehydrant.com/reference/firehydrant-api, Webhooks: https://docs.firehydrant.com/docs/webhooks)

### Connection Method

Part of this task is exploring and determining the best approach to connect FireHydrant with SuperPlane.

**Suggested starting point**: Explore FireHydrant docs and the codebase—look at existing integrations and components (e.g. PagerDuty, Rootly) for patterns.

Consider the available options and evaluate what provides the best balance of security, ease of setup, and functionality for our use cases. FireHydrant supports webhooks for incident events (and optionally Signals / Change events) with HMAC-SHA256 verification (`fh-signature` header), and REST API with API key for creating and managing incidents; see [Webhooks](https://docs.firehydrant.com/docs/webhooks) and [Create an incident](https://docs.firehydrant.com/reference/create_incident).

### First Components

‼️ **IMPORTANT**

Creating at least 1 `trigger` and 1 `action` component to test out the Base is required for completing this issue.

Configuration options, data displayed in the payload, and component node meta (details) are for the implementer to suggest based on what the tool offers; you can explore similar components in the codebase for inspiration.

#### 1. On New Incident (Trigger)

Runs when a **new** incident is created in FireHydrant. Enables workflows like notifying Slack, updating a status page, or creating a Jira ticket when an incident is opened. For this first version, the trigger should be focused on incident creation only (filter by `event.operation` / payload so only new incidents fire the workflow). 

#### 2. Create Incident (Action)

Creates a new incident in FireHydrant via the REST API (name, severity, description/summary, and other supported fields). Enables workflows like opening an incident from monitoring alerts, error tracking, or other SuperPlane triggers—similar to the existing PagerDuty or Rootly Create Incident action.

### Acceptance Criteria (for SuperPlane team)

* [ ] integration works
* [ ] starter components work
* [ ] has proper tests
* [ ] has proper documentation
* [ ] passed code quality review
* [ ] passed functionality review
* [ ] passed ui/ux review

PR #3255, opened 2026-02-25
feat: Add FireHydrant integration

felixgateru:
Implements #3004 

## What changed
The pr implements the requested FireHydrant Integration. This integration allows listening to incident creation events for further processing and also allows creating incidents from the superplane workflow.

## Why
This is to solve this issue: #3004

## How
The pr is of two components:
- **firehydrant.onIncident** trigger - on setup creates a webhook on FireHydrant to receive `incident`  events, `incident.private`  events or both. The webhooks are HMAC signed and verified.
- **firehyrant.createIncident** action - uses the FireHydrant API to create an incident. The authentication is provided by the user API key sent in the request.

The frontend is also updated for the above chnages.

## Demo
[Screencast From 2026-02-25 11-55-19.webm](https://github.com/user-attachments/assets/e8e853bb-90fb-4394-8f11-8a53f5f00cd5)

### DIFF
--- a/pkg/integrations/firehydrant/on_incident_test.go
+++ b/pkg/integrations/firehydrant/on_incident_test.go
@@ -21,7 +21,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 		code, err := trigger.HandleWebhook(core.WebhookRequestContext{
 			Headers:       http.Header{},
 			Configuration: map[string]any{},
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        &contexts.EventContext{},
 		})
 
@@ -38,7 +38,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Body:          body,
 			Headers:       headers,
 			Configuration: map[string]any{},
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        &contexts.EventContext{},
 		})
 
@@ -56,7 +56,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Configuration: map[string]any{
 				"current_milestone": []any{"started"},
 			},
-			Webhook: &contexts.WebhookContext{Secret: ""},
+			Webhook: &contexts.NodeWebhookContext{Secret: ""},
 			Events:  eventContext,
 		})
 
@@ -75,7 +75,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Body:          body,
 			Headers:       headers,
 			Configuration: map[string]any{},
-			Webhook:       &contexts.WebhookContext{Secret: secret},
+			Webhook:       &contexts.NodeWebhookContext{Secret: secret},
 			Events:        &contexts.EventContext{},
 		})
 
@@ -94,7 +94,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Body:          body,
 			Headers:       headers,
 			Configuration: map[string]any{},
-			Webhook:       &contexts.WebhookContext{Secret: secret},
+			Webhook:       &contexts.NodeWebhookContext{Secret: secret},
 			Events:        eventContext,
 		})
 
@@ -114,7 +114,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Body:          body,
 			Headers:       headers,
 			Configuration: map[string]any{},
-			Webhook:       &contexts.WebhookContext{Secret: secret},
+			Webhook:       &contexts.NodeWebhookContext{Secret: secret},
 			Events:        eventContext,
 		})
 
@@ -151,7 +151,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 			Configuration: map[string]any{
 				"current_milestone": []any{"started"},
 			},
-			Webhook: &contexts.WebhookContext{Secret: secret},
+			Webhook: &contexts.NodeWebhookContext{Secret: secret},
 			Events:  eventContext,
 		})
 
@@ -199,7 +199,7 @@ func Test__OnIncident__HandleWebhook(t *testing.T) {
 				"current_milestone": []any{"started"},
 				"severities":        []any{"SEV1", "SEV0"},
 			},
-			Webhook: &contexts.WebhookContext{Secret: secret},
+			Webhook: &contexts.NodeWebhookContext{Secret: secret},
 			Events:  eventContext,
 		})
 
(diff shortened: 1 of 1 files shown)


AleksandarCole:
@felixgateru looks great! Few suggested changes:

In the integration copy, add info what kind of permissions are needed for the key
This could be clearer:
image
As I wasn't 100% sure what to select here:

image


Details tab improvements (both components):

- Too many timestamps - keep just one at the top
- If possible provide a link to the issue
- 
image


Moved to code review.

felixgateru:
> @felixgateru looks great! Few suggested changes:
> 
> In the integration copy, add info what kind of permissions are needed for the key This could be clearer: image As I wasn't 100% sure what to select here:
> 
> image
> Details tab improvements (both components):
> 
> * Too many timestamps - keep just one at the top
> * If possible provide a link to the issue
> 
> image
> Moved to code review.

Update the above to:
Screenshot From 2026-02-25 17-35-58
 and 
 
Screenshot From 2026-02-25 17-53-46

pkg/integrations/firehydrant/common.go
```diff
@@ -0,0 +1 @@
+package firehydrant
```
lucaspin: Empty file

pkg/integrations/firehydrant/create_incident_test.go
```diff
@@ -0,0 +1,393 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/test/support/contexts"
+)
+
+func Test__CreateIncident__Configuration(t *testing.T) {
```
lucaspin: No need to have unit tests for Configuration()

pkg/integrations/firehydrant/create_incident_test.go
```diff
@@ -0,0 +1,393 @@
...
+		assert.Equal(t, 99, incident.Number)
+		assert.Equal(t, "DB unreachable", incident.Summary)
+		assert.Equal(t, "Primary database is not responding", incident.Description)
+		assert.Equal(t, "started", incident.CurrentMilestone)
+		assert.Equal(t, "SEV2", incident.Severity)
+		assert.Equal(t, "P2", incident.Priority)
+	})
+}
+
+func Test__CreateIncident__Cancel(t *testing.T) {
```
lucaspin: Also no need to have unit tests for no-op methods, or metadata - Name(), Icon(), ...

pkg/integrations/firehydrant/example_data_on_incident.json
```diff
@@ -0,0 +1,69 @@
+{
```
lucaspin, line 1: The examples are incorrect. All emitted events follow the format:

```json
{
  "type": "...",
  "timestamp": "...",
  "data": { ... }
}
```

pkg/integrations/firehydrant/firehydrant.go
```diff
@@ -0,0 +1,120 @@
...
+		return fmt.Errorf("error creating client: %v", err)
+	}
+
+	// Validate connection by listing severities
+	severities, err := client.ListSeverities()
+	if err != nil {
+		return fmt.Errorf("error listing severities: %v", err)
+	}
+
+	ctx.Integration.SetMetadata(Metadata{Severities: severities})
```
lucaspin: Don't save severities in the metadata. Doing that will lead to new created severities not being shown unless we refresh it.

pkg/integrations/firehydrant/firehydrant_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/test/support/contexts"
+)
+
+func Test__FireHydrant__Configuration(t *testing.T) {
```
lucaspin: Same comment about unit tests for static methods

pkg/integrations/firehydrant/on_incident.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (t *OnIncident) Color() string {
+	return "gray"
+}
+
+func (t *OnIncident) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "subscriptions",
```
lucaspin: Looking at the available subscriptions [here](https://docs.firehydrant.com/docs/webhooks), I don't think we should have this here. I think the `onIncident` trigger should subscribe to `incidents` events only. We can have separate triggers for the other subscriptions - onPrivateIncident, onChangeEvent, onSignal. This allows us to have more specific filters on each trigger.

For example, on the `onIncident` trigger, it would be good to allow filtering per "state" - started / acknowledged / mitigated / resolved, similar to how we have with the PagerDuty one.

pkg/integrations/firehydrant/on_incident.go
```diff
@@ -0,0 +1,324 @@
...
+						{Label: "SEV2", Value: "SEV2"},
+						{Label: "SEV3", Value: "SEV3"},
+						{Label: "SEV4", Value: "SEV4"},
+						{Label: "SEV5", Value: "SEV5"},
+						{Label: "UNSET", Value: "UNSET"},
+						{Label: "MAINTENANCE", Value: "MAINTENANCE"},
+						{Label: "GAMEDAY", Value: "GAMEDAY"},
+					},
+				},
+			},
```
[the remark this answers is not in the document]
lucaspin: This is a valid comment. I think the IntegrationResource configuration field already supports a [`Multi` variation that uses a MultiSelect instead of a Select.](https://github.com/superplanehq/superplane/blob/main/pkg/configuration/field.go#L126)

pkg/integrations/firehydrant/on_incident.go
```diff
@@ -0,0 +1,324 @@
...
+						{Label: "Detected", Value: "detected"},
+						{Label: "Acknowledged", Value: "acknowledged"},
+						{Label: "Identified", Value: "identified"},
+						{Label: "Mitigated", Value: "mitigated"},
+						{Label: "Resolved", Value: "resolved"},
+						{Label: "Retrospective Started", Value: "retrospective_started"},
+						{Label: "Retrospective Completed", Value: "retrospective_completed"},
+						{Label: "Closed", Value: "closed"},
+					},
+				},
```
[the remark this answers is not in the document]
lucaspin: This seems like a valid concern. Also:

> Only trigger for incident updates matching the selected milestones. When set, only UPDATED operations are accepted. When empty, only newly created incidents trigger the workflow.

Not sure I like this also. I think it would be less confusing to have the "started" milestone in this list, and selected by default, without even mentioning operations.

forestileao:
@felixgateru 
image

I tested here and it is all good and working. Merging it. Thanks :)

cc: @AleksandarCole

Issue #2913, opened 2026-02-06
[LaunchDarkly] Base


AleksandarCole:
### Description

Create a **SuperPlane integration for LaunchDarkly** and **three starter components** (one trigger, two actions) so users can build workflows with LaunchDarkly in SuperPlane. This issue covers the base integration and the first components listed below.

**Link**: https://docs.launchdarkly.com/

### Connection

Explore LaunchDarkly's docs to choose how to connect. LaunchDarkly offers a REST API for actions and webhooks for events: see [API access tokens](https://docs.launchdarkly.com/home/account/api) and [Webhooks](https://docs.launchdarkly.com/home/account/webhooks). Decide on auth (e.g. API token scope) and how users register a webhook URL for triggers.

### First Components

‼️ **IMPORTANT**

Creating at least 1 `trigger` and 1 `action` component to test out the Base is required for completing this issue.

#### 1. On Feature Flag Change (Trigger)

A trigger that runs when LaunchDarkly sends webhook events about feature flag changes (e.g. flag updated, turned on/off, targeting changed). Consider allowing the user to filter by event type so workflows only run for the changes they care about. Enables workflows like notifying Slack or running a deployment when a flag is modified.

#### 2. Get Feature Flag (Action)

An action that fetches a feature flag by project key and flag key. Enables workflows that branch on flag state or display/sync flag details.

#### 3. Delete Feature Flag (Action)

An action that deletes a feature flag by project key and flag key. Enables workflows that remove deprecated or one-off flags.

### Acceptance Criteria (for SuperPlane team)

* integration works
* starter components work
* has proper tests
* has proper documentation
* passed code quality review
* passed functionality review
* passed ui/ux review

shiroyasha:
Integration base + 5 components could be hard to review at once.
It would stall progress on the PR for sure.

@AleksandarCole @lucaspin @forestileao lets review this.

AleksandarCole:
Base issue scoped to 3 components (trigger + Get + Delete). Update feature flag and Create feature flag moved to **#2928** – [LaunchDarkly] Expand integration.

devroy10:
Hi, I am interested in this and would like to work on creating the base integration and the three starter components.

malikibric:
Starting to work this!

Manideepchopperla:
Hey @AleksandarCole I want to work on this issue. Can you assign me

PR #3256, opened 2026-02-25
feat: Add Launchdarkly integration

WashingtonKK:
Implements #2913 

## What changed

Adds a LaunchDarkly integration with the following components:

Actions:
- Get Feature Flag — fetches a feature flag by project key and flag key
- Delete Feature Flag — deletes a feature flag by project key and flag key

Trigger:
- On Feature Flag Change — listens to feature flag events from LaunchDarkly via a signed webhook

## Why
Enables SuperPlane workflows to react to feature flag changes in LaunchDarkly — for example, triggering a deployment rollback when a flag is turned off, orsending notifications when flags are created or deleted.

## How

**Trigger — On Feature Flag Change**                                                 
  - Fires when LaunchDarkly sends a feature flag event to the configured webhook   
  - Optional event type filter to scope the trigger to specific kinds of changes
  - Emits a launchdarkly.flag.<action> event (e.g. launchdarkly.flag.updateOn) with
   the full webhook payload

**Action — Get Feature Flag**
  - Fetches a feature flag from LaunchDarkly by project key and flag key
  - Emits a launchdarkly.flag output event with the full flag details

**Action — Delete Feature Flag**
  - Deletes a feature flag from LaunchDarkly by project key and flag key
  - Emits a launchdarkly.flag.deleted output event on success

**Integration setup**
  - Authenticates via a LaunchDarkly API access token configured by the user
  - Automatically creates a signed webhook in LaunchDarkly on canvas save — no
  manual setup required
  - Verifies incoming webhook requests using HMAC-SHA256 signatures (X-LD-Signature
   header)
  - Automatically deletes the webhook from LaunchDarkly when the trigger is removed

## Demo

https://youtu.be/UTdqRKFKCXs

### DIFF
--- a/pkg/integrations/honeycomb/on_alert_fired_test.go
+++ b/pkg/integrations/honeycomb/on_alert_fired_test.go
@@ -63,7 +63,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       http.Header{},
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        &contexts.EventContext{},
 			Metadata:      &contexts.MetadataContext{},
 		})
@@ -79,7 +79,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        &contexts.EventContext{},
 			Metadata:      &contexts.MetadataContext{},
 		})
@@ -96,7 +96,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          []byte(`not valid json`),
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        events,
 			Metadata:      &contexts.MetadataContext{},
 		})
@@ -117,7 +117,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        events,
 			Metadata:      meta,
 		})
@@ -139,7 +139,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        events,
 			Metadata:      meta,
 		})
@@ -157,7 +157,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        events,
 			Metadata:      &contexts.MetadataContext{},
 		})
@@ -178,7 +178,7 @@ func Test__OnAlertFired__HandleWebhook(t *testing.T) {
 			Headers:       h,
 			Body:          body,
 			Configuration: validConfig,
-			Webhook:       &contexts.WebhookContext{Secret: "test-secret"},
+			Webhook:       &contexts.NodeWebhookContext{Secret: "test-secret"},
 			Events:        events,
 			Metadata:      meta,
 		})


AleksandarCole:
Hey @WashingtonKK, didn't test this yet fully, changes are needed:

### Trigger (on Flag Change)
1. Why do we need `Events` field in the config?
2. I don't think we need that Note about webhook "LaunchDarkly webhook is being configured automatically". If user doesn't need to configure anything manually and we're creating things in the background then there is no reason to add new UI elements.

### Actions (get/delete)
Project and Flag keys need to be `IntegrationResource`, that has that single select with option to have expression
Check how it was implemented for any GitHub component that has repository in config:
image


### Other changes
Icon seems to be broken (show puzzle)

image

pkg/integrations/launchdarkly/on_feature_flag_change.go
```diff
@@ -0,0 +1,296 @@
...
+
+	if strings.TrimSpace(config.ProjectKey) == "" {
+		return fmt.Errorf("project key is required")
+	}
+
+	if strings.TrimSpace(config.FlagKey) == "" {
+		return fmt.Errorf("flag key is required")
+	}
+
+	if ctx.Integration == nil {
```
lucaspin: This is never nil

pkg/integrations/launchdarkly/on_feature_flag_change.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	if err := ctx.Integration.RequestWebhook(WebhookConfiguration{
+		ProjectKey: config.ProjectKey,
+		FlagKey:    config.FlagKey,
+	}); err != nil {
+		return err
+	}
+
+	var webhookURL string
```
lucaspin: No need to have this at all. The webhooks should be set up through the integration

pkg/integrations/launchdarkly/on_feature_flag_change.go
```diff
@@ -0,0 +1,296 @@
...
+func (t *OnFeatureFlagChange) Actions() []core.Action {
+	return []core.Action{}
+}
+
+func (t *OnFeatureFlagChange) HandleAction(ctx core.TriggerActionContext) (map[string]any, error) {
+	return nil, fmt.Errorf("action %s not supported", ctx.Name)
+}
+
+func (t *OnFeatureFlagChange) HandleWebhook(ctx core.WebhookRequestContext) (int, error) {
+	if ctx.Logger != nil {
```
lucaspin: No need to nil-check ctx.Logger

pkg/integrations/launchdarkly/webhook_handler.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create LaunchDarkly client: %w", err)
+	}
+
+	config := WebhookConfiguration{}
+	if err := mapstructure.Decode(ctx.Webhook.GetConfiguration(), &config); err != nil {
+		return nil, fmt.Errorf("failed to decode webhook configuration: %w", err)
+	}
+
+	resource := fmt.Sprintf("proj/%s:env/*:flag/%s", config.ProjectKey, config.FlagKey)
```
lucaspin: This means we create a separate webhook for each project/flag pair, right? That seems a bit too much for me, as a single project can have a lot of feature flags. Is it possible to set a webhook per project, and then filter by flag name on the trigger side? Check how the [github.onPush](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/on_push.go#L17) handles pushes for different refs, for example.

lucaspin:
@WashingtonKK just reviewed it again, not while messing on things on LaunchDarkly side, and have a couple of extra notes:

### launchdarkly.onFeatureFlagChange

The way the configuration is exposed is weird, and can lead to strange scenarios. For example, on the trigger itself, we select a flag, which lists all the available flags in the project. And after selecting the flag, you can select `action=createFlag`, which will lead to a trigger that never receives anything, because the flag has to be created for you to select it beforehand.

I think we can improve with the following:
- We don't select a specific flag. Instead, flags is a `AnyPredicateList` configuration field which allows you to select multiple flags, or just one. Check [github.onPush](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/on_push.go#L17) trigger for example.
- We need to include the environment here too. We can still create a single webhook per project, but we at least need a way to filter flag events for environments we don't care about.

With this, I think we can have more flexible configurations. For example:

```yaml
# Listen to flags being created
configuration:
  project: superplane
  environments: [development, production]
  actions: [createFlag]
  flags:
    - type: matches
      name: .*

# Listening to flags being deleted
configuration:
  project: superplane
  environments: [development, production]
  actions: [deleteFlag]
  flags:
    - type: matches
      value: .*

# Listening to changes on specific flag
configuration:
  project: superplane
  environment: [development, production]
  actions: [updateOn, updateOffVariation, updateFallthrough, updateRules, updateTargets]
  flags:
    - type: equals
      value: feature-flag-1
```

Note: we should also have good defaults. I think a sensible one would be: listen to changes on all flags on all environments.

### For other components

The `getFeatureFlag` / `deleteFeatureFlag` components are missing UI mappers for them, so their UI doesn't look great: details tab on the event just dumps everything on the event, we don't see the project / flag in the node itself.

pkg/integrations/launchdarkly/on_feature_flag_change.go
```diff
@@ -0,0 +1,294 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				Resource: &configuration.ResourceTypeOptions{
+					Type: "project",
+				},
+			},
+		},
+		{
+			Name:        "environments",
+			Label:       "Environments",
+			Type:        configuration.FieldTypeAnyPredicateList,
```
lucaspin: I don't think `environments` should be a `FieldTypeAnyPredicateList`. Reason: you probably don't have more than 5 environments. Differently from feature flags, which you can easily have 100+ flags. Also, it would be nice to list the environments on the UI for selection, so we should probably use a FieldTypeIntegrationResource field type, with [Multi=true](https://github.com/superplanehq/superplane/blob/main/pkg/configuration/field.go#L126) for environments instead.

pkg/integrations/launchdarkly/on_feature_flag_change.go
```diff
@@ -0,0 +1,294 @@
...
+		return fmt.Errorf("failed to decode configuration: %w", err)
+	}
+
+	if strings.TrimSpace(config.ProjectKey) == "" {
+		return fmt.Errorf("project key is required")
+	}
+
+	return ctx.Integration.RequestWebhook(WebhookConfiguration{
+		ProjectKey:      config.ProjectKey,
+		WebhooksBaseURL: ctx.Webhook.GetBaseURL(),
```
lucaspin: Why do you need this to pass the base URL here? The webhook handler already has access to the webhook URL through its context

pkg/integrations/launchdarkly/webhook_handler_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/test/support/contexts"
+)
+
+// testHandlerWebhookContext implements core.WebhookContext for webhook handler tests.
+type testHandlerWebhookContext struct {
```
lucaspin: Can we implement this dummy context in [test/support/contexts/contexts.go](https://github.com/superplanehq/superplane/blob/main/test/support/contexts/contexts.go)? [WebhookContext](https://github.com/superplanehq/superplane/blob/main/test/support/contexts/contexts.go#L30) already exists there, but that one is actually implementing the [NodeWebhookContext](https://github.com/superplanehq/superplane/blob/main/pkg/core/trigger.go#L135), so maybe we can rename that one to `NodeWebhookContext`, and move this one there as `WebhookContext`.

Moving it there would allow other integrations to use for their tests too.

PR #3273, opened 2026-02-25
fix: Implement WebSocket origin checking in production

salignatmoandal:
## What

In production, WebSocket connections are now restricted to allowed origins only. In development, all origins are still accepted (no change for local dev).

## Why

Previously, the server accepted WebSocket connections from any origin. That could allow a third-party site to open a WebSocket to your SuperPlane instance using the user's cookies. Restricting origins in production reduces that risk and aligns with common security practice for WebSockets.

## How

- **`allowedWebSocketOrigins()`** builds the list of allowed origins:
  - In **development** (`APP_ENV=development`): returns no list → all origins accepted (unchanged behavior).
  - In **production**: uses `WEBSOCKET_ALLOWED_ORIGINS` if set (comma-separated), otherwise derives a single origin from `BASE_URL` (scheme + host).
- The WebSocket **`CheckOrigin`** callback now allows a connection only when:
  - there is no restriction list (dev / fallback), or
  - the request’s `Origin` header is in the allowed list.

## Configuration

- **`BASE_URL`**: In production, the origin derived from this URL is allowed (e.g. `https://app.example.com` → `https://app.example.com`).
- **`WEBSOCKET_ALLOWED_ORIGINS`** (optional): Comma-separated list of origins when you need multiple (e.g. `https://app.example.com,https://dashboard.example.com`).
- No change needed for local development; all origins remain accepted when `APP_ENV=development`.

### DIFF
--- a/pkg/public/server.go
+++ b/pkg/public/server.go
@@ -86,6 +86,31 @@ func (s *Server) WebsocketHub() *ws.Hub {
 	return s.wsHub
 }
 
+// allowedWebSocketOrigins returns the list of origins allowed for WebSocket connections.
+// In development (isDev true) returns nil (no restriction). Otherwise uses
+// WEBSOCKET_ALLOWED_ORIGINS if set (comma-separated), or the origin derived from baseURL (scheme + host).
+// Returns nil on parse error or when no origins are configured (fallback: allow all for backward compatibility).
+func allowedWebSocketOrigins(baseURL string, isDev bool) []string {
+	if isDev {
+		return nil
+	}
+	if env := os.Getenv("WEBSOCKET_ALLOWED_ORIGINS"); env != "" {
+		var list []string
+		for _, s := range strings.Split(env, ",") {
+			if o := strings.TrimSpace(s); o != "" {
+				list = append(list, o)
+			}
+		}
+		return list
+	}
+	u, err := url.Parse(baseURL)
+	if err != nil {
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


shiroyasha:
Hey @salignatmoandal 🙌 I like to idea of this PR. 

I would like to see some small code cleanup before we merge it. 

1/ Lets create a new file dedicated to this
2/ Instead of adding low level details in the main server code, lets extract the logic into the other file as much as possible
3/ In what scenarios we want anything other than BASE_URL? 🤔 Do we have a real use for that additional environment variable?

A closure could work well for this use case. e.g.:

```go
// pkg/server/websocket_origin_checker.go

func WebsocketOriginChecker() func(*http.Request) bool {
  var allowedOrigins
  // calculate allowed origins here ...

  return func(r *http.Request) bool {
    // verify here    
  }
}

// pkg/server/server.go

server := &Server{
  upgrader: &websocket.Upgrader{
    CheckOrigin: WebsocketOriginChecker()
  }
}
```

salignatmoandal:
Yep, I followed that approach.

I  moved the WebSocket origin logic to a dedicated file, and server.go now only wires `CheckOrigin` via a closure (`newWebSocketCheckOrigin(appEnv, baseURL)`), so the low-level details are out of the main server setup.

 I also simplified the config to rely on `BASE_URL` in production, while keeping development unrestricted.

lucaspin:
Fixed in https://github.com/superplanehq/superplane/pull/4328

PR #3322, opened 2026-03-01
feat: Add New Relic integration

vukasin-djuricic:
Implements #2552

## Summary

This PR adds a New Relic integration to SuperPlane, enabling users to react to alerts and query telemetry data from New Relic.

Authorization uses two keys:
- **User API Key** — for NerdGraph/NRQL queries (no additional permissions needed)
- **License Key** — for metric ingestion via the Metric API

Supports both **US** and **EU** data center regions.

## Components

| Name | Type | Description |
|------|------|-------------|
| **Run NRQL Query** | Action | Executes a NRQL query via the NerdGraph API. Supports configurable timeout (up to 120s). |
| **Report Metric** | Action | Sends custom metrics (gauge, count, summary) to the New Relic Metric API. |
| **On Issue** | Trigger | Receives New Relic alert issue webhooks. Validates payloads and optional Bearer token auth. |

## Implementation Details

- **Backend**: `pkg/integrations/newrelic/` — client, integration, two components, one trigger, webhook handler, example data
- **Frontend**: `web_src/src/pages/workflowv2/mappers/newrelic/` — mappers for all three components with execution details, metadata, and event sections
- **Docs**: Auto-generated via `make gen.components.docs` → `docs/components/NewRelic.mdx`
- **Tests**: Unit tests for `Sync`, `Execute`, `Setup`, webhook parsing, and edge cases across all components
- **HTTP safety**: All HTTP requests use `http.NewRequestWithContext` to avoid context leaks
- **GraphQL validation**: `ValidateCredentials` checks the response body for GraphQL errors (NerdGraph returns HTTP 200 even for auth failures)

## Video Demo

https://youtu.be/_6P0luCxrpA

## Checklist

- [x] Backend implementation in `pkg/integrations/newrelic/`
- [x] Frontend mappers in `web_src/src/pages/workflowv2/mappers/newrelic/`
- [x] Unit tests for all components and triggers
- [x] Example output data for components
- [x] Component documentation generated with `make gen.components.docs`
- [x] Integration icon (SVG) added and registered
- [x] Lint passes (`revive`)
- [x] All commits signed off (DCO)

### DIFF
--- a/pkg/integrations/newrelic/webhook_handler.go
+++ b/pkg/integrations/newrelic/webhook_handler.go
@@ -23,12 +23,12 @@ type WebhookProvisioningMetadata struct {
 const defaultPayloadTemplate = `{
   "issueId": {{json issueId}},
   "issueUrl": {{json issueUrl}},
-  "title": {{json title.[0]}},
+  "title": {{#if title.[0]}}{{json title.[0]}}{{else}}null{{/if}},
   "priority": {{json priority}},
   "state": {{json state}},
   "policyName": {{#if policyName.[0]}}{{json policyName.[0]}}{{else}}null{{/if}},
   "conditionName": {{#if conditionName.[0]}}{{json conditionName.[0]}}{{else}}null{{/if}},
-  "accountId": {{json accumulations.tag.account.[0]}},
+  "accountId": {{#if accumulations.tag.account.[0]}}{{json accumulations.tag.account.[0]}}{{else}}null{{/if}},
   "createdAt": {{json createdAt}},
   "updatedAt": {{json updatedAt}},
   "sources": {{json sources}}


pkg/integrations/newrelic/on_issue.go
```diff
@@ -0,0 +1,332 @@
...
+		// Optional fields can be stored as null in configuration.
+		if strings.Contains(errText, "not a string") {
+			return "", nil
+		}
+
+		return "", err
+	}
+
+	return string(value), nil
+}
```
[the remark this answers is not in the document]
vukasin-djuricic: Follows the pattern from the Harness integration, which also handles null config values. Happy to refactor into a shared utility if preferred.

AleksandarCole:
Hey @vukasin-djuricic, thanks for trying this one!

I tried testing this, and there are quite a few issues. It could be my lack of domain knowledge with New Relic, I will list them bellow, but **I strongly encourage you joining our Discord as comms will be faster**.  

### Action components do not work
Tried chaining together `reportMetric` and running `NRQL Query` to get said metric back, it returned null. Not sure which is failing - pushing data or fetching it
image

### Trigger component doesn't work
Even though configured properly - the trigger component did not receive and show an incident that was triggered on New Relic

### Trigger component webhook setup
It is possible to manage webhooks via UserAPI, there is no reason for us to push user to set up these things manually. Check dash0 pattern. 

Those are only the critical ones, I will not list the UX improvements yet until the core functionality is achieved.

vukasin-djuricic:
GitHub PR komentar:                                                                                                                                                                                       
                                                                                                                                                                                                            
  ## Update                                                                                                                                                                                                 
                                                                                                                                                                                                            
  All BugBot review comments addressed across 3 rounds:                                                                                                                                                     
                                                                                                                                                                                                            
  **Round 1:** mapstructure tags ✅, null webhookSecret handling ✅ (+test), empty-string filtering ✅, interval.ms ✅, accountId quoting (false positive — used directly, not with quoteGraphQL) ✅,       
  duplicated optionalIntegrationConfig (acknowledged, out of scope)                                                                                                                                         
                                                                                                                                                                                                            
  **Round 2:** accountId changed from int→any (NR sends as string) ✅, zero-value tip fixed (only shows when ALL values null, not 0) ✅                                                                     
                                                                                                                                                                                                            
  **Round 3:** accountId validation added in NewClient() for defensive GraphQL injection prevention ✅                                                                                                      
                                                                                                                                                                                                            
  ### Video demo                                                                                                                                                                                            
  https://www.youtube.com/watch?v=vfT2Lybz_cI                                                                                                                                                               
                                                                                                                                                                                                            
  ### Pre-existing issue found during testing

  `CanvasCleanupWorker.processCanvas()` (line 118) hard-deletes CanvasNode records directly instead of calling `models.DeleteCanvasNode()`. This means webhooks are not soft-deleted when a canvas is
  deleted, so `WebhookCleanupWorker` never cleans up auto-provisioned resources (NR destinations/channels, LD webhooks, etc.) in external services. Affects all integrations with webhook handlers. Happy to
   open a separate issue if helpful.

AleksandarCole:
`newrelic.onIssue` provisioning is mostly working (destination/channel are created), but the default webhook payload template is brittle and can generate payloads that SuperPlane rejects (HTTP 400), especially in New Relic test/notification paths. As a result, real New Relic issues may fire while no trigger event appears in Canvas. The `reportMetric` and `runNRQLQuery` components are fine now; the gap is specifically in the onIssue webhook payload mapping/compatibility.

### onIssue improvements
To properly fix the `newrelic.onIssue` trigger, we should harden both outbound payload generation and inbound parsing.

- **Harden the default New Relic channel payload template**
  - Avoid brittle nested fields that may be missing.
  - Ensure generated JSON is always valid.
  - Provide safe defaults for optional fields.

- **Make `HandleWebhook` tolerant to payload variants**
  - Normalize different shapes (string vs array, optional/missing fields).
  - Parse defensively before applying status/priority filters.

- **Validate provisioning during setup**
  - After channel creation/update, run a test notification.
  - Fail setup with a clear actionable error if payload is invalid.

- **Migrate existing channels**
  - Existing channels do not auto-update when template code changes.
  - Add a reprovision/update path so old channels receive the corrected template.

- **Improve observability**
  - Log explicit webhook parse/validation failure reasons.
  - Make 400-class failures diagnosable from logs/UI.

## Expected outcome

- New Relic issues consistently reach SuperPlane.
- Canvas trigger events are reliably created.
- Fewer silent failures caused by template/payload mismatch.

@forestileao I need sanity check on this one ☝️ 
Also, I think we can start with code review. 

____

@vukasin-djuricic some additional UX changes:

### Integration setup
Since we are creating webhooks automatically - there's no need to have `optional` secret. Let's have a secret that is automatically set in the background and remove this input from UI. 

### onIssue
The details tab that I see and the one that is in your video look very different. Let's have details look like in your video with limited data, also let's just have single timestamp at the top instead of two of them. 
image

### Report metric and Run Query
For the `Value` configuration field we are using "expression" input - the same one that is used in `filter` and `if` components, this doesn't make sense since these inputs are for writing `conditions`, instead, use same input like you use for Metric Name. 

Same for NRQL Query field

pkg/integrations/newrelic/newrelic.go
```diff
@@ -0,0 +1,195 @@
...
+			Description: "New Relic License Key for metric ingestion",
+		},
+		{
+			Name:        "webhookSecret",
+			Label:       "Webhook Secret",
+			Type:        configuration.FieldTypeString,
+			Required:    false,
+			Sensitive:   true,
+			Description: "Optional secret for validating incoming webhook requests. If set, New Relic must send an Authorization: Bearer <secret> header.",
+		},
```
forestileao: What if we always set an webhook secret. But not a webhook secret that the user provided? On webhook context from webhook handler it is possible to Get and Set secret using `GetSecret` and `SetSecret`. Then it would enforce to always use secure authenticated webhooks in Superplane.

We have some examples in the codebase (try searching by .SetSecret)

pkg/integrations/newrelic/run_nrql_query.go
```diff
@@ -0,0 +1,214 @@
...
+			Placeholder: "SELECT count(*) FROM Transaction SINCE 1 hour ago",
+		},
+		{
+			Name:        "timeout",
+			Label:       "Timeout (seconds)",
+			Type:        configuration.FieldTypeNumber,
+			Required:    false,
+			Default:     30,
+			Description: "Query timeout in seconds",
+		},
```
forestileao, line 93: Here it is possible to add 
`
				Number: &configuration.NumberTypeOptions{
					Min: func() *int { min := 0; return &min }(),
					Max: func() *int { max := maxNRQLTimeout; return &max }(),
				},
			},
`

So the UI or Agent will also know that 120 is the max timeout

pkg/integrations/newrelic/run_nrql_query.go
```diff
@@ -0,0 +1,214 @@
...
+	responseBody, err := client.NerdGraphQuery(context.Background(), graphQLQuery)
+	if err != nil {
+		return fmt.Errorf("failed to execute NRQL query: %v", err)
+	}
+
+	var response NerdGraphNRQLResponse
+	err = json.Unmarshal(responseBody, &response)
+	if err != nil {
+		return fmt.Errorf("error parsing response: %v", err)
+	}
```
forestileao: It would be nice if every client response is already in the retuning struct. then you wouldn't need to unmarshal the json outside the client.

pkg/integrations/newrelic/run_nrql_query.go
```diff
@@ -0,0 +1,214 @@
...
+	} `json:"data"`
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+func quoteGraphQL(s string) string {
+	b, _ := json.Marshal(s)
+	return string(b)
+}
```
forestileao: And we could move this to the client.go as well

forestileao:
@vukasin-djuricic The code looks good at all, just added minor comments

vukasin-djuricic:
Thanks for the thorough review! Working on the following based on your feedback:                                                                                                                          

  Security / Webhook:
  - Switching to auto-generated secrets using GetSecret/SetSecret — removing the user-facing webhookSecret field entirely
  - Hardening the payload template with Handlebars conditionals for optional fields (policyName, conditionName)

  Client refactor (forestileao):
  - Moving NRQL query execution logic to client.go
  - Returning typed structs from client methods instead of raw bytes
  - Adding NumberTypeOptions (min/max) for the timeout field

  UX (AleksandarCole):
  - Changing value and query fields from expression to string/text input type
  - Cleaning up the details tab — single timestamp, less noise

  Will push an update shortly.

docker-compose.dev.yml
```diff
@@ -97,6 +97,11 @@ services:
       rabbitmq:
         condition: service_healthy
 
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+
```
forestileao: Why do we need this?

vukasin-djuricic: Removed! It was added locally to fix an EMFILE (too many open files) error during Vite dev server startup — not related to this PR.
Sorry for the noise.

forestileao (approved):

forestileao:
@vukasin-djuricic Everything looks good now. Merging this PR. Thanks for your Contribution <3

Issue #3321, opened 2026-03-01
Dash0 On Notification: not receiving events


AleksandarCole:
**Description:**  
The Dash0 "On Notification" trigger (`dash0.onNotification`) is set up (integration configured, webhook URL in place) but no events are received when Dash0 sends notifications. Workflows that depend on this trigger never run.

When the same notification is configured in Dash0 to use the regular, core SuperPlane webhook (generic webhook URL), the event is received. So the failure is on the SuperPlane Dash0 component side, not on Dash0 delivery.

**Steps to reproduce:**  
1. Configure the Dash0 integration in SuperPlane and copy the webhook URL.
2. In Dash0, configure notifications to POST to that URL.
3. Trigger a notification from Dash0 (e.g. alert condition met).
4. (Observe: no execution is started in SuperPlane; no events received.)

*(Please add any specific steps, environment details, or log excerpts that show the webhook being called or failing.)*

**Expected / Actual:**  
Expected: When Dash0 sends a notification to the configured webhook URL, the trigger fires and a workflow execution starts. Actual: No executions are created; events do not appear to be received.

*(If you have logs (SuperPlane server or Dash0 outbound requests), error messages, or a screenshot of the integration/webhook config, adding them to the issue after creation will help triage.)*

felixgateru:
Taking this task

felixgateru:
## Problem
The issue is a mismatch in the webhook payload, specifically the labels field for notifications sent for `Synthetic` checks. In the documentation, the labels field is formatted as:
```json
"labels": [
        {
          "key": "environment",
          "value": {
            "stringValue": "production"
          }
        },
        {
          "key": "region",
          "value": {
            "stringValue": "us-west-2"
          }
        }
      ]
```
an array of key-value pair objects, but the actual payload for notifications for `Synthetic` checks received is as follows:
```json
"labels": [
                [
                    "0",
                    {
                        "key": "dash0.resource.type",
                        "value": {
                            "stringValue": "synthetic"
                        }
                    }
                ],
                [
                    "1",
                    {
                        "key": "dash0.synthetic_check.attempt_id",
                        "value": {
                            "stringValue": "73768e2c"
                        }
                    }
                ],
                [
                    "2",
                    {
                        "key": "dash0.synthetic_check.failed_critical_assertions",
                        "value": {
[...]

PR #3336, opened 2026-03-02
feat: separate dash0 onNotification trigger into onAlertNotification trigger and onSyntheticCheckNotification trigger

felixgateru:
Resolves: #3321 

## What changed:
This PR separates the current onNotification trigger into the onAlertNotification and onSyntheticCheckNotification triggers to handle the emitted payloads separately.

## Why: 
There is a mismatch between the webhook payload for the two notification types. 

## How: 
The pr replicates the existing logic for onNotification into the two triggers.
A guard is added on each trigger to only accept either alert notifications or synthetic check notifications.
The UI mappers are updated for the change.

## Related issues:
Fixes: #3321
 
## Demo:

https://www.youtube.com/watch?v=TD0mBZJxeRM


pkg/integrations/dash0/example_data_on_notification.json
```diff
@@ -33,20 +33,6 @@
...
-          "value": {
-            "stringValue": "api"
-          }
-        },
-        {
-          "key": "env",
-          "value": {
-            "stringValue": "prod"
-          }
-        }
```
forestileao, line 49: The solution here is not fixing the format instead of removing it from the output?

AleksandarCole:
@felixgateru works well. @WashingtonKK can you do code review please?

Issue #2906, opened 2026-02-06
[Dash0] Expand integration – 7 components


AleksandarCole:
### Description

Expand the **existing SuperPlane integration for Dash0** with **seven additional components** (one trigger, six actions) so users can build richer workflows with Dash0 in SuperPlane. The base integration and **Run PromQL Query** are already implemented; this issue covers adding the components listed below.

**Link**: https://www.dash0.com/

### Connection method

The current Dash0 integration connects via API token (Bearer). It can be **expanded or updated** if any of the new components require a different connection method (e.g. additional auth or endpoints) or extra setup/settings—for example, ingest permissions for Send Log Event, or config-API access for check rules and synthetic checks. Implementers should extend the base integration as needed to support the new components.

#### 1. On Alert Event (Trigger)

Single webhook trigger for Dash0 alerting. User configures the SuperPlane webhook URL as a notification channel in Dash0. The trigger runs when a check fails (fired) and/or when it recovers (resolved); event type is selectable (e.g. fired only, resolved only, or both). Outputs include check identity, severity, labels, summary/description, event type (fired/resolved), and timestamp so downstream steps can branch or enrich notifications.

#### 2. Send Log Event (Action)

Sends one or a small batch of log records from the workflow into Dash0 via OTLP HTTP ingestion. Workflow events (e.g. deployment completed, approval granted, alert acknowledged) become log lines in Dash0’s Logs Explorer and can be correlated with traces and metrics. Useful for audit trails and for tying workflow activity to observability data.

#### 3. Get Check Details (Action)

Fetches full context for a check by ID (e.g. from an On Alert Event payload). Returns rich data such as the check’s query, thresholds, current value, affected resources or services, and optionally recent history. Use case: the webhook gives only a check ID and summary; this action enriches the workflow so you can post a detailed summary to Slack, pre-fill a postmortem, or branch on affected services. Depends on Dash0 exposing a failed-check-by-ID (or equivalent) in their API.

#### 4. Create Synthetic Check (Action)

[...]

PR #3359, opened 2026-03-04
feat: Add dash0 get synthetic check component

felixgateru:
Related to: #2906

## What changed:
Added new "Get HTTP Synthetic Check" component for the Dash0 integration that retrieves check configuration and operational metrics

## Why:
Enable users to fetch the synthetic check status and performance metrics
Provide best-effort metrics fetching

### Backend:
Created GetHTTPSyntheticCheck component that calls Dash0 REST API (/api/synthetic-checks/{id}) for configuration
Fetches 8 operational metrics from Dash0 Prometheus API using instant queries

### Frontend:
Created mapper with execution details showing execution time, name, target (method + URL), expected assertions, scheduling, notification channels, metrics, and enabled status

## Demo
https://www.youtube.com/watch?v=fxHUdl4A4uI

### DIFF
--- a/docs/components/Dash0.mdx
+++ b/docs/components/Dash0.mdx
@@ -389,14 +389,15 @@ Note: Metrics are fetched on a best-effort basis. If Prometheus metrics are unav
       }
     },
     "metrics": {
-      "avgDuration24hMs": 540,
-      "avgDuration7dMs": 520,
-      "criticalRuns24h": 3,
-      "criticalRuns7d": 12,
-      "healthyRuns24h": 58,
-      "healthyRuns7d": 390,
-      "totalRuns24h": 61,
-      "totalRuns7d": 402
+      "avgDuration24hMs": 1004,
+      "avgDuration7dMs": 1004,
+      "criticalRuns24h": 2275,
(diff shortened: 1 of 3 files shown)


AleksandarCole:
Hey @felixgateru this one is good. 
I would like to do few UX changes:

1. Status is currently just `fetched` - it would be nice if it could reflect the state of the check (critical/degraded/passing)
2. We should also have different output channel matching the states
3. Details tab could include the link to the check itself
4. The metadata in the node contains check ID but it makes more sense to me to have check name instead

felixgateru:
Hey @AleksandarCole 
I have updated the status to reflect the check and updated the node metadata to include the check Name. As shown in the attached screenshots below:
Screenshot From 2026-03-05 12-56-06
and  
Screenshot From 2026-03-05 12-51-19

However, I encountered the following challenges:
1. From the synthetic check run metrics, I can only obtain whether the outcome of a run was `Healthy` or `Critical`, hence only those two states are supported.
2. The API response for the fetch method does not return a public URL for the Synthetic Check in the payload. For that reason, I have not included a URL in the details. However, I may not have fully understood the comment.

Issue #3335, opened 2026-03-02
Allow me to rename a canvas


shiroyasha:
Currently, there is no way to do it.

AleksandarCole:
@shiroyasha there is but it's not intuitive. Better UX is needed - @ropsii 

<img width="379" height="312" alt="Image" src="https://github.com/user-attachments/assets/2d557da5-f25f-4956-87bb-7feb2d95783e" />

<img width="632" height="423" alt="Image" src="https://github.com/user-attachments/assets/2a4289ac-4288-4f45-8493-1e99dac53d3f" />

shiroyasha:
The new settings panel addressed this issue.

PR #3367, opened 2026-03-04
feat: Enable canvas rename

usernamenenad:
This PR addresses #3335.

It reuses existing _UpdateCanvas_ service to rename a canvas. There is a sanity check if the same canvas name exists already, showing toast on successful rename.


shiroyasha:
@usernamenenad thank you for opening this PR. 

It looks good, however @forestileao already introduced a setting page on all canvases that addresses this concern as well.

Closing it.

PR #3397, opened 2026-03-08
feat: Canvas Settings Page

forestileao:
## Changes:

Added Canvas Settings page with the following changes:
- Change name/description (note that rename doesn't work now)
- Turn on/off versioning
- Let's not call it sandbox mode
- Whatever is in org settings should be a default state for new Canvases
- Fix the links so they are actual links and not JS
- deleting a Canvas should not try to open it


shiroyasha:
Should we add some tests for this? 🤔 
Doesn't feel right that we haven't covered it at all.

forestileao:
@shiroyasha I will add e2e tests

forestileao:
Updated to correct behavior:

Org canvas versioning ON => all canvases effectively ON.
Org canvas OFF => canvas can still be ON/OFF

Decision made based on 2MFA security enforcements

PR #3399, opened 2026-03-09
feat: add dash0 createCheckRule, getCheckRule, updateCheckRule and deleteCheckRule components

felixgateru:
Implements #2906 

## What changed
Added four new CRUD actions to the Dash0 integration for managing alerting check rules (Prometheus-style alert rules):
Create Check Rule: Create new alert rules with PromQL expressions, thresholds, and notification settings
Get Check Rule: Retrieve check rule configuration by ID
Update Check Rule: Modify existing check rule settings
Delete Check Rule: Remove check rules

## Why
Enable users to automate Dash0 alert management directly within SuperPlane workflows. 

### Backend:
Created four component files in dash0 implementing client methods (CreateCheckRule, GetCheckRule, UpdateCheckRule, DeleteCheckRule) to interact with Dash0's REST API

### Frontend :
Created mappers in dash0 for UI display

## Demo
https://www.youtube.com/watch?v=Lll0hWS6-4w

### DIFF
--- a/pkg/integrations/dash0/create_check_rule.go
+++ b/pkg/integrations/dash0/create_check_rule.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/mitchellh/mapstructure"
@@ -179,26 +181,26 @@ func (c *CreateCheckRule) Configuration() []configuration.Field {
 			Label:    "Trigger Grace Period",
 			Type:     configuration.FieldTypeSelect,
 			Required: true,
-			Default:  "0s",
+			Default:  "0",
 			TypeOptions: &configuration.TypeOptions{
 				Select: &configuration.SelectTypeOptions{
 					Options: checkRuleGracePeriodOptions(),
 				},
 			},
-			Description: "How long the expression must be true before triggering",
+			Description: "Multiplier of the evaluation interval the expression must be true before triggering",
 		},
 		{
 			Name:     "keepFiringFor",
 			Label:    "Keep Firing Grace Period",
 			Type:     configuration.FieldTypeSelect,
 			Required: true,
-			Default:  "0s",
+			Default:  "0",
 			TypeOptions: &configuration.TypeOptions{
 				Select: &configuration.SelectTypeOptions{
 					Options: checkRuleGracePeriodOptions(),
 				},
 			},
-			Description: "How long to keep firing after expression becomes false",
+			Description: "Multiplier of the evaluation interval to keep firing after expression becomes false",
 		},
 		{
 			Name:        "labels",
@@ -307,15 +309,8 @@ func buildCheckRuleRequest(spec CreateCheckRuleSpec) CheckRuleRequest {
 		interval = "1m"
 	}
 
-	forDuration := spec.For
-	if forDuration == "" {
-		forDuration = "0s"
-	}
-
-	keepFiringFor := spec.KeepFiringFor
-	if keepFiringFor == "" {
-		keepFiringFor = "0s"
-	}
+	forDuration := multiplyInterval(interval, spec.For)
+	keepFiringFor := multiplyInterval(interval, spec.KeepFiringFor)
 
 	request := CheckRuleRequest{
 		Name:          spec.Name,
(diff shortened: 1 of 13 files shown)


forestileao:
@felixgateru Remember to run `make gen.components.docs` so the doc can be regenerated

pkg/integrations/dash0/create_check_rule.go
```diff
@@ -0,0 +1,383 @@
...
+			Label:       "Thresholds",
+			Type:        configuration.FieldTypeObject,
+			Required:    false,
+			Togglable:   true,
+			Description: "Threshold values for degraded and critical states. Required when using $__threshold in expression",
+			TypeOptions: &configuration.TypeOptions{
+				Object: &configuration.ObjectTypeOptions{
+					Schema: checkRuleThresholdsSchema(),
+				},
+			},
```
forestileao, line 146: Here it is possible to add a default value to thresholds?

I think it would be nice, since on UI when it is toggled it adds a default configuration that we just need to update. What do you think?

pkg/integrations/dash0/create_check_rule.go
```diff
@@ -0,0 +1,383 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				List: &configuration.ListTypeOptions{
+					ItemLabel: "Annotation",
+					ItemDefinition: &configuration.ListItemDefinition{
+						Type:   configuration.FieldTypeObject,
+						Schema: keyValueListSchema(),
+					},
+				},
+			},
+		},
```
forestileao, line 238: Here too, if it makes sense

pkg/integrations/dash0/delete_check_rule.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (c *DeleteCheckRule) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (c *DeleteCheckRule) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "checkRuleId",
+			Label:       "Check Rule ID",
```
forestileao: Let's rename this field name, label and variable to just checkRule to follow the `FieldTypeIntegrationResource` we've being addopting. Athough we know it is an id ;)

pkg/integrations/dash0/get_check_rule.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (c *GetCheckRule) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (c *GetCheckRule) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "checkRuleId",
```
forestileao: same here for renaming it to `checkRule`

pkg/integrations/dash0/update_check_rule.go
```diff
@@ -0,0 +1,327 @@
...
+func (c *UpdateCheckRule) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (c *UpdateCheckRule) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "checkRuleId",
+			Label:       "Check Rule ID",
+			Type:        configuration.FieldTypeIntegrationResource,
```
forestileao: same for renaming to `checkRule`

forestileao:
@felixgateru Overall, I only added minor comments, the rest looks good.
Just waiting for a functional test from @AleksandarCole ;)

pkg/integrations/dash0/create_check_rule.go
felixgateru: I have not found this in the UI, could you please share a screenshot of a default configuration so that I can replicate this. I am not sure what default values to put in the threshold fields. Currently, I have put only placeholders

pkg/integrations/dash0/create_check_rule.go
felixgateru: Please provide, if possible, a screenshot of the UI. I saw that they offer an empty key-value pair dialogue similar to what we have now.

felixgateru:
@AleksandarCole I have made the following changes on the pr as requested.
1. I improved the threshold input, I removed the unnecessary toggle on each of the fields, allowing better UX.
Screenshot From 2026-03-11 01-09-55

2. I have matched the number of options for check intervals with the dash0 UI:
image

3. I have changed the Trigger grace period and Keep firing for to be multipliers of the selected interval similar to what is on dash0 UI
Screenshot From 2026-03-11 01-10-07


4. The component now shows the name and not the id for better UX
[Screencast From 2026-03-11 07-44-35.webm](https://github.com/user-attachments/assets/40617e9b-9bdc-4683-8692-30dde5e84f01)


The create and update dialogues now have the exact feel as the dash0 UI

Issue #3250, opened 2026-02-24
Agent Issue: Silent dropping of unknown YAML fields


AleksandarCole:
### Issue 4: Silent dropping of unknown YAML fields

**Severity:** High — made Issue #3247 much harder to debug

**Problem:** The CLI silently drops unrecognized fields during `canvases update`. When incorrect field names like `integrationId` or `connectedIntegrationId` were used, the update succeeded with no warning — the field was simply gone from the output. This turned a single mistake into a long trial-and-error loop.

**Recommended fix:**

| Channel | Action |
|---------|--------|
| CLI | Warn on unrecognized fields during `canvases update` (or add a `--strict` mode). |

PR #3400, opened 2026-03-09
feat: yaml editor

AleksandarCole:
Adds a YAML preview allowing users to understand the YAML structure of the Canvas they are editing.

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -4176,6 +4176,11 @@ export function WorkflowPageV2() {
     [canvasId, updateCanvasMutation],
   );
 
+  const { yamlPayload, handleYamlViewCopy, handleYamlViewDownload } = useCanvasYaml({
+    nodes,
+    getYamlExportPayload,
+  });
+
   const isInitialCanvasBootstrapLoading =
     !canvas &&
     (canvasLoading ||
@@ -4378,10 +4383,6 @@ export function WorkflowPageV2() {
             : hasRunBlockingChanges
               ? "Save canvas changes before running"
               : undefined;
-  const { yamlPayload, handleYamlViewCopy, handleYamlViewDownload } = useCanvasYaml({
-    nodes,
-    getYamlExportPayload,
-  });
 
   const dataViewContent =
     topViewMode === "yaml" && yamlPayload ? (


web_src/src/pages/workflowv2/CanvasYamlView.tsx
```diff
@@ -0,0 +1,348 @@
...
+// ---------------------------------------------------------------------------
+
+function toMonacoSeverity(monaco: Monaco, severity: YamlDiagnostic["severity"]): number {
+  return severity === "error" ? monaco.MarkerSeverity.Error : monaco.MarkerSeverity.Warning;
+}
+
+/** Injects CSS for block-level decorations (idempotent). */
+function injectDecorationStyles() {
+  const styleId = "yaml-block-highlight-styles";
+  if (document.getElementById(styleId)) return;
```
shiroyasha: Don't mix React with with jquery style building 🙈

web_src/src/pages/workflowv2/CanvasYamlView.tsx
```diff
@@ -0,0 +1,348 @@
...
+    .${BLOCK_WARNING_CLASS} {
+      background: rgba(250, 204, 21, 0.12);
+      border-left: 3px solid rgba(234, 179, 8, 0.6);
+    }
+    .${BLOCK_ERROR_CLASS} {
+      background: rgba(239, 68, 68, 0.10);
+      border-left: 3px solid rgba(239, 68, 68, 0.5);
+    }
+  `;
+  document.head.appendChild(style);
```
shiroyasha: Same here. Is there a reason why we have to do it like this?

Good chance that lines like this could break our React app in unforeseen ways.

web_src/src/pages/workflowv2/CanvasYamlView.tsx
```diff
@@ -0,0 +1,348 @@
...
+  }, [yamlText]);
+
+  // Keep autocomplete data in sync without re-registering providers
+  useEffect(() => {
+    updateYamlCompletionData({
+      components: components ?? [],
+      triggers: triggers ?? [],
+      widgets: widgets ?? [],
+    });
+  }, [components, triggers, widgets]);
```
shiroyasha: This is a bit weird. We don't we update the data directly, instead of going through useEffect?

web_src/src/pages/workflowv2/CanvasYamlView.tsx
```diff
@@ -0,0 +1,348 @@
...
+      setEditedText(text);
+
+      if (debounceRef.current) {
+        clearTimeout(debounceRef.current);
+      }
+
+      debounceRef.current = setTimeout(() => {
+        runValidation(text);
+
+        // Only propagate changes upstream when there are no errors
```
shiroyasha: Does that mean that auto-save is not working if there is any issue?

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -4345,8 +4355,105 @@ export function WorkflowPageV2() {
...
+      canvas,
+      organizationId,
+      canvasId,
+      queryClient,
+      saveWorkflowSnapshot,
+      handleSaveWorkflow,
+      canAutoSave,
+      markUnsavedChange,
+    ],
+  );
```
shiroyasha: Extract this to a dedicated file, we are already on line ~4000.

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -4401,6 +4508,9 @@ export function WorkflowPageV2() {
           onTopViewModeChange={(mode) => {
             setIsCreateChangeRequestMode(false);
             setTopViewMode(mode);
+            if (mode !== "yaml") {
+              setYamlServerError(null);
```
shiroyasha: When you leave the YAML view it deleted server errors?

web_src/src/pages/workflowv2/yamlCanvasValidation.ts
```diff
@@ -0,0 +1,516 @@
...
+ *      "block" → background highlight spanning the full node block
+ *  - Diagnostics with `severity: "error"` block saves; "warning" does not.
+ *  - An optional `code` field (DiagnosticCode) tags fixable diagnostics so
+ *    the code-action provider in yamlQuickFixes.ts can offer quick fixes.
+ *
+ * To add a new validation rule:
+ *  1. Add the check inside `validateCanvasYaml` (or a helper it calls).
+ *  2. If the rule should be fixable, add a new DiagnosticCode and tag the
+ *     diagnostic, then register a matching quick fix in yamlQuickFixes.ts.
+ */
```
shiroyasha: Do we have to do yaml validation on the frontend? 

This means that the CLI would not be able to support this => which in turn means that AI cannot access this.

shiroyasha:
Overall, good PR 👍 

Please add at least some e2e tests to cover the major parts (good yaml, yaml errors, quickfix application, completly broken yaml). There are a lot of moving parts in this PR, it will be super easy to break it if we have no tests for it.

web_src/src/pages/workflowv2/CanvasYamlView.tsx
AleksandarCole: Fix: https://github.com/superplanehq/superplane/pull/3400/changes/88b071b9c033b47387df22fb4ae9efd3f9e4c788

web_src/src/pages/workflowv2/CanvasYamlView.tsx
AleksandarCole: Fix: https://github.com/superplanehq/superplane/pull/3400/changes/88b071b9c033b47387df22fb4ae9efd3f9e4c788

web_src/src/pages/workflowv2/CanvasYamlView.tsx
AleksandarCole: Fix: https://github.com/superplanehq/superplane/pull/3400/changes/8140ae9f2166988fd524dda9859e4e1a78bd1fee

shiroyasha: 🤔 This one is now running it on every render, could make the whole yaml page very slow. Lets be careful with this one, our canvas page is already buggy.

web_src/src/pages/workflowv2/CanvasYamlView.tsx
AleksandarCole: Auto-save is only blocked when there are error-severity diagnostics (duplicate IDs, missing required fields, wrong field types, broken YAML syntax). These would cause the backend to reject the save with a proto deserialization error anyway, so blocking client-side gives better UX.

Warning-severity diagnostics (e.g. errorMessage: "field 'url' is required", position overlaps, duplicate component names) do not block saves. Users can save partially configured canvases just like in the regular canvas editor.

web_src/src/pages/workflowv2/index.tsx
AleksandarCole: Server errors are YAML-editor-only UI state. They're displayed as Monaco editor markers (squiggly underlines and that red code block). You cannot save yaml with errors (will be rejected by server) and if you go outside the editor these changes will disapear, so these server error strings are meaningless outside the YAML editor.

Stale errors shouldn't reappear. If the user switches to Canvas view, and switches back to YAML, the YAML text is regenerated from the current canvas state. The old server error would no longer correspond to the current YAML content, so showing it would be misleading.

Btw, failed saves no longer corrupt local state. With the fix in this PR, `handleYamlChange` now only updates the local query cache after `handleSaveWorkflow` returns success. **If the backend rejects the save**, the local canvas state remains unchanged (the last known-good state). So when the user switches back to YAML, they'll see the YAML regenerated from the last successfully saved state and therefore the error condition is naturally gone.

web_src/src/pages/workflowv2/CanvasYamlView.tsx
AleksandarCole: Switched to `useMemo` is this better approach? https://github.com/superplanehq/superplane/pull/3400/changes/111b879c9be85a95dc30ee602390db3e7ba2f0b2

web_src/src/pages/workflowv2/yamlCanvasValidation.ts
AleksandarCole: @shiroyasha 
the backend's `serialization.go` is and should be the source of truth. The frontend validation (`yamlCanvasValidation.ts`) is a UX optimization for instant feedback as I lack any other way of having that atm. A follow-up task could expose a `/validate` API endpoint that the CLI (and by extent AI) and frontend both use, however, that seems like out of scope for this PR (or at least I would not mix it). 

Once we decide on the architecture and how to approach this in backend - this feature can be updated and client side validation ditched and replaced.

AleksandarCole: btw, I already had some pain caused by the lack of this validation in CLI, logged some of it: https://github.com/superplanehq/superplane/issues/3250

web_src/src/pages/workflowv2/index.tsx
AleksandarCole: Should be addressed with: https://github.com/superplanehq/superplane/pull/3400/changes/01dcd72e40d9f748054321e321f4ab357bf42442

web_src/src/pages/workflowv2/CanvasYamlView.tsx
shiroyasha: Maybe ask @forestileao to help you out.

AleksandarCole:
@shiroyasha thanks for the review. I ended up keeping the `Preview mode` only - which is a lot simpler change.

PR #3408, opened 2026-03-09
fix: add resizable dialog for configuration fields (#2807)

### LINKED ISSUE #2807: Payload / config modals: responsive size and drag-to-resize
**Problem:** The modals used to expand payload (e.g. in workflow execution) or rich text editor configuration fields are too small and don't adapt well to different screen widths. Users can't resize them, so exploring large JSON or config content is awkward.

**Current vs proposed:**
- **Current:** Fixed or narrow modal size; no user-controlled resizing; poor adaptation to screen width.
- **Proposed:** Responsive modal sizing that adapts to screen width, plus drag-to-resize (e.g. drag edge/corner) so users can widen the modal as needed.

**Use case:** Anyone inspecting payloads or editing rich text config in the UI benefits—especially on smaller or very wide screens—by being able to see more content at once and adjust the modal width.

---

**Screenshot**

Image

masicaaa:
Fixes #2807
Adds a resizable dialog for configuration fields and updates related UI components.


AleksandarCole:
@masicaaa thanks for trying this one. Please check CI - there are a lot of things failing, including build.

AleksandarCole:
@masicaaa thanks for trying this one 🙌 
The CI is still failing, we cannot merge anything that has a broken build. 

Few notes regarding UX:

### Easing effect - not responsive
Resizing seems a bit sluggish, and not responsive. I drag the mouse and things happen with delay + they don't seem to be accurate. Any chance we can improve this? Please compare to the resizing of the annotation nodes - they feel snappier. 

https://github.com/user-attachments/assets/802227a1-ca53-4d8d-ab1b-59cdced555eb

### Resizing indicators
Can we think of some better visual indicators that this is resizable? It seems hard for user to discover this with current design (those faint edge changes on hover), and also doesn't seem to match the rest of the style. Our sidebar is also resizable - maybe check that for inspiration.

AleksandarCole:
@masicaaa I did a quick review on implementation approach too and I'm not sure that creating this new  `ResizableDialogContent` component is the best way to go. 

### Nice touch - should keep
- localStorage size persistence via `storageKey` > good instinct on adding this

### Problems

1. **Full duplication of `dialog.tsx`**. Every sub-component is copy-pasted with a different `data-slot`. Two components to maintain going forward.
2. **No touch or keyboard support**. Uses `mousedown`/`mousemove`/`mouseup` only.
3. **Left/top resize causes jitter**. The dialog stays centered via `translate(-50%, -50%)`, so resizing from any edge moves both sides. (probably cause of that UX issue mentioned)

### Suggestion

We already have `react-resizable-panels` installed and the shadcn `resizable.tsx` wrapper. The standard approach is to compose resizable panels inside a Dialog, not to fork the Dialog itself. That gives us touch, keyboard, and persistence support without a new component or dependency.

Rework this to extend the existing `DialogContent` with a `resizable` prop and use what we already have.

AleksandarCole:
Closing this one as outdated. Feel free to reopen

PR #3426, opened 2026-03-11
fix: Remove Deleted Webhooks

e-todorovski-bm:


pkg/workers/webhook_provisioner.go
```diff
@@ -65,68 +73,126 @@ func (w *WebhookProvisioner) Start(ctx context.Context) {
...
+	if err != nil {
+		return err
+	}
+	if lockedWebhook == nil {
+		// Already being processed or no longer pending.
+		return nil
+	}
+
+	// Non-integration webhooks don't need external calls — mark ready directly.
+	if lockedWebhook.AppInstallationID == nil {
```
lucaspin: Can't we avoid the `pending -> provisioning -> ready` when AppInstallationID is nil, and just `pending -> ready` directly?

e-todorovski-bm: @lucaspin yes, we can do that. The code is already updated reflecting that change

PR #3428, opened 2026-03-11
feat: introduce webhook response body

e-todorovski-bm:
This PR is an extract and fix of the implementation in https://github.com/superplanehq/superplane/pull/3230

As this implementation touches a lot of files, I thought of an alternative that will be with localised impact (3 files instead of 236). The alternative was adding a Response *WebhookResponseBody field to WebhookRequestContext. Only triggers that need a custom response (like Azure Event Grid validation) would write to ctx.Response; all others would ignore it. 

I didn't choose it because it relies on a side effect: mutating a shared struct pointer, rather than an explicit return value. The interface signature wouldn't tell you that handlers can produce a response body; you'd only discover that by reading the struct definition. It's also harder to test (assert a mutation vs assert a return value) and fragile if the server forgets to check the field. The return value approach makes the contract visible and compiler-enforced, which is worth the mechanical cost of adding , nil to every existing implementation. 
That implementation would have taken the first non-empty response, so there was no risk of other components rewriting the initial body. However, that too didnt seem like a clean solution, so with all things considered, I went with this massive one.

### DIFF
--- a/pkg/integrations/dash0/create_check_rule.go
+++ b/pkg/integrations/dash0/create_check_rule.go
@@ -403,8 +403,8 @@ func (c *CreateCheckRule) HandleAction(ctx core.ActionContext) error {
 	return nil
 }
 
-func (c *CreateCheckRule) HandleWebhook(ctx core.WebhookRequestContext) (int, error) {
-	return http.StatusOK, nil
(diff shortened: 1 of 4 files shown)


e-todorovski-bm:
@lucaspin

Issue #3427, opened 2026-03-11
[GCP Cloud DNS] Expand integration


AleksandarCole:
Actions:
- Create Record
- Delete Record
- Update Record

WashingtonKK:
Working on it.

PR #3433, opened 2026-03-12
feat(gcp/cloud-dns): add gcp cloud dns integration

WashingtonKK:
Implements #3427

Expands the gcp integration by adding three actions on [CloudDNS](https://cloud.google.com/dns):

- `createRecord`
- `deleteRecord`
- `updateRecord`

### Demo

[Screencast from 2026-03-12 13-08-09.webm](https://github.com/user-attachments/assets/d89d4232-b751-47bc-8321-b02233e28b4a)

### DIFF
--- a/web_src/src/ui/CanvasPage/Header.tsx
+++ b/web_src/src/ui/CanvasPage/Header.tsx
@@ -3,8 +3,10 @@ import {
   CloudAlert,
   CloudCheck,
   CloudUpload,
+  Copy,
   Home,
   ChevronDown,
+  Download,
   LogOut,
   Palette,
   RotateCcw,
@@ -20,6 +22,7 @@ import { Link, useParams } from "react-router-dom";
 import { useEffect, useRef, useState, type ReactNode } from "react";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { DropdownMenu, DropdownMenuContent, DropdownMenuTrigger } from "@/ui/dropdownMenu";
+import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "@/ui/select";
 
 export interface BreadcrumbItem {
   label: string;
@@ -122,6 +125,7 @@ export function Header({
   const [isYamlMenuOpen, setIsYamlMenuOpen] = useState(false);
   const [isEditingMenuOpen, setIsEditingMenuOpen] = useState(false);
   const [isSaveMenuOpen, setIsSaveMenuOpen] = useState(false);
+  const [exportAction, setExportAction] = useState<string>("");
   const menuRef = useRef<HTMLDivElement | null>(null);
 
   // Get the workflow name from the workflows list if workflowId is available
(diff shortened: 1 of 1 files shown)


pkg/integrations/gcp/clouddns/poll.go
```diff
@@ -0,0 +1,61 @@
...
+
+	output := map[string]any{
+		"change": map[string]any{
+			"id":        change.ID,
+			"status":    change.Status,
+			"startTime": startTime,
+		},
+		"record": map[string]any{
+			"name": meta.RecordName,
+			"type": meta.RecordType,
```
forestileao, line 53: From cursor: pollChangeUntilDone does not short-circuit when ctx.ExecutionState.IsFinished() is already true.
In this codebase, many poll handlers guard for this (to handle at-least-once action delivery / late queued polls). Without that guard, a late poll can call Emit again and create duplicate downstream events.

pkg/integrations/gcp/clouddns/common.go
```diff
@@ -0,0 +1,255 @@
...
+// getRecordSet fetches an existing record set by name and type, or nil if not found.
+type rrsetListResponse struct {
+	Rrsets []ResourceRecordSet `json:"rrsets"`
+}
+
+func getRecordSet(ctx context.Context, client Client, projectID, managedZone, name, recordType string) (*ResourceRecordSet, error) {
+	url := fmt.Sprintf(
+		"%s/projects/%s/managedZones/%s/rrsets?name=%s&type=%s",
+		cloudDNSBaseURL, projectID, managedZone, name, recordType,
+	)
```
forestileao, line 226: name and type are interpolated directly into query strings.
This is brittle and differs from common patterns elsewhere in GCP integrations (url.Values / url.QueryEscape). Special characters in dynamic input can produce malformed or ambiguous queries.

pkg/integrations/gcp/clouddns/common.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	return &resp.Rrsets[0], nil
+}
+
+// listRecordSetsByName fetches all record sets for a given name regardless of type.
+func listRecordSetsByName(ctx context.Context, client Client, projectID, managedZone, name string) ([]ResourceRecordSet, error) {
+	url := fmt.Sprintf(
+		"%s/projects/%s/managedZones/%s/rrsets?name=%s",
+		cloudDNSBaseURL, projectID, managedZone, name,
+	)
```
forestileao, line 251: same as https://github.com/superplanehq/superplane/pull/3433/changes#r2925817534

pkg/integrations/gcp/clouddns/poll.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/superplanehq/superplane/pkg/core"
+)
+
+const (
+	pollChangeActionName = "pollChange"
+	pollInterval         = 5 * time.Second
```
forestileao: since dns takes a while to update. I would prefer updating this to 1 minute. WDYT

forestileao (commented):
Could we add test coverage for the new branches introduced here to reduce regression risk?

1. Delete without type (delete_record.go, if config.Type == "" path):
- verify we delete all rrsets returned by listRecordSetsByName
- verify output record.type is the joined list of deleted types
2. Polling action behavior (poll.go):
- one test where status is pending and we re-schedule poll
- one test where status is done and we emit final output

Current tests mainly cover the explicit-type delete path, so adding these would make the new behavior safer.

And also fix the conflicts

pkg/integrations/gcp/clouddns/poll.go
WashingtonKK: I agree with you.

forestileao:
@WashingtonKK I was wrong, then I reverted the polling time, since it is the same we use in Route53. 

It. is to verify if the change was made on gcp, not in some domain server, my bad.

PR #3434, opened 2026-03-12
fix: GitHub integration sync race condition  

alabro-bm:
Issue: When a GitHub App is created but not yet installed, the Sync worker runs on an interval and regenerates the OAuth state (since GitHubApp.ID is still 0 at that point in the original code). This causes the state stored in metadata to diverge from the state the user's browser is carrying through the OAuth flow, resulting in "invalid installation ID or state" on the setup callback.

Fix:
- Early-return in Sync when GitHubApp.ID != 0 — prevents state regeneration after app creation
- In afterAppCreation, explicitly write metadata.State = state from the redirect query param before saving — ensures the stored state always matches the one flowing through the OAuth redirect, even if Sync ran between the browser action being rendered and the user submitting the form.

### DIFF
--- a/pkg/integrations/github/github.go
+++ b/pkg/integrations/github/github.go
@@ -153,6 +153,14 @@ func (g *GitHub) Sync(ctx core.SyncContext) error {
 		return nil
 	}
 
+	//
+	// App has been created but not yet installed - preserve existing state
+	// so the pending installation callback can validate it.
+	//
+	if metadata.GitHubApp.ID != 0 {
+		return nil
+	}
+
 	state, err := crypto.Base64String(32)
 	if err != nil {
 		return fmt.Errorf("Failed to generate GitHub App state: %v", err)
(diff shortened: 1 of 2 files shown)


lucaspin:
> When a GitHub App is created but not yet installed, the Sync worker runs on an interval and regenerates the OAuth state

Not sure I understand how Sync() runs on an interval for the GH integration. I don't see `ctx.Integration.ScheduleResync()` or `ctx.Integration.ScheduleActionCall()` being used in the GitHub integration to schedule things for the request worker. Could you detail a bit more how the Sync() is running on an interval?

lucaspin:
Ok, I think I understand it now. The problem seems to be that the UI is making a request to update the integration when the browser action is clicked. Issue is on [this line](https://github.com/superplanehq/superplane/blob/main/web_src/src/pages/organization/settings/IntegrationDetails.tsx#L192), which seems to have been added as part of https://github.com/superplanehq/superplane/pull/3378.

@shiroyasha @e-todorovski-bm why did we change that? This change could potentially have broke not just GitHub, but a lot of other integrations. Also, it doesn't make sense to run a `PATCH /integrations/{id}` when the browser action button is clicked

shiroyasha:
🙈, missed that change while reviewing the code. Sucks that we have no tests for it.

shiroyasha:
Lets revert that one. I would rather have a broken teams integration than broken github one.

shiroyasha:
Assigning this one to me. I'll clean it up and add tests for GitHub.

shiroyasha:
Closing this PR in favor of this:
https://github.com/superplanehq/superplane/pull/3462

PR #3438, opened 2026-03-12
feat: add the Perplexity Search and Agent APIs as integrations

kesku:
Adds two Perplexity components:

- `perplexity.webSearch`: runs a ranked web search via the Search API (`POST /search`) and emits results with titles, URLs, snippets, and dates
- `perplexity.runAgent`: runs an AI agent via the Agent API (`POST /v1/responses`) with configurable presets, model selection, and tool flags (`web_search`, `fetch_url`), emitting the response text and citations

The integration verifies credentials on sync using a minimal Search API call, and populates the model dropdown live from `GET /v1/models`

### Components

**Web Search** — `perplexity.webSearch`
- Fields: query, max results (1–10), domain filter, recency filter (day/week/month/year)
- Emits: `perplexity.search.results`

**Run Agent** — `perplexity.runAgent`
- Fields: preset (fast-search, pro-search, deep-research, advanced-deep-research), model (live from API, visible only when no preset), input, instructions, web search toggle, fetch URL toggle
- Emits: `perplexity.agent.response` with text, citations, usage, and full raw response

### DIFF
--- a/pkg/integrations/perplexity/run_agent.go
+++ b/pkg/integrations/perplexity/run_agent.go
@@ -232,9 +232,9 @@ func (c *runAgent) Execute(ctx core.ExecutionContext) error {
 		Input: spec.Input,
 	}
 
-	if spec.Preset != "" {
+	if spec.ModelSource == "preset" {
 		req.Preset = spec.Preset
-	} else if spec.Model != "" {
+	} else {
 		req.Model = spec.Model
 	}
 


AleksandarCole:
@kesku this is great! Thanks for submitting this one ❤️ 
The runAgent component is clear - we will make some smaller UX adjustments and try to merge this soon. 

I am having a bit of a challenge in figuring out the Web search one though, I can see use leveraging this preplexity functionality in some way for our agent Canvas builder, but I'm having trouble figuring out the clear usecase for the Canvas Component. 

What would be the use-case where someone would use this one in their Canvas workflow?

Maybe we can keep the `runAgent` one in this PR just to keep it moving, and we can figure out other Perplexity components and the place for `webSearch` in the new extension issue, wdyt?

kesku:
Hey! I was thinking webSearch as a feeder into another LLM component in the pipeline once that's viable, can be a follow up

pkg/integrations/perplexity/run_agent.go
```diff
@@ -0,0 +1,306 @@
...
+			Description: "Model to use when no preset is specified",
+			TypeOptions: &configuration.TypeOptions{
+				Resource: &configuration.ResourceTypeOptions{
+					Type: "agent-model",
+				},
+			},
+			VisibilityConditions: []configuration.VisibilityCondition{
+				{Field: "preset", Values: []string{""}},
+			},
+		},
```
forestileao, line 151: According to the description `Model to use when no preset is specified`, is it mandatory to set  model when preset is not specified? If so, add `RequiredConditions` in addition to  `VisibilityConditions`. With same structure.

Don't forget to validate this requirement in Setup() (`if preset == "" && model == "" return error.new(...)`)

AleksandarCole: @forestileao good catch, I didn't even notice the `model` option since for me preset is always set unless I switch to expression mode. I will change a UX of this a bit

AleksandarCole: @forestileao can you check now?

Issue #3405, opened 2026-03-09
[Digital Ocean] Expand Integration


forestileao:
## Description

DigitalOcean integration currently supports only one action: `digitalocean.createDroplet`.

This issue tracks expanding DigitalOcean coverage with core lifecycle and networking actions using the official DigitalOcean API.

## Why

Users can provision droplets today, but cannot:
- read/delete/manipulate existing droplets
- manage snapshot and DNS lifecycle
- manage load balancer and reserved IP operations

Adding these actions brings DigitalOcean closer to parity with other infrastructure integrations already available in SuperPlane.

## Scope

1. `digitalocean.getDroplet` (Action)  
Fetch droplet details by ID.  
API: `GET /v2/droplets/{droplet_id}`  
Docs: https://docs.digitalocean.com/reference/pydo/reference/droplets/get/

2. `digitalocean.deleteDroplet` (Action)  
Delete a droplet.  
API: `DELETE /v2/droplets/{droplet_id}`  
Docs: https://docs.digitalocean.com/reference/pydo/reference/droplets/destroy/

3. `digitalocean.manageDropletPower` (Action)  
Run power operations (`power_on`, `shutdown`, `reboot`, `power_cycle`, `power_off`) and wait for completion.  
API: `POST /v2/droplets/{droplet_id}/actions`, poll action status via `GET /v2/actions/{action_id}`  
Docs: https://docs.digitalocean.com/reference/pydo/reference/droplet_actions/post/ and https://docs.digitalocean.com/reference/pydo/reference/droplet_actions/get/

4. `digitalocean.createSnapshot` (Action)  
Create a droplet snapshot and wait for action completion.  
API: `POST /v2/droplets/{droplet_id}/actions` (`type=snapshot`)  
Docs: https://docs.digitalocean.com/reference/pydo/reference/droplet_actions/post/

5. `digitalocean.deleteSnapshot` (Action)  
Delete snapshot image.  
API: `DELETE /v2/snapshots/{snapshot_id}`  
Docs: https://docs.digitalocean.com/reference/pydo/reference/snapshots/delete/

6. `digitalocean.createDNSRecord` (Action)  
Create domain DNS record.  
API: `POST /v2/domains/{domain}/records`  
Docs: https://docs.digitalocean.com/reference/pydo/reference/domains/create_record/

7. `digitalocean.deleteDNSRecord` (Action)  
Delete domain DNS record.  
[...]

PR #3439, opened 2026-03-12
feat: add DigitalOcean getDroplet, deleteDroplet and manageDropletPower actions

felixgateru:
Implements: #3405

## What changed:

Added three new DigitalOcean components :
digitalocean.getDroplet - Retrieve droplet information by ID
digitalocean.deleteDroplet - Delete a droplet
digitalocean.manageDropletPower - Manage droplet power state with 5 operations (power_on, power_off, shutdown, reboot, power_cycle)

## Why:

To provide complete lifecycle management for DigitalOcean droplets within SuperPlane workflows, enabling users to:
Automate droplet operations (creation, retrieval, deletion, power management)

### Backend:
Implemented get_droplet.go, delete_droplet.go, manage_droplet_power.go components following SuperPlane component interface
Created list_resources.go implementing ListResources() for 4 resource types

### Frontend:
Created mappers for all three components with custom metadata display

## Demo
https://youtu.be/BQT958J1OuU

### DIFF
--- a/web_src/src/pages/workflowv2/mappers/digitalocean/delete_droplet.ts
+++ b/web_src/src/pages/workflowv2/mappers/digitalocean/delete_droplet.ts
@@ -36,7 +36,7 @@ export const deleteDropletMapper: ComponentBaseMapper = {
     const details: Record<string, string> = {};
 
     if (context.execution.createdAt) {
-      details["Created At"] = new Date(context.execution.createdAt).toLocaleString();
+      details["Executed At"] = new Date(context.execution.createdAt).toLocaleString();
     }
 
     const outputs = context.execution.outputs as { default?: OutputPayload[] } | undefined;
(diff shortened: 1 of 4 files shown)


AleksandarCole:
@felixgateru looks good, but I'm unsure about few things:

### Delete droplet:
The timestamp says "Created at" which doesn't make sense. Why didn't we go with "Deleted at"? 

image

### Manage Power
The statuses in the node and in the sidebar do not match (one says "Powered off" other "Managed") also they are different color. Why did we decide on this approach?

image


I also wonder why we went with so many timestamps in details tab? 
image

felixgateru:
Hey @AleksandarCole 
I have addressed the comments as follows:
1. Standardized timestamp to `Executed At` for the details page:
image

2. Fixed the mismatch in the status tag and colour between the node and the sidebar
image

3. Removed the unnecessary timestamp in power management node details:
image

forestileao:
@felixgateru Really good job here, Looks good!

PR #3440, opened 2026-03-13
feat: add DigitalOcean createSnapshot and deleteSnapshot actions 

felixgateru:
Implements: #3405 

## What changed:

Added createSnapshot component to create snapshots of DigitalOcean droplets
Added deleteSnapshot component to delete DigitalOcean snapshots

## Why:

Enable users to create backup snapshots of droplets before risky operations
Support snapshot-based workflows for droplet migration and image creation

### Backend:
Added CreateSnapshot component with polling mechanism to track snapshot creation progress
Added DeleteSnapshot component with immediate execution (no polling required)
Added client methods: CreateDropletSnapshot(), GetAction(), GetDropletSnapshots(), DeleteSnapshot(), ListSnapshots(), GetSnapshot()

### Frontend:
Created create_snapshot.ts mapper with execution details showing snapshot info, regions, and size
Created delete_snapshot.ts mapper with execution details showing deletion status
Added SnapshotNodeMetadata and configuration interfaces in types.ts

## Demo
https://youtu.be/YMyGQyEJxPA

### DIFF
--- a/pkg/integrations/digitalocean/example_output_create_snapshot.json
+++ b/pkg/integrations/digitalocean/example_output_create_snapshot.json
@@ -3,7 +3,7 @@
     "created_at": "2026-03-13T13:35:43Z",
     "id": 220464921,
     "min_disk_size": 10,
-    "name": "omg-this-again",
+    "name": "superplane-1773328904",
     "regions": ["blr1"],
     "resource_id": "98145763",
     "resource_type": "droplet",
--- a/docs/components/DigitalOcean.mdx
+++ b/docs/components/DigitalOcean.mdx
@@ -135,7 +135,7 @@ Returns the snapshot details including:
     "created_at": "2026-03-13T13:35:43Z",
     "id": 220464921,
     "min_disk_size": 10,
-    "name": "omg-this-again",
+    "name": "superplane-1773328904",
     "regions": [
       "blr1"
     ],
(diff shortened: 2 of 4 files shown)


pkg/integrations/digitalocean/list_resources.go
```diff
@@ -100,6 +100,27 @@ func (d *DigitalOcean) ListResources(resourceType string, ctx core.ListResources
...
+		resources := make([]core.IntegrationResource, 0, len(snapshots))
+		for _, snapshot := range snapshots {
+			resources = append(resources, core.IntegrationResource{
+				Type: resourceType,
+				Name: snapshot.Name,
+				ID:   snapshot.ID.String(),
+			})
+		}
+		return resources, nil
+
```
forestileao: this function is getting very big, can we modularize it in functions calls in the case clauses?

pkg/integrations/digitalocean/example_output_create_snapshot.json
```diff
@@ -0,0 +1,14 @@
...
+    "min_disk_size": 10,
+    "name": "omg-this-again",
+    "regions": ["blr1"],
+    "resource_id": "98145763",
+    "resource_type": "droplet",
+    "size_gigabytes": 2.04
+  },
+  "timestamp": "2026-03-13T13:36:14.803060936Z",
+  "type": "digitalocean.snapshot.created"
+}
```
forestileao, line 14: Why this response has resource_type droplet and in the video the response has a lot of empty fields. That is correct?

Image

felixgateru: I have addressed this. I was using the list snapshots API response as a reference instead of the list droplet snapshots which has a different bofy. I hae addressed this in code.
Now we have:
image
 and 
 
image

PR #3457, opened 2026-03-13
fix: canvas update 500 SQL errors

forestileao:
Fixes intermittent `workflow_nodes_pkey` duplicates and `app_installation_id_fkey` failures during canvas updates, and sanitizes API error responses.

The `workflow_nodes_pkey` duplicates were caused by `remapNodeIDsForConflicts` not considering soft-deleted nodes, leading to conflicts when re-using node IDs. The `app_installation_id_fkey` failures occurred because the `integration.id` was not validated before database operations, resulting in foreign key violations if the referenced app installation was invalid or missing. Raw SQL errors were exposed due to improper error handling.

<div><a href="https://cursor.com/agents/bc-ca21af00-375a-4aa6-92e7-a3069e4a1760"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ca21af00-375a-4aa6-92e7-a3069e4a1760"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/pkg/grpc/actions/canvases/runtime_node_setup.go
+++ b/pkg/grpc/actions/canvases/runtime_node_setup.go
@@ -28,7 +28,11 @@ const (
 )
 
 func readdedNodeID(workflowID uuid.UUID, originalNodeID string) string {
-	sum := sha256.Sum256([]byte(workflowID.String() + "\x00" + originalNodeID))
+	return readdedNodeIDWithAttempt(workflowID, originalNodeID, 0)
+}
+
+func readdedNodeIDWithAttempt(workflowID uuid.UUID, originalNodeID string, attempt int) string {
+	sum := sha256.Sum256([]byte(workflowID.String() + "\x00" + originalNodeID + "\x00" + fmt.Sprintf("%d", attempt)))
 	hash := hex.EncodeToString(sum[:])[:readdedNodeIDHashSize]
 
 	available := nodeIDMaxLen - len(readdedNodeIDSuffix) - len(hash)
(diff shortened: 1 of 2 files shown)


forestileao:
@cursoragent You dont need to return an error if it is an error in the node level. You can put node.state as errored and put an error message there. Just like we already do. But please format the messages so it can be used in frontnend (dont expose the database)

forestileao:
@cursoragent update this message   "failed to update live canvas without versioning" to "failed to update live canvas"

forestileao:
if I delete a node with some id and the nreadd it with same id. it is recreated from scratch and all events are relaoded. what amkes sense.
But now every time I update this node the events on its response disappear. SHould we rename the node-id once they are deleted? to like deleted-{rest of the node id}

forestileao:
@cursoragent  if I delete a node with some id and the nreadd it with same id. it is recreated from scratch and all events are relaoded. what amkes sense.
But now every time I update this node the events on its response disappear. SHould we rename the node-id once they are deleted? to like deleted-{rest of the node id}

forestileao:
@cursoragent wHAT I AM not undestanding is that:

1. it is okay that  WHen I delete and readd I lose the events.
2. What is not okay is just updating this node-id and losing my config because it was deleted in the past

forestileao:
@cursoragent not it throws 500 error again

Replciate by test this behavor and fix

forestileao:
@cursoragent this will go back to the same problem of unassinging events, no?
SHouldnt we jsut add an ONConflict fir the behavior 2 using database conflict?

forestileao:
./start

forestileao:
Everything works now :)

Issue #3435, opened 2026-03-12
[Elastic] Base


AleksandarCole:
**Describe the request:**
Add an Elastic (Elasticsearch/Kibana) integration with two starter components: a trigger that receives Kibana alert webhooks, and an action that indexes documents into Elasticsearch.

___

**Describe your use-case:**
Platform teams use Elastic to monitor infrastructure and applications. When Elastic detects a problem, they need SuperPlane to orchestrate the response across multiple tools. They also need to write structured data back to Elastic from SuperPlane workflows for dashboards, search, and further alerting.

___

**Describe functionality:**

- **Trigger: "When Alert Fires":** Receives an event from a Kibana webhook connector when an alert rule fires. 
- **Action: "Index Document":** Writes a JSON document to a specified Elasticsearch index. Inputs: index name (string) and document body (JSON object). This is a general-purpose write action, not limited to any specific use case.

**Use cases:**

1. **Alert-driven incident response:** Kibana alert fires on error spike, SuperPlane notifies Slack, creates a Jira ticket, and indexes the incident record back to Elastic for dashboards and audit.
2. **Workflow audit logging:** Any SuperPlane canvas (triggered by GitHub, Slack, deploy events, etc.) indexes a structured log entry into Elastic at the end of its run, giving teams a centralized, searchable record of all automated actions across their stack.

WashingtonKK:
Working on this

PR #3467, opened 2026-03-16
feat: Add elastic base integration

WashingtonKK:
Implements #3435

This PR adds the new Elastic integration with two starter components:

- `When Alert Fires` trigger
- `Index Document` action

The goal is to let SuperPlane react to Kibana alerts and write structured workflow data back into Elasticsearch for search, dashboards, and further alerting.

### Demo Video

[Screencast from 2026-03-16 11-03-50.webm](https://github.com/user-attachments/assets/621e43a7-6b80-467c-b297-05ff44d1f4d7)

### DIFF
--- a/pkg/integrations/elastic/elastic.go
+++ b/pkg/integrations/elastic/elastic.go
@@ -72,11 +72,12 @@ func (e *Elastic) Configuration() []configuration.Field {
 			Description: "Full URL of your Kibana instance (e.g. https://my-cluster.kb.us-east-1.aws.found.io:9243).",
 		},
 		{
-			Name:     "authType",
-			Label:    "Auth Method",
-			Type:     configuration.FieldTypeSelect,
-			Required: true,
-			Default:  "apiKey",
+			Name:        "authType",
+			Label:       "Auth Method",
+			Type:        configuration.FieldTypeSelect,
+			Required:    true,
+			Default:     "apiKey",
+			Description: "Choose whether SuperPlane should authenticate to Elasticsearch and Kibana with an API key or a username/password.",
 			TypeOptions: &configuration.TypeOptions{
 				Select: &configuration.SelectTypeOptions{
 					Options: []configuration.FieldOption{
@@ -99,20 +100,22 @@ func (e *Elastic) Configuration() []configuration.Field {
 			},
 		},
 		{
-			Name:     "username",
-			Label:    "Username",
-			Type:     configuration.FieldTypeString,
-			Required: false,
+			Name:        "username",
+			Label:       "Username",
+			Type:        configuration.FieldTypeString,
+			Required:    false,
+			Description: "Username for basic authentication. Use an account with permission to index documents and manage Kibana connectors.",
 			VisibilityConditions: []configuration.VisibilityCondition{
 				{Field: "authType", Values: []string{"basic"}},
 			},
 		},
 		{
-			Name:      "password",
-			Label:     "Password",
-			Type:      configuration.FieldTypeString,
-			Required:  false,
-			Sensitive: true,
+			Name:        "password",
+			Label:       "Password",
+			Type:        configuration.FieldTypeString,
+			Required:    false,
+			Sensitive:   true,
+			Description: "Password for basic authentication.",
 			VisibilityConditions: []configuration.VisibilityCondition{
 				{Field: "authType", Values: []string{"basic"}},
 			},
--- a/pkg/integrations/elastic/example_data_on_alert.json
+++ b/pkg/integrations/elastic/example_data_on_alert.json
@@ -1,20 +1,17 @@
 {
   "type": "elastic.alert",
   "data": {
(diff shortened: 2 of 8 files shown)


pkg/integrations/elastic/on_alert.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	if !matchesFilters(payload, config) {
+		return http.StatusOK, nil, nil
+	}
+
+	event := map[string]any{
+		"payload":    payload,
+		"receivedAt": time.Now().UTC().Format(time.RFC3339),
+	}
```
forestileao: I think the event structure can be the payload directly. So we dont need to wrap two times with data.payload etc. Since we already have a tiemstamp too.

So I would just emit the payload directly: `ctx.Events.Emit("elastic.alert", payload)`

pkg/integrations/elastic/example_data_on_alert.json
```diff
@@ -0,0 +1,20 @@
...
+      "tags": ["infrastructure", "prod"],
+      "status": "active",
+      "severity": "critical",
+      "context": {
+        "message": "Error rate exceeded threshold: 15%",
+        "threshold": 10,
+        "value": 15
+      }
+    },
+    "receivedAt": "2026-03-12T09:00:00Z"
```
forestileao: then we need to update it here too and regen the docs

web_src/src/pages/workflowv2/mappers/elastic/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export const componentMappers: Record<string, ComponentBaseMapper> = {
+  indexDocument: indexDocumentMapper,
+};
+
+export const triggerRenderers: Record<string, TriggerRenderer> = {
+  onAlertFires: onAlertFiresTriggerRenderer,
+};
+
+export const eventStateRegistry: Record<string, EventStateRegistry> = {
+  indexDocument: buildActionStateRegistry("Document indexed"),
```
forestileao: I think that just "indexed" should be okay for state name

web_src/src/pages/workflowv2/mappers/elastic/on_alert.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (severities.length > 0) items.push({ icon: "alert-triangle", label: severities.join(", ") });
+  if (statuses.length > 0) items.push({ icon: "activity", label: statuses.join(", ") });
+
+  return items;
+}
+
+function alertTitle(payload: Record<string, any> | undefined): string {
+  if (!payload) return "Elastic alert received";
+  return payload.ruleName || payload.alertName || payload.name || payload.title || "Elastic alert received";
+}
```
forestileao, line 104: I think the title can have 

`${currentTitle} · ${spaceId}` if exists in the title

pkg/integrations/elastic/elastic.go
```diff
@@ -0,0 +1,248 @@
...
+			Sensitive: true,
+			Description: "Base64-encoded Elasticsearch API key (id:api_key format). " +
+				"Create one in Kibana → Stack Management → API Keys.",
+			VisibilityConditions: []configuration.VisibilityCondition{
+				{Field: "authType", Values: []string{"apiKey"}},
+			},
+		},
+		{
+			Name:     "username",
+			Label:    "Username",
```
forestileao: *Elastic).Configuration() has fields missing descriptions: authType, username, password.
Add Description for each field.

pkg/integrations/elastic/on_alert.go
```diff
@@ -0,0 +1,308 @@
...
+			TypeOptions: &configuration.TypeOptions{
+				Resource: &configuration.ResourceTypeOptions{
+					Type:  ResourceTypeKibanaSpace,
+					Multi: true,
+				},
+			},
+		},
+		{
+			Name:        "tags",
+			Label:       "Tags",
```
forestileao, line 104: the Configuration() uses ad-hoc filter types (FieldTypeList, FieldTypeMultiSelect) plus custom matcher logic in matchesFilters.

Let's prefer configuration.FieldTypeAnyPredicateList + configuration.MatchesAnyPredicate for equality/non-equality/regex filtering.

web_src/src/pages/workflowv2/mappers/elastic/index_document.ts
```diff
@@ -57,30 +52,54 @@ export const deleteDropletMapper: ComponentBaseMapper = {
 
 function metadataList(node: NodeInfo): MetadataItem[] {
   const metadata: MetadataItem[] = [];
-  const nodeMetadata = node.metadata as DropletNodeMetadata | undefined;
-  const configuration = node.configuration as DeleteDropletConfiguration;
+  const configuration = node.configuration as any;
```
forestileao: Lets avoid `as any` and use a proper typed interface

PR #3512, opened 2026-03-18
fix: Azure List Resources fail from stale OIDC

e-todorovski-bm:
Before this fix the various network calls from the components would fail when the main integration would have no valid OIDC. That could have happened when the server would restart, or after a timeout (e.g. a day)

Now the components are reestablishing the sync connection with Azure and providing a new OIDC to allow for the getters to work


pkg/integrations/azure/integration.go
```diff
@@ -277,6 +277,16 @@ func (a *AzureIntegration) HandleRequest(ctx core.HTTPRequestContext) {
 	ctx.Response.Write([]byte("not found"))
 }
 
+// SetOIDCProvider supplies the server-wide OIDC provider to this integration.
+// Called once at startup so that ListResources works even before the first Sync.
+func (a *AzureIntegration) SetOIDCProvider(p oidc.Provider) {
```
lucaspin: This is not OK, and looking at it, I'm not sure having oidcProvider as state in AzureIntegration is OK, either.

Why do we need that? Is there a way to implement Azure without that?

e-todorovski-bm: I tried a third approach, with setting the oidc on service registration. Please check it out.

I didnt want to include the oidc in the core, because its azure-specific behaviour. 

Im not sure how else could i re-init the oidc so its available for all components (created and new) when they have to use it to fetch things from Azure (e.g. ListResources)

lucaspin: What I mean by the comment above is that the Azure integration struct should not have a `oidcProvider` field on it at all. It should have no fields (state) in it, like all the other integrations. Having state there makes it hard to make these integration implementations extensible.

e-todorovski-bm: Oh, I get it! Makes sense. Initially it wasnt stateless because we were caching some things that were more expensive, like the creation of the provider using some disk actions (and other caches)
However, that is i guess smaller gain compared to the lenght of the network call, and the benefits of making this more extensible and in-line with the rest of the integrations.

So, within this PR not only did I remove the oidc prop, but also the rest of the state.

This of cource triggered an azure-wide refactor

pkg/grpc/organization_service.go
```diff
@@ -32,6 +32,8 @@ func NewOrganizationService(
 	baseURL string,
 	webhooksBaseURL string,
 ) *OrganizationService {
+	registry.InitializeOIDC(oidcProvider)
```
lucaspin: No need for this anymore, right?

pkg/integrations/azure/azure_test.go
```diff
@@ -235,28 +235,23 @@ func TestAzureIntegration_HandleRequest_Unknown(t *testing.T) {
...
+	ctx := &mockIntegrationContext{
+		id: "00000000-0000-0000-0000-000000000002",
+		config: map[string]string{
+			"tenantId":       "test-tenant",
+			"clientId":       "test-client",
+			"subscriptionId": "test-sub",
+		},
 	}
-
-	ctx := &mockIntegrationContext{id: testID}
```
lucaspin, line 246: Can we use https://github.com/superplanehq/superplane/blob/main/test/support/contexts/contexts.go#L87 instead of implementing another one here?

pkg/integrations/azure/integration.go
```diff
@@ -285,41 +276,52 @@ func (a *AzureIntegration) HandleRequest(ctx core.HTTPRequestContext) {
...
+		config.TenantID,
+		config.ClientID,
+		config.SubscriptionID,
+		getAssertion,
+		logrus.NewEntry(logrus.StandardLogger()),
+	)
+}
+
+// oidcProviderFromEnv creates an OIDC provider using keys and issuer from environment variables.
+func oidcProviderFromEnv() (oidc.Provider, error) {
```
lucaspin: This is also not OK. We should use the oidc.Provider provided in contexts and not reach out to environment variables to build it.

pkg/grpc/organization_service.go
e-todorovski-bm: Yea, sorry, the PR was still not ready when i pinged you, because I had implemented it on another branch. I was cleaning up this branch until now :)

pkg/integrations/azure/azure_test.go
e-todorovski-bm: Yup, done

pkg/integrations/azure/integration.go
e-todorovski-bm: @lucaspin  please check the latest implementation. Now it saves the token in DB and resyncs every 30 minutes. The components will read the DB value and use that one, which ( i hope) solves all previous problems.

pkg/integrations/azure/azure_test.go
```diff
@@ -3,61 +3,18 @@ package azure
...
-	return nil
-}
-func (m *mockIntegrationContext) ListSubscriptions() ([]core.IntegrationSubscriptionContext, error) {
-	return nil, nil
-}
-func (m *mockIntegrationContext) FindSubscription(func(core.IntegrationSubscriptionContext) bool) (core.IntegrationSubscriptionContext, error) {
-	return nil, nil
-}
-
 func TestAzureIntegration_Name(t *testing.T) {
```
lucaspin: No need to have unit tests for static things like Name(), Icon(), Configuration(), Triggers(), Actions(), ...

pkg/integrations/azure/integration.go
```diff
@@ -158,19 +146,22 @@ func (a *AzureIntegration) Sync(ctx core.SyncContext) error {
 		config.TenantID, config.SubscriptionID)
 
 	integrationID := ctx.Integration.ID().String()
+	getAssertion := func(_ context.Context) (string, error) {
+		subject := fmt.Sprintf("app-installation:%s", integrationID)
```
lucaspin, line 150: Shouldn't the subject here be `integration:%s` instead?

pkg/integrations/azure/azure_test.go
e-todorovski-bm: will remove

pkg/integrations/azure/integration.go
e-todorovski-bm: @lucaspin  this relates to the credentials in Azure - for that so far we have been using the pattern with "app-installation".

To me it doesnt make much difference between both options, but since existing Azure setup depends on this, I would leave it as it is, as - on the other hand - i dont see the gain much in renaming

PR #3542, opened 2026-03-19
feat: Send email notification component

AleksandarCole:
Adds a simple send email notification component that relies on SuperPlane email notifications. 
Unlike the SMTP>Send Email - this one doesn't require any additional setup from user.

### DIFF
--- a/docs/components/Core.mdx
+++ b/docs/components/Core.mdx
@@ -28,6 +28,7 @@ import { CardGrid, LinkCard } from "@astrojs/starlight/components";
   <LinkCard title="Merge" href="#merge" description="Merge multiple upstream inputs and forward" />
   <LinkCard title="No Operation" href="#no-operation" description="Just pass events through without any additional processing" />
   <LinkCard title="Read Memory" href="#read-memory" description="Find values from canvas memory by namespace and field matches" />
+  <LinkCard title="Send Email Notification" href="#send-email-notification" description="Send an email notification using the system email provider" />
   <LinkCard title="SSH Command" href="#ssh-command" description="Run a command on a remote host via SSH. Authenticate using an organization Secret (SSH key or password)." />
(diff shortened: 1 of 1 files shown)


pkg/components/send_email/send_email.go
```diff
@@ -0,0 +1,448 @@
...
+type SendEmail struct{}
+
+type Config struct {
+	RecipientMode string      `json:"recipientMode" mapstructure:"recipientMode"`
+	To            string      `json:"to" mapstructure:"to"`
+	Recipients    []Recipient `json:"recipients" mapstructure:"recipients"`
+	Subject       string      `json:"subject" mapstructure:"subject"`
+	Body          string      `json:"body" mapstructure:"body"`
+	URL           string      `json:"url" mapstructure:"url"`
+	URLLabel      string      `json:"urlLabel" mapstructure:"urlLabel"`
```
forestileao: Do we want to allow sending email for any email? Asking  because this is going to use the whole instance email provider.

Or should we not allow this component depending on an env variable?

pkg/components/send_email/send_email.go
```diff
@@ -0,0 +1,448 @@
...
+				}
+			}
+		}
+
+		receivers.Emails = mapKeys(emailSet)
+		receivers.Groups = mapKeys(groupSet)
+		receivers.Roles = mapKeys(roleSet)
+
+	default:
+		return receivers, fmt.Errorf("unknown recipient mode %q", config.RecipientMode)
```
forestileao: lets avoid putting large logics inside case clauses (modularizing into functions)

pkg/components/send_email/send_email.go
AleksandarCole: No we don't removed that option.

PR #3570, opened 2026-03-23
feat: Node grouping

AleksandarCole:
Introduces the option to group nodes on Canvas via multi select menu. 

https://github.com/user-attachments/assets/ed2bfa98-9257-4778-aedc-6ad43f5eca00

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -222,6 +222,20 @@ function expandWorkflowNodeDeletionIds(nodes: ComponentsNode[], seedIds: string[
   return toRemove;
 }
 
+/**
+ * Strips deleted node IDs from surviving group nodes' childNodeIds so the
+ * backend validation in serialization.go doesn't reject stale references.
+ */
+function pruneGroupChildReferences(nodes: ComponentsNode[], removedIds: Set<string>): ComponentsNode[] {
+  return nodes.map((node) => {
+    if (node.type !== "TYPE_WIDGET" || node.widget?.name !== "group") return node;
+    const childIds = collectGroupChildIds(node);
+    const prunedChildIds = childIds.filter((id) => !removedIds.has(id));
+    if (prunedChildIds.length === childIds.length) return node;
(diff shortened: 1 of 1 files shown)


shiroyasha:
This should be part of the YAML, no? 
Otherwise, agents will not be aware of it. 

cc @lucaspin

pkg/widgets/group/group.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (g *Group) Icon() string {
+	return "group"
+}
+
+func (g *Group) Color() string {
+	return "purple"
+}
+
+func (g *Group) Configuration() []configuration.Field {
```
lucaspin, line 34: I see width / height / childNodeIds being used as configuration for this in the UI, but not defined here. How is that possible?

lucaspin:
@shiroyasha yes, it should, and I think it is. Widgets are just nodes, so they are in the `spec.nodes` YAML field, so I think with the changes in this PR, they will look like this:

```yaml
spec:
  nodes:
    - id: start-node-123
      configuration: {}
    - id: ci-node-123
      configuration: {}
    - id: deploy-node-123
      configuration: {}
    - id: group-node-123
      configuration: {}
      childNodeIds:
        - start-node-123
        - ci-node-123
        - deploy-node-123
```

---

Is that a good-looking way to represent node groups in YAML? That's another question 😅

shiroyasha:
Good! I was afraid that we are adding this only in the frontend 😌 

The YAML is acceptable I guess. Weird that child nodes are not part of the configuration.
@AleksandarCole what happens if someone adds a non-existant node-id to the list?

AleksandarCole:
@shiroyasha @lucaspin I was actually in process of rewriting some parts based on the feedback that @forestileao shared in DM - initially I didn't have this `group.go`

I'm done now, maybe you can do a quick review (@lucaspin).

pkg/widgets/group/group.go
AleksandarCole: Outdated, changed a few things since then

PR #3633, opened 2026-03-25
feat: add scoped auth for agent chat sessions

lucaspin:
Currently, we have 2 issues regarding authentication / authorization in the agent builder flow:
1. UI <-> Agent - no authentication is used
2. Agent <-> SuperPlane API - requires the agent to be started with a `SUPERPLANE_API_TOKEN` environment variable available.

This pull request addresses both issues through:
1. Use of short-lived and scoped JWT tokens for SuperPlane and Agent API calls.
2. New `/api/v1/agents/*` SuperPlane API endpoint for generating a scoped JWT token for an agent chat session.

### Scoped JWT tokens

SuperPlane can issue scoped JWT tokens for a principal. A scoped JWT token looks like this:

```json
{
  "sub": "<USER_ID>",
  "aud": "superplane_api",
  "token_type": "scoped",
  "purpose": "agent-builder",
  "exp": 1774403764,
  "nbf": 1774402864,
  "iat": 1774402864,
  "org_id": "<ORG_ID>",
  "scopes": [
    "org:read",
    "integrations:read",
    "canvases:read:<CANVAS_ID>"
  ]
}
```

A scoped JWT token always refers to a principal (user or service account) in the `sub` claim.

### SuperPlane API

The SuperPlane API accepts scoped JWT tokens for authentication / authorization. When authorizing an API call, the API always checks the RBAC permissions for the underlying principal first, and after that, we verify if the scopes in the token itself allow the request to proceed. This ensures that RBAC for the principal is always respected, even for tokens that were already issued before an RBAC change.

### Agent API

The agent API requires JWT scoped tokens for authenticating requests to create agent chat sessions now. The JWT used must have a `purpose=agent-builder` claim.

### DIFF
--- a/agent/evals/evaluators.py
+++ b/agent/evals/evaluators.py
@@ -1,7 +1,12 @@
+from collections.abc import Iterator
 from dataclasses import dataclass
-from pydantic_evals.evaluators import Evaluator, EvaluatorContext, EvaluationReason
-from ai.models import CanvasAnswer, CanvasOperation
 from typing import Any
+
+from pydantic_evals.evaluators import EvaluationReason, Evaluator, EvaluatorContext
+
+from ai.models import CanvasAnswer, CanvasOperation
+
+_FORBIDDEN_DOLLAR_DATA = "$.data."
 
 @dataclass
 class WorkflowShape(Evaluator):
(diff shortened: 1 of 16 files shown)


shiroyasha:
Code looks good 👍 Two things make me uneasy:

1/ We are calling the endpoint `CreateAgentChatSession`, but this is slightly misleading as what we are actually doing is creating a token which we can use to start the session.

2/ We don't have any integration test for the token exchange. It will be easy to break this without us noticing in the pull-requests. We can deal with this later as well, when adding integration tests for the chat sidebar.

lucaspin:
@cursor push 23dfdf9321

lucaspin:
@shiroyasha 

Regarding [1]: renamed the endpoint to GenerateAgentChatToken.
Regarding [2]: we don't have any integration testing logic for this yet, so I'm not adding that as part of this pull request. We can add on subsequent PRs

Issue #3661, opened 2026-03-25
The agent should understand how to use $["node-name"] selectors


shiroyasha:
https://docs.superplane.com/concepts/expressions

PR #3686, opened 2026-03-25
feat: Steer AI Builder toward using $[node-name] statements when refering to upstream nodes

re-pixel:
### What changed

AI Builder now knows how to reference upstream nodes with $['node-name'] statements.

### How

Added a clear explanation on how to use $['node-name'] statements to the system prompt. Added an eval which passes when the agent successfully uses statements. The prompt in the eval is designed to induce the statement use without explicitly explaining how.

### Related Issues

https://github.com/superplanehq/superplane/issues/3661


shiroyasha:
Lets try to simplify it with, e.g:

```
UsesExprNodeSelector("slack.sendTextMessage", '$[".*"]')
```

PR #3688, opened 2026-03-25
fix: broken integrations flow

AleksandarCole:


pkg/integrations/github/github.go
```diff
@@ -43,7 +43,8 @@ type GitHub struct {
 }
 
 type Configuration struct {
-	Organization string `json:"organization"`
+	Organization string `mapstructure:"organization" json:"organization"`
+	ReturnPath   string `mapstructure:"returnPath" json:"returnPath"`
```
lucaspin: ReturnPath should not be here, as it has no relationship to an integration's configuration

AleksandarCole: Ok 👍 
I have removed it

PR #3691, opened 2026-03-25
feat: add digitalocean getObject component

gaga1307:
## What changed
Adds the `digitalocean.getObject` component for retrieving objects and their metadata from DigitalOcean Spaces Object Storage.

## How

**Configuration**
- Bucket — dropdown populated across all regions automatically; region is encoded in the bucket ID so users don't need to select it separately
- File Path — path to the object within the bucket
- Include Body — optionally download object content (text-based content types only)

**Output channels**
- `found` — object exists; returns bucket, file path, origin endpoint, content type, size, last modified, ETag, metadata, tags, and optional body
- `not_found` — object does not exist (404)

**Integration**
- Spaces Access Key ID and Secret Access Key added as optional fields to the DigitalOcean integration — only required for Spaces components; all other components continue to work with the API Token alone
- Integration instructions updated to explain when Spaces credentials are needed and how to generate them

## Demo
[Video](https://drive.google.com/drive/folders/1ba47om_GoyPcRsgRBz9bWN6SPeRKI5Nq?usp=sharing)


AleksandarCole:
@gaga1307 looking good. Few notes:

Output channels - check `read memory` component to see the naming style for 2 word channels (it's camelCase not snakeCase)

Run item status - as discussed live if you have multiple output channels ideally status needs to match the channels, so it should be "found" and "not found" not "fetched" imo. This gives user a clear overview at glance without having to open the sidebar.

Let's change that and in the meantime I will kick off code review. 

Switching PR from Draft to Ready so you can get feedback from Cursor BugBot

Also note that CI is failing (the EsLint job has to pass too)

AleksandarCole:
Hah. Seems like Cursor BugBot had same ideas :)

PR #3731, opened 2026-03-26
feat: persist agent messages into database

lucaspin:
In this one, we close the loop initiated by https://github.com/superplanehq/superplane/pull/3735 and https://github.com/superplanehq/superplane/pull/3737, persisting agent chat messages in a Postgres database.

### Main changes

- Set up database migrations for the agent service. Decided to go with the same tool for handling migrations as we use for the main SuperPlane service. Migrations and database dump are located at agent/db folder
- Update the HTTP stream to save the agent messages into the database while streaming
- Provide a real implementation for the internal gRPC service, backed by the data in the Postgres database
- Do not store agent messages in browser local storage anymore
- Update UI interface to allow user to see and resume all their previous agent chats, and create new agent chats

### Other changes

- Fixes `fix react-hooks/rules-of-hooks` warnings in BuildingBlocksSidebar/index.tsx, and decrease the budget for that kind of error
- New `make test.agent.unit` Makefile target for running tests for the agent Python service

### Demo

https://github.com/user-attachments/assets/7a1f3e0f-8141-4418-acad-a91604a6a61a

### DIFF
--- a/pkg/cli/commands/canvases/active.go
+++ b/pkg/cli/commands/canvases/active.go
@@ -17,8 +17,8 @@ func (c *ActiveCommand) Execute(ctx core.CommandContext) error {
 		return c.setActiveByID(ctx, ctx.Args[0])
 	}
 
-	if !ctx.Renderer.IsText() {
-		return fmt.Errorf("interactive canvas selection requires text output")
+	if !ctx.IsInteractive() || !ctx.Renderer.IsText() {
+		return (&listCommand{}).Execute(ctx)
 	}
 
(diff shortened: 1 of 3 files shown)


agent/src/ai/repl_web.py
```diff
@@ -50,6 +56,143 @@ class ReplStreamRequest(BaseModel):
     base_url: str | None = None
 
 
+class PersistedRunRecorder:
```
shiroyasha: Lets move this to a separate file. This one is getting really big.

PR #3732, opened 2026-03-26
feat: Azure Invoke Function 

alabro-bm:
Added new component for azure to run function (equivalent to aws lambda).

To execute this function app needs to be created first in azure as well as the function inside. Then with the dropdown we can choose the function we want to execute

Screenshot 2026-03-26 at 15 05 59
Screenshot 2026-03-26 at 15 05 50


shiroyasha:
Closing this PR.

Unfortunately, we didn't review it on time.
When we restart this effort, we should look into this one for inspiration.

PR #3738, opened 2026-03-26
fix: example outputs that were not wrapped

forestileao:
Fix example outputs that were not wrapped by our default wrapping 
```json
{
  "data": {...},
  "timestamp": "some-timestamp",
  "type": "component.type"
}

```


forestileao:
@cursoragent fix the issues and remove the json if they are not used for anything.

use make gen.component.docs after that

PR #3764, opened 2026-03-27
feat: okta sign in

e-todorovski-bm:
Introduces Enterprise SSO login using OKTA

IMPORTANT NOTES:
- The CI wasnt passing until i've committed the changes for the db file (the one regenerated via db.migrate). This dropped some keys. Not sure if relevant and should be reverted.
- There is a flag to forceAuthn. This is because on logout we are clearing only our session, not OKTA's. To clear OKTA we have to implement SLO (single logout), which is heavy both on dev, and on users side (they have to setup many more things in order to make it work), and IMO it is not needed at all. forceAuthn makes the user enter credentials every time they login, which has the same effect as logging out of OKTA. Security wise there is no change to abuse this (as far as i could tell, and it is a standard way of doing things)
- Logging it with your SSO account logs you only to a given org. Even if there are other orgs associated to the same email, or even with the same credentials, you cannot see them, unless you logout and sign in for that org. Examples in screenshots
- CI fails on dependency check for a (IMO) safe dependency which is transient of the SAML required libraries

Screenshot 2026-03-31 at 15 59 00

Screenshot 2026-03-31 at 15 58 41

Screenshot 2026-03-31 at 15 59 49
Screenshot 2026-03-31 at 15 59 14
Screenshot 2026-03-31 at 15 59 39


shiroyasha:
Closing this PR.

Unfortunately, we didn't review it on time.
When we restart this effort, we should look into this one for inspiration.

PR #3768, opened 2026-03-27
chore: rename api version to v1alpha

forestileao:
### Update

Rename API version to v1alpha instead of v1, since it still may have changes.

Mandatory to rollback to v1 after SuperPlane 1.0


shiroyasha:
I have ptsd with v1alpha in the API name on Semaphore.
IMHO, I would leave it at /v1.

forestileao:
@shiroyasha Closing it. Changing to v1alpha would break webhooks

PR #3824, opened 2026-03-30
[DRAFT] feat: Add blob storage foundation 

0vertake:
- Define BlobStorage interface with in-memory and filesystem backends
- Add blob metadata model and DB migration
- Add StoreBlob, ListBlobs, DescribeBlob, DeleteBlob API endpoints
- Wire blob authorization rules
- Expose BlobContext to component execution context
- Add blob UI panel across org, canvas, node, and execution scopes
- Regenerate proto, OpenAPI spec, and TypeScript/Go SDK clients


shiroyasha:
Closing. Stale PR.

PR #3853, opened 2026-03-31
fix: improve error message for empty canvas creation response in CLI

Vukotije:
This pull request makes a minor update to the error message shown when the server returns an empty response after attempting to create a canvas. The new message is clearer and more concise. 

* Updated the error message in `validateAndPrintCreateResponse` to "failed to create canvas: the server returned an empty response" for improved clarity.


shiroyasha:
Looks good 👍 But please add tests :)

Vukotije:
@shiroyasha i already put up the PR with the tests for this, prior to this fix #3851

PR #3866, opened 2026-03-31
fix: Add PagerDuty pagination for ListServices, ListUsers, ListEscalationPolicies, and ListIncidents methods

### LINKED ISSUE #3884: [PagerDuty] Drop-down lists do not contain all the values
### Description

The drop-down lists used to select services, users, escalation policies, and/or incidents do not contain all possible values. This is because the PagerDuty API returns the values with pagination (by default, 25 items per page).

It should be possible to manage the pagination to retrieve all values.

NickDub:
## Summary

Fixes #3884

The ListServices, ListUsers, ListEscalationPolicies, and ListIncidents methods return only the first 25 values (default number of items).

## Changes

This PR implements the pagination for ListServices, ListUsers, ListEscalationPolicies, and ListIncidents methods for PagerDuty Integration to retrieve all the values.

### DIFF
--- a/pkg/integrations/pagerduty/client.go
+++ b/pkg/integrations/pagerduty/client.go
@@ -99,6 +99,48 @@ func (c *Client) execRequest(method, url string, body io.Reader) ([]byte, error)
 	return responseBody, nil
 }
 
+// paginatedResponse wraps a paginated API response.
+type paginatedResponse[T any] struct {
+	Items []T
+	More  bool
+}
+
+// paginatedList fetches all items from a paginated PagerDuty API endpoint.
+// urlBuilder receives limit and offset and returns the full URL for that page.
+// parseResponse parses the response body and returns the items and whether there are more pages.
+func paginatedList[T any](
+	c *Client,
+	urlBuilder func(limit, offset int) string,
+	parseResponse func([]byte) (paginatedResponse[T], error),
+) ([]T, error) {
+	var allItems []T
+	offset := 0
+	limit := 100
+
+	for {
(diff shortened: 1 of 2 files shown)


pkg/integrations/pagerduty/client.go
```diff
@@ -633,28 +672,25 @@ type PriorityRef struct {
 // ListIncidents retrieves incidents from PagerDuty filtered by status and optionally by service IDs.
 // By default, it returns open incidents (triggered and acknowledged).
 func (c *Client) ListIncidents(serviceIDs []string) ([]Incident, error) {
```
shiroyasha, line 674: Paginating this one is problematic, no?

We could end up with tens of thousand of incidents in the output.

NickDub: That's true, but the [documentation](https://docs.superplane.com/components/pagerduty/#list-incidents) says, “Query PagerDuty to get a list of **all open incidents** (triggered and acknowledged).”
So we can't just arbitrarily set some aside.

What do you suggest?

shiroyasha: Hard question. 
Here are some thoughts:

1/ Would be dangerous to paginate over a large list.
2/ Even if we set a an arbitrary limit, like max=100 that will still drop some.
3/ The best solution would be to have a search instead of fetch all.

Not sure if our plumbing allows us to wire up a search instead of a list.
This is a similar problem that I've faced the other day with AWS https://github.com/superplanehq/superplane/issues/5858.

cc @forestileao @lucaspin you have more experience with this part of the code.

PR #3878, opened 2026-03-31
chore: Set up policy bot

shiroyasha:


shiroyasha:
Realized too late that I need to self-host it :'(

Issue #3674, opened 2026-03-25
SSH component: Unable to select 'None' as passphrase choice


markoa:
Clicking on the option has no effect:
https://github.com/user-attachments/assets/5251ce31-5dfb-40e0-8e63-139355b532a2

shiroyasha:
Clicking on None returns it back to "empty state".
It is more of a copy issue than a bug, e.g. the empty state should say "No passphrase".

PR #3889, opened 2026-04-01
fix: normalize optional secret key none state

forestileao:
## Summary

Fix optional `secret-key` field behavior so the empty and cleared states are consistent.

## Changes

- Show `None` for optional secret-key fields when no credential is selected.
- Keep `None` as the visible selected value after choosing the clear option.
- Render `None` as an actual select value instead of placeholder text, so it does not appear with placeholder opacity.
- Keep secret-key selection parsing and option rendering refactored to stay within the UI lint budget.

Closes: https://github.com/superplanehq/superplane/issues/3674

### DIFF
--- a/pkg/cli/commands/canvases/change_requests.go
+++ b/pkg/cli/commands/canvases/change_requests.go
@@ -80,6 +80,11 @@ func (c *changeRequestListCommand) Execute(ctx core.CommandContext) error {
 	}
 
 	return ctx.Renderer.RenderText(func(stdout io.Writer) error {
+		if len(changeRequests) == 0 {
+			_, err := fmt.Fprintln(stdout, "No change requests found.")
+			return err
+		}
+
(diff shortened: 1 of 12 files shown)


web_src/src/ui/configurationFieldRenderer/index.tsx
```diff
@@ -418,7 +426,7 @@ export const ConfigurationFieldRenderer = ({
             value={value as SecretKeyRefValue}
             onChange={(v) => onChange(v)}
             organizationId={organizationId ?? domainId}
-            placeholder={field.placeholder ?? "Select credential"}
+            placeholder={getSecretKeyPlaceholder(field, isRequired)}
```
shiroyasha: Would be clearer if this logic would be inside of the `SecretKeyFieldRenderer`.
Pass the field and isRequired, and let the component decide how to render itself.

forestileao: makes sense, addressed

PR #3890, opened 2026-04-01
fix: ListCanvasEvents should not return full execution information

lucaspin:
Right now ListCanvasEvents loads one page of root events, then SerializeCanvasEventsWithExecutions serializes each event separately, and each call into SerializeNodeExecutions, which fires 4 more queries for root events, input events, output events, and users. That makes this endpoint effectively N+1 at the page level. A 50-event page can easily turn into hundreds of queries.

This is obviously a bad pattern. Since exposing full execution information is expensive, when embedding executions into other messages, we should use the minimal information necessary. Here, we introduce a CanvasNodeExecutionRef message, which is included in the CanvasEventWithExecutions one.

### Where is this used?

This is used in the runs console, which only displays some information about the executions:

Screenshot 2026-04-01 at 09 13 52

Since we need the full execution details to render the execution list, when a user clicks on a root event, we load the executions through ListEventExecutions, lazily and scope-loading the more expensive information.

### DIFF
--- a/web_src/src/pages/workflowv2/CanvasRunsView.tsx
+++ b/web_src/src/pages/workflowv2/CanvasRunsView.tsx
@@ -322,7 +322,7 @@ export function RunRow({
   const executionRefs = useMemo(() => event.executions || [], [event.executions]);
   const executionDetailsQuery = useQuery({
     ...eventExecutionsQueryOptions(canvasId, eventId),
-    enabled
+    enabled,
   });
 
   const detailedExecutions = useMemo(
--- a/web_src/src/pages/workflowv2/utils.ts
+++ b/web_src/src/pages/workflowv2/utils.ts
@@ -393,7 +393,7 @@ export function buildRunItemFromExecutionRef(options: {
   timestampOverride?: string;
 }): LogRunItem {
   const state = getRunItemStateFromRef(options.execution);
-  return buildRunItemFromExecutionBase({...options, resolvedState: state});
+  return buildRunItemFromExecutionBase({ ...options, resolvedState: state });
 }
 
 export function buildRunEntryFromEvent(options: {


web_src/src/pages/workflowv2/CanvasRunsView.tsx
```diff
@@ -297,9 +313,17 @@ export function RunRow({
     triggerEvent?: SidebarEvent;
   }) => void;
 }) {
-  const executions = useMemo(() => event.executions || [], [event.executions]);
   const triggerNode = useMemo(() => nodes.find((n) => n.id === event.nodeId), [nodes, event.nodeId]);
   const triggerSidebarEvent = useMemo(() => buildTriggerSidebarEvent(event, triggerNode), [event, triggerNode]);
+  const executionRefs = useMemo(() => event.executions || [], [event.executions]);
+  const executionDetailsQuery = useQuery({
+    ...eventExecutionsQueryOptions(event.canvasId || "", event.id || ""),
```
shiroyasha: It's hard to read lines where multiple logical things are done.
Lets create variables for this:

```
const canvasId = event.canvasId || "";
const executionId = event.id || "";
const enabled = isExpanded && !!event.canvasId && !!event.id;

...

eventExecutionsQueryOptions(canvasId, id)
```

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -5926,10 +5926,14 @@ function useExecutionChainData(workflowId: string, queryClient: QueryClient, wor
...
-        }
-        const data = await queryClient.fetchQuery(queryOptions);
+        const data = await queryClient.fetchQuery(
+          forceReload
+            ? {
+                ...queryOptions,
+                staleTime: 0,
+              }
+            : queryOptions,
+        );
```
shiroyasha: Same here, function call + if + object creation in one flow. Lets break it:

```
const options = forceReload ? withZeroStaleTime(queryOptions) : queryOptions;
const data = await queryClient.fetchQuery(options);
```

web_src/src/pages/workflowv2/utils.ts
```diff
@@ -262,8 +263,17 @@ export function mapExecutionStateToLogType(
...
-export function buildRunItemFromExecution(options: {
-  execution: CanvasesCanvasNodeExecution;
+function getRunItemFallbackState(execution: CanvasesCanvasNodeExecutionRef): string {
+  if (execution.state === "STATE_PENDING") return "queued";
+  if (execution.state === "STATE_STARTED") return "running";
+  if (execution.result === "RESULT_FAILED") return "error";
+  if (execution.result === "RESULT_CANCELLED") return "cancelled";
+  if (execution.result === "RESULT_PASSED") return "success";
+  return "unknown";
+}
```
shiroyasha, line 272: Why is this called Fallback State? It looks like it is always used.

Also this should be a map, instead of ifs:

```
const stateNames = {
  "STATE_PENDING": "queued",
  "STATE_STARTED": "running"
  ...
};

function getStateName(execution) {
  return stateNames[execution.state] || "unknown";
}
```

web_src/src/pages/workflowv2/utils.ts
```diff
@@ -274,14 +284,15 @@ export function buildRunItemFromExecution(options: {
...
+  const { execution, nodes, onNodeSelect, timestampOverride, resolvedState, timestampFallback } = options;
   const { onExecutionSelect, event } = options;
   const componentNode = nodes.find((node) => node.id === execution.nodeId);
-  const componentName = componentNode?.component?.name || "";
-  const stateResolver = getState(componentName);
-  const state = stateResolver(buildExecutionInfo(execution));
-  const executionState = execution.resultReason === "RESULT_REASON_ERROR_RESOLVED" ? "error" : state || "unknown";
+  const fallbackState = getRunItemFallbackState(execution);
+  const executionState =
+    execution.resultReason === "RESULT_REASON_ERROR_RESOLVED" ? "error" : resolvedState || fallbackState;
```
shiroyasha: This should just be one function `getState(execution)`. Hard to reason about it here where it is mixed with 10 other things.

web_src/src/pages/workflowv2/utils.ts
```diff
@@ -262,8 +263,17 @@ export function mapExecutionStateToLogType(
   return state === "error" ? "error" : "success";
 }
 
-export function buildRunItemFromExecution(options: {
-  execution: CanvasesCanvasNodeExecution;
+function getRunItemFallbackState(execution: CanvasesCanvasNodeExecutionRef): string {
+  if (execution.state === "STATE_PENDING") return "queued";
```
forestileao: I think even for STATE_PENDING we are using running state.
queued state we use for queue items as far as I know

web_src/src/pages/workflowv2/utils.ts
lucaspin: Cannot be a map because both state and result are used to map it. I will refactor these parts in a subsequent pull request. I didn't want to dirty this pull request with TypeScript refactors, and wanted to focus on the API change

PR #3892, opened 2026-04-01
chore: Fix no constant binary expressions

shiroyasha:


web_src/src/ui/configurationFieldRenderer/BooleanFieldRenderer.tsx
```diff
@@ -1,13 +1,49 @@
...
+  value: FieldRendererProps["value"],
+  defaultValue: string | undefined,
+): boolean {
+  // if the value is a boolean, return it
+  if (typeof value === "boolean") {
+    return value;
+  }
+
+  // if the value is a string, parse it as a boolean
+  if (typeof value === "string") {
```
shiroyasha, line 27: Previously things like `"poop"` would return false silently.
Now, we return only true/false if the string is "true"/"false".

If there is anything else here like "poop", we still return `false`, but log a sentry warning so we know that something suspicious is going on.

PR #3899, opened 2026-04-01
fix: harden canvas node rendering and workflow mapper preparation

forestileao:
## Summary
- make workflowv2 node preparation resilient to mapper, additional-data, and custom-field failures
- isolate canvas node render failures so a single broken node no longer breaks the whole canvas
- extract workflowv2 node-preparation and canvas block logic into smaller utilities/modules

## What Changed
- added safe wrappers for additional-data builders and custom-field renderers
- normalized malformed mapper output before it reaches `ComponentBase` and `Trigger`
- moved trigger, component, composite, annotation, and group node preparation into dedicated workflowv2 lib files
- added per-node error boundaries in `CanvasPage`
- replaced the monolithic `Block.tsx` with a `Block/` module split into `index`, `types`, `content`, `data`, and `handles`
- made `Block` fall back safely for malformed or unknown node data instead of throwing
- hardened `componentBase` against invalid arrays, invalid strings, and custom field render errors
- added regression tests for safe mappers, workflow node preparation, block fallback rendering, and canvas node isolation

image

### DIFF
--- a/web_src/src/lib/canvas-node-fallback.ts
+++ b/web_src/src/lib/canvas-node-fallback.ts
@@ -1,0 +1,1 @@
+export const CANVAS_NODE_FALLBACK_MESSAGE = "Can't display";
--- a/web_src/src/ui/CanvasPage/index.tsx
+++ b/web_src/src/ui/CanvasPage/index.tsx
@@ -61,6 +61,7 @@ import {
 } from "@/api-client";
 import { buildSidebarComponentDocsPayload } from "@/lib/componentDocsUrl";
 import { parseDefaultValues } from "@/lib/components";
+import { CANVAS_NODE_FALLBACK_MESSAGE } from "@/lib/canvas-node-fallback";
 import { getActiveNoteId, restoreActiveNoteFocus } from "@/ui/annotationComponent/noteFocus";
 import {
   AiCanvasOperation,
@@ -456,7 +457,7 @@ function createNodeRenderFallbackData(data: BlockData): BlockData {
     group: undefined,
     renderFallback: {
       source: "node-render",
-      message: "Can't display",
+      message: CANVAS_NODE_FALLBACK_MESSAGE,
     },
   };
 }
--- a/web_src/src/lib/records.ts
+++ b/web_src/src/lib/records.ts
@@ -1,0 +1,5 @@
+export type UnknownRecord = Record<string, unknown>;
+
+export function isRecord(value: unknown): value is UnknownRecord {
+  return typeof value === "object" && value !== null;
+}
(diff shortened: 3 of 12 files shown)


web_src/src/pages/workflowv2/lib/canvas-annotation-node.ts
lucaspin, line 1: Files inside of workflowv2/lib/* are using kebab-case, let's not break the pattern

web_src/src/pages/workflowv2/lib/canvasNodeFallback.ts
```diff
@@ -0,0 +1,87 @@
...
+}): CanvasNode {
+  return {
+    id: node.id!,
+    position: { x: node.position?.x || 0, y: node.position?.y || 0 },
+    data: {
+      type: "trigger",
+      label: displayLabel,
+      state: "pending" as const,
+      outputChannels: ["default"],
+      _triggerName: node.trigger?.name,
```
lucaspin: _triggerName?

forestileao: Not a good pattern. Let me remove the _componentName too

web_src/src/ui/CanvasPage/index.tsx
```diff
@@ -378,57 +391,228 @@ const EDGE_STYLE = {
...
+      scope.setExtra("nodeLabel", nodeLabel);
+      scope.setExtra("componentStack", errorInfo.componentStack);
+      Sentry.captureException(error);
+    });
+  }
+
+  componentDidUpdate(prevProps: CanvasNodeErrorBoundaryProps) {
+    if (this.state.hasError && (prevProps.nodeId !== this.props.nodeId || prevProps.nodeData !== this.props.nodeData)) {
+      this.setState({ hasError: false });
+    }
```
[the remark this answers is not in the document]
forestileao: true, got this one now

web_src/src/lib/canvas-node-fallback.ts
```diff
@@ -0,0 +1 @@
+export const CANVAS_NODE_FALLBACK_MESSAGE = "Can't display";
```
shiroyasha, line 1: Weird that we have a dedicated file for this and not for other files. Why?

shiroyasha:
IMHO, there are too many changes for this one pull-request to be able to provide high-quality feedback. 

It is ok now, the code generally looks ok, but looking at this list, every one of them would be self-sufficient for its own PR:

- added safe wrappers for additional-data builders and custom-field renderers
- normalized malformed mapper output before it reaches ComponentBase and Trigger
- moved trigger, component, composite, annotation, and group node preparation into dedicated workflowv2 lib files
- added per-node error boundaries in CanvasPage
- replaced the monolithic Block.tsx with a Block/ module split into index, types, content, data, and handles
- made Block fall back safely for malformed or unknown node data instead of throwing
- hardened componentBase against invalid arrays, invalid strings, and custom field render errors
- added regression tests for safe mappers, workflow node preparation, block fallback rendering, and canvas node isolation

forestileao:
@shiroyasha I don;t think we need that level of granularity because some changes depend on other's. But i am going to split this PR in more.
Closing by now

PR #3900, opened 2026-04-01
feat: drain active SSE streams before agent shutdown

0vertake:
## Summary
- Track in-flight SSE streams with ActiveStreamTracker so the agent waits for them to finish on SIGTERM instead of killing them mid-response
- Reject new requests with 503 during shutdown
- Set up timeout layering: drain (300s) < uvicorn (310s) < Docker/K8s (330s), each independently configurable
- Centralize all agent environment variables into a single ai/config.py module with typed parse helpers

## How it works
1. K8s sends SIGTERM — no new traffic is routed to the pod
2. FastAPI lifespan teardown calls begin_shutdown() (new requests get 503) then wait_for_drain() (blocks until active streams finish or DRAIN_TIMEOUT expires)
3. Once drained, gRPC server and session store are cleaned up and uvicorn exits

### DIFF
--- a/db/migrations/20260402003713_alter_casbin_schema.up.sql
+++ b/db/migrations/20260402003713_alter_casbin_schema.up.sql
@@ -1,0 +1,9 @@
+BEGIN;
+
+ALTER TABLE casbin_rule
+  ALTER COLUMN ptype DROP NOT NULL;
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_casbin_rule_unique
+    ON public.casbin_rule USING btree (ptype, v0, v1, v2, v3, v4, v5);
+
+COMMIT;
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -207,7 +207,7 @@ CREATE TABLE public.canvas_memories (
 
 CREATE TABLE public.casbin_rule (
     id integer NOT NULL,
-    ptype character varying(100) NOT NULL,
+    ptype character varying(100),
     v0 character varying(100),
     v1 character varying(100),
     v2 character varying(100),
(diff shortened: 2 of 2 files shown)


agent/src/ai/repl_web.py
```diff
@@ -39,6 +39,101 @@
...
+    raw = os.getenv("DRAIN_TIMEOUT", "").strip()
+    if not raw:
+        return _DEFAULT_DRAIN_TIMEOUT
+    try:
+        return max(float(raw), 0.0)
+    except ValueError:
+        return _DEFAULT_DRAIN_TIMEOUT
+
+
+class ActiveStreamTracker:
```
lucaspin: This should be in a separate file

0vertake: This is implemented now

agent/src/ai/stream_tracker.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def _resolve_drain_timeout() -> float:
+    raw = os.getenv("DRAIN_TIMEOUT", "").strip()
+    if not raw:
+        return _DEFAULT_DRAIN_TIMEOUT
+    try:
+        return max(float(raw), 0.0)
+    except ValueError:
+        return _DEFAULT_DRAIN_TIMEOUT
```
shiroyasha: Lets introduce a global config package that contains all of our external configuration. It will be 10x easier to track the configuration options that the app uses.

e.g

``` python
# agent/src/config.py

import os
from dataclasses import dataclass

class Config:
    drain_timeout: int

    def __init__(self):
      self.drain_timeout = self.parse_int("DRAIN_TIEOUT", min=0, max=1000, default=310)

    def parse_int(env_name, min, max, default):
      raw = os.getenv(env_name, "").strip()

      # ... rest of the code ...

    
config = Config()
```

then in stream tracker:

```
import config from config

config.drain_timeout # use it here
```

0vertake: This is implemented now and added into the PR description.

PR #3906, opened 2026-04-02
fix: update db to correct casbin schema

forestileao:
This PR updates the DB to correct schema, since casbin already tries to turn into this state.

If the DB state is already correct, the migration will not affect  it and the query will not generate any error.


shiroyasha:
The changes look reasonable. I can't vouch that it is correct, but I trust you 101%.

Issue #3901, opened 2026-04-01
Groups: First group loses config on update


AleksandarCole:
**Description:**  
When updating a canvas via `superplane canvases update`, the first `group` widget in the node list loses its `configuration` block on the server response, while other `group` widgets keep theirs.

**Steps to reproduce:**  
1. Create a canvas YAML with 4 `group` widgets, each with `configuration` containing `label`, `color`, and `childNodeIds`.  
2. Run `superplane canvases update -f canvas.yaml`.  
3. Run `superplane canvases get <Canvas ID> -o yaml`.  
4. Check nodes in output YAML.

**Expected / Actual:**  
Expected: All `group` widgets retain `configuration` after update, regardless of node order.  
Actual: The first `group` widget in the list has `configuration` silently dropped (no `label`, `color`, or `childNodeIds`), with no CLI warning or error.

**Notes:**  
Tried changing color values, removing trigger/filter nodes from `childNodeIds`, renaming the node ID, and reordering nodes. The first processed group still loses configuration.

PR #3911, opened 2026-04-02
fix: CLI group loses config on update

Vukotije:
#3901

- Traced the full flow end to end — CLI YAML parsing, gRPC gateway, server processing, DB round-trip, response serialization. No code anywhere explicitly treats the first node differently from the rest.
- Found that structpb.NewStruct errors are silently discarded in NodesToProto (common.go:725), which means if that conversion ever fails for a node, its configuration just vanishes with no log or error. Added logging for that.
- Wrote a reproduction test that simulates every layer (YAML → protojson → ProtoToNodes → JSON DB roundtrip → NodesToProto → protojson response). All 4 groups keep their config. Also tried it live against the server with a test canvas — couldn't reproduce either.

So the conversion/serialization code itself is fine. The bug is most likely data-specific, something in the reporter's actual YAML that Im not testing with, since I tested on the versioned and non-versioned update path.

The logging change may is still be worth merging, so the next time it happens we'll actually see what failed instead of silent data loss.

### DIFF
--- a/pkg/grpc/actions/common_test.go
+++ b/pkg/grpc/actions/common_test.go
@@ -1,12 +1,18 @@
 package actions
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/encoding/protojson"
+	"gopkg.in/yaml.v3"
 
 	"github.com/superplanehq/superplane/pkg/configuration"
+	"github.com/superplanehq/superplane/pkg/models"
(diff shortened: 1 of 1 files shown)


forestileao:
@Vukotije I think we can close this PR if the described bug is not happening. That maybe is outdated, you can record a short video to prove or show the result of the steps described in the issue and comment in the referencing task

Vukotije:
@forestileao Here is the video, could not reproduce. All group widget configurations preserved after canvas update


https://github.com/user-attachments/assets/00f87a6d-9db5-41e5-9215-96eb78e7dd91

PR #3912, opened 2026-04-02
chore: Separate dev and test rabbitmq to stabilize E2E tests

shiroyasha:


Makefile
```diff
@@ -113,6 +115,8 @@ dev.setup:
 	@touch agent/.env
 	$(COMPOSE) build
 	$(COMPOSE) pull
+	$(COMPOSE) up -d rabbitmq
```
lucaspin: Why do we need to start rabbit on setup?

shiroyasha: Good point. I'll look into alternatives.

shiroyasha: I'm attempting to do it with just rabbitmq configuration. This would be faster to boot in CI, and would make the test.setup and dev.setup side-effect free.

Waiting for CI result...

db/structure.sql
```diff
@@ -37,21 +37,6 @@ SET default_tablespace = '';
 
 SET default_table_access_method = heap;
 
---
--- Name: account_magic_codes; Type: TABLE; Schema: public; Owner: -
---
-
-CREATE TABLE public.account_magic_codes (
```
lucaspin, line 44: Why did this change?

PR #3913, opened 2026-04-02
chore: run agent service unit tests in CI

0vertake:
## Summary
- Add a dedicated Semaphore CI block that runs agent (Python) unit tests via `make test.agent.unit`
- Add a lightweight `test.agent.setup` Makefile target that only builds the `app` and `agent` containers and sets up the `agents_test` database, avoiding unnecessary Go module downloads and superplane_test DB setup

## Test plan
- [x] Ran `make test.agent.setup && make test.agent.unit` locally, all tests pass

### DIFF
--- a/web_src/src/pages/organization/settings/Profile.tsx
+++ b/web_src/src/pages/organization/settings/Profile.tsx
@@ -13,6 +13,7 @@ import { LoadingButton } from "@/components/ui/loading-button";
 import { withOrganizationHeader } from "../../../lib/withOrganizationHeader";
 import { useOrganizationId } from "../../../hooks/useOrganizationId";
 import { meKeys, useMe } from "@/hooks/useMe";
+import { showErrorToast, showSuccessToast } from "@/lib/toast.ts";
 
 export function Profile() {
   usePageTitle(["Profile"]);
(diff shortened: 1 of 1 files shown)


shiroyasha:
Good start! But I want to see the tests in this tab:
https://superplanehq.semaphoreci.com/jobs/72838568-d023-48a8-bdbe-98f70b712878/summary

0vertake:
I also noticed that ruff and mypy are configured in agent/pyproject.toml, but there are no Makefile targets. This PR only adds unit tests, since the format and lint checks probably aren't being enforced anywhere.

lucaspin:
@0vertake good point. Let's add those too. We should also probably add these Makefile targets in the [agent/Makefile](https://github.com/superplanehq/superplane/blob/main/agent/Makefile) Makefile, and not on the root one

0vertake:
@lucaspin Here is the current state:
- ruff format: 25 files needed reformatting that were fixed by running ruff format
- ruff check (lint): 120 errors across the codebase, 22 were fixable with `ruff check --fix`, the remaining 47 were line length and import ordering issues I fixed manually
- mypy (typecheck): 83 errors across 18 files that include missing type annotations, generated protobuf stubs missing attributes, type mismatches

If I add these fixes to this PR, it would maybe become too bloated. I can do them in this PR or perhaps create a new issue followed by a PR fixing these format/lint/type checks and including it in the CI.

lucaspin:
> If I add these fixes to this PR, it would maybe become too bloated

@0vertake good intuition. Let's do that in a another PR

PR #3916, opened 2026-04-02
fix: harden mappers with safe wrappers

forestileao:
## Summary
- Add safe wrapper functions (`createSafeAdditionalDataBuilder`, `createSafeCustomFieldRenderer`) and normalization (`normalizeComponentBaseProps`, `normalizeTriggerProps`) so all mapper output is sanitized before reaching the canvas
- Extract `findRegisteredComponentMapper` private helper to deduplicate component lookup logic
- Add shared `@/lib/records` type guard utility and `@/lib/canvas-node-fallback` fallback constant

## Context
Split from #3899. This is PR 1/3 — adds the error-handling foundation at the data layer. PRs 2 and 3 depend on this.

image

### DIFF
--- a/pkg/cli/commands/canvases/update.go
+++ b/pkg/cli/commands/canvases/update.go
@@ -17,6 +17,98 @@ type updateCommand struct {
 	autoLayoutNodes *[]string
 }
 
+type versioningPlan struct {
+	effectiveEnabled      bool
+	enableAfterSpecUpdate bool
+}
+
+func updateCanvasVersioningEnabled(ctx core.CommandContext, canvasID string, enabled bool) error {
+	body := openapi_client.CanvasesUpdateCanvasBody{}
+	body.SetVersioningEnabled(enabled)
+
+	_, _, err := ctx.API.CanvasAPI.
+		CanvasesUpdateCanvas(ctx.Context, canvasID).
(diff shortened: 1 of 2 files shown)


web_src/src/pages/workflowv2/mappers/safeMappers.ts
```diff
@@ -1,6 +1,291 @@
...
+
+function asBoolean(value: unknown): boolean | undefined {
+  return typeof value === "boolean" ? value : undefined;
+}
+
+function getFallbackComponentTitle(context: ComponentBaseContext): string {
+  return context.node?.name || context.componentDefinition?.label || context.componentDefinition?.name || "Component";
+}
+
+function getFallbackTriggerTitle(context: TriggerRendererContext): string {
```
shiroyasha: The function name is not correct, it suggest that this will return only the fallback, while in reality it only returns the fallback if various other options are not satisfied.

```
safemapper.getTriggerTitle(...)
safemapper.getComponentTitle(...)
```

would be good and more direct, even if they don't have fallback in their names.

shiroyasha (dismissed):

forestileao:
test

forestileao:
👍

PR #3920, opened 2026-04-02
docs: Development base image

shiroyasha:


docs/prd/superplane-dev-base.md
```diff
@@ -0,0 +1,140 @@
...
+- This is faster by around 100 seconds from what we have atm.
+
+## One vs. Multiple base images
+
+I experimented with two approaches:
+
+1/ Having one base image that is shared between the app and agent
+2/ Having two base images, one for app, one for the agent
+
+Option 1/ was faster on CI. This approach has downsides, but I'm mostly looking for performance here.
```
lucaspin, line 36: Surprised that [1] was faster on CI. Any clues on why? I'd expect [2] to be faster since the images would be smaller

forestileao: me too, even if they are built in parallel?

shiroyasha: I was also surprised. I think it boils down to "it needs to download more data" from the internet. Even in parallel, this is slower.

shiroyasha:
Merging this. It seems that we are generally ok with this approach. The question of single or multiple docker images remains to be tested during implementation.

Issue #3714, opened 2026-03-26
Component config: in-context AI assist


AleksandarCole:
**Problem:**  
AI builder exists on the canvas (and is planned for the home page), but configuring a **single component** still requires manual work for messages, expressions, and other fields. Users often want lightweight help for just the active component—not a full canvas build.

**Current vs proposed:** **Current:** AI at canvas/home scope; component panels are manual. **Proposed:** In-context AI in the **individual component configuration** UI (e.g. generate copy, suggest or refine expressions, field-level assistance) scoped to the selected component.

**Use case:**  
Authors iterating on one node who want quick AI help without leaving the inspector or invoking full-canvas generation.

shiroyasha:
Returning this to the backlog. We have a good start with @re-pixel, but we are not going to focus on this week.

PR #3921, opened 2026-04-02
docs: Plan out the implementation of in-component configuration assistance

re-pixel:
### Related Issues

https://github.com/superplanehq/superplane/issues/3714


docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+
+We want a **narrow** assistive surface: “describe intent → see proposed string → apply to
+this field” without opening the full builder or a separate chat product.
+
+## Goals
+
+1. Offer **one-shot suggest** for eligible fields on the workflow **Settings** / configuration
+   UI, gated by product and permissions.
+2. Keep **trust boundaries clear**: browser → authenticated SuperPlane API → agent HTTP with
+   **scoped JWT** (`purpose: config-assistant`), separate from AI Builder tokens.
```
shiroyasha, line 31: Nice! Haven't even thought about this. Good that we covered it.

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+3. Run a **small** Python agent (`agent/src/config_assistant/`) with **structured output**
+   (`value`, optional `explanation`); **no** import of the builder’s `agent.py` or canvas tools.
```
shiroyasha: It will need some tools, or would you pass all the info needed in the request?

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+   later).
+
+## Scope and Semantics
+
+### Eligible fields (frontend allowlist)
+
+Assistance is shown only when the feature is enabled **and** the field passes
+`isConfigAssistantSupportedField` (`web_src/src/lib/configAssistantFields.ts`):
+
+- **Allowed types:** `expression`, `string`, `text`, `url`, `cron`.
```
shiroyasha, line 70: Lets sanity check this line @lucaspin @forestileao

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+  **Mitigation:** document Compose defaults; log failed forwards from Go.
+
+## Open Questions
+
+1. Should we align **org-level “AI off”** with AI Builder so suggests return 403 when the org
+   disables AI features?
+2. Do we add **catalog-level** opt-in (`Field` schema) without widening blast radius, or keep
+   the allowlist until product stabilizes?
+3. Should we add **streaming** suggestions or multi-turn inside the panel only?
+4. Do we need **structured `value`** (e.g. JSON) for future field types, or stay string-only?
```
shiroyasha: String only it 👍

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+- **Risk:** Operator misconfigures `AGENT_HTTP_URL`.  
+  **Mitigation:** document Compose defaults; log failed forwards from Go.
+
+## Open Questions
+
+1. Should we align **org-level “AI off”** with AI Builder so suggests return 403 when the org
+   disables AI features?
+2. Do we add **catalog-level** opt-in (`Field` schema) without widening blast radius, or keep
+   the allowlist until product stabilizes?
+3. Should we add **streaming** suggestions or multi-turn inside the panel only?
```
shiroyasha: No.

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+  **Mitigation:** instruction and context limits; future rate limits and org AI flags.
+
+- **Risk:** Operator misconfigures `AGENT_HTTP_URL`.  
+  **Mitigation:** document Compose defaults; log failed forwards from Go.
+
+## Open Questions
+
+1. Should we align **org-level “AI off”** with AI Builder so suggests return 403 when the org
+   disables AI features?
+2. Do we add **catalog-level** opt-in (`Field` schema) without widening blast radius, or keep
```
shiroyasha: I don't understand this one.

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Acceptance Criteria
+
+1. Eligible fields show the assistant; **sensitive** and disallowed types do not.
+2. Successful suggest returns a non-empty `value`; errors surface in the panel without
+   breaking the rest of the sidebar.
+3. Builder JWTs cannot be used for `/config-assistant/suggest` and vice versa (purpose
+   separation).
+4. User without canvas **update** cannot obtain successful suggests for that canvas.
+5. Agent and app share JWT signing secret; misconfiguration yields clear 503/4xx behavior
```
shiroyasha: Response speed is a mandatory criteria for this system to be good 👍

docs/prd/inline-config-assistant.md
```diff
@@ -0,0 +1,166 @@
...
+Go enforces caps: instruction length (runes) and max `field_context_json` bytes.
+
+### Response
+
+- `value` (string): proposed field value; must be non-empty for success.
+- `explanation` (optional): short human-readable note.
+
+## Public API (Go)
+
+- **HTTP:** `POST /api/v1/config-assistant/suggest` (gRPC-Gateway for
```
lucaspin: Not sure having a separate /api/v1/* path just for this is the best idea. Can't we put this under /api/v1/agents/* too?

shiroyasha:
Good start! One thing I don't yet understand from this document is how it will know about the existence of other nodes that it needs to reference, or they payload that it can use.

e.g. User is configuring the `if` node in this flow: gh.onIssueComemnt -> if -> ...

Input: Check if the github comment starts with "/command"
Expected Output: `root().data.issue.content contains "/command"`.

Basically the querstion is: How will it know to create this: `root().data.issue.content`.

docs/prd/inline-config-assistant.md
re-pixel: we already have `field_context_json` that helps for non-ai autocomplete for expressions. i thought we could just use that instead of defining tools as it contains sufficient information about upstream nodes.

docs/prd/inline-config-assistant.md
re-pixel: 2. basically says do we have a list of `Field`s for which the AI assistance is enabled or do we add a property to the `Field` (e.g. configAssistEnabled). i didn't want to change the `Field` schema for the initial implementation.

docs/prd/inline-config-assistant.md
re-pixel: you're right, it makes more sense to put it under /api/v1/agents/

shiroyasha:
@re-pixel looks good. Please update the doc based on the convo and we are ready to merge it.

PR #3933, opened 2026-04-03
prd: new canvas editing

AleksandarCole:


docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
+## 1. What ships
+
+Draft-first canvases so you’re not **editing the graph that production is running**. You work on an unpublished draft in edit mode, save there, then publish to live if change management is off, or go through change request if it’s on. Live is for running and watching what’s actually live, not for rewiring it in place.
```
shiroyasha: I like this separation, matches my intuition how release pipelines work.

docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
...
+
+Versioning stays on for canvases. Change management is its own setting: it only changes what “go live” means.
```
shiroyasha: Basically this is the de-facto mode, no way to turn it off or on, right?
We don't even need to specifically mention it as a special thing.

docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
...
+People who already think in terms of **CI pipelines, deployment tools, and other workflow builders**: you define or change the pipeline in one place, and you **watch runs, logs, and outcomes somewhere else**. They expect editing configuration and digging into execution detail to be **two different modes or screens**, not one mushy view where the graph is both “what I’m changing” and “what’s live right now.”
+
+That’s who draft vs live is for. Builders who mostly want the operational view can stay on live; when they need to change the workflow they step into edit, same mental model as opening a pipeline editor vs opening a run.
+
+---
+
+## 4. Behavior
+
+**Live vs edit**  
+Live: structure is read-only, you can run and dig in. Edit: your draft, saves go to the draft. Exit: save and leave, or discard and back to live, with prompts when there’s unsaved stuff. If a draft already exists when you hit Edit, choose continue draft or start over from live.
```
shiroyasha: Exit is a weird term here. How my brain reads it "I exit the draft == I delete it".

I would need to see it in action.

docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Publishing**  
+CM off: one publish action promotes the draft to live. CM on: the publish button is submit for review, not straight to live.
+
+**Settings**  
+Change management shows up in canvas settings where we already expose policy.
```
shiroyasha: Maybe not for this PR to address, but I'll ask it anyway:
If I'm the maintainer of a canvas e.g. ephemeral machines, can I publish without going through the change management.

My expectation is:
- maintainer = I can change it directly, or merge stuff that other people proposed

docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Before publish**  
+If there are blocking warnings on the graph, publish is disabled and we show what’s wrong next to the button.
+
+**Versions**  
+Versions panel and the Versions button only show in edit mode. On live you get Edit, not the full version browser.
+
+In edit, you can pick another version to look at (draft, live snapshot, history, open CRs, same rules as today). After you switch, the view fits all components like the zoom “fit all”, even if nodes showed up a beat late.
```
shiroyasha: What does "live snapshot" mean?

docs/prd/updated-canvas-editing.md
```diff
@@ -0,0 +1,90 @@
...
+- **Internal:** Our own team uses the new model **without anyone having to explain it**. If we’re still translating draft vs live in Discord, we’re not there.
+- **Prod dogfood:** Ship to production, use it ourselves for a week, and see **whether anything bubbles up** (confusion, workarounds, “this feels wrong” feedback).
+- **Sessions with outsiders:** Product reviews, hackathons, similar. Do people **get the UI on their own** or do they wander, click randomly, and ask what they’re looking at?
+- **Cloud metrics:** Pull usage and funnel-style signals from the **hosted cloud instance** and read them alongside the qualitative stuff above.
+
+Iterate speed and dry run stay separate from this checklist until that follow-up ships.
+
+---
+
+Details live in specs; this is the short version.
```
shiroyasha: Now that this is a PR we can write down the rest as well, no?

forestileao (commented):
**What I liked:**

* **Enforced versioning:** Having an edit mode ensures changes are controlled and prevents people from accidentally breaking production canvases.
* **Change request flow:** The current Togglab setup for approving/rejecting CRs works well—similar to GitHub branch protection rules.

**Open questions:**

* In the video, I noticed a **“Save and Exit”** option in the dropdown. Do we always need to click that to update the canvas?
  In the past, it was easy to forget to save and lose work, so I’d strongly prefer **autosaving drafts**. Let me know if I misunderstood how this works.

**Ideas / things we discovered while proposing changes:**

* What if we could have **multiple drafts**?
  This might be out of scope, but it would be useful to keep several change proposals without losing previous versions, so we can compare and decide which one is best.

docs/prd/updated-canvas-editing.md
AleksandarCole: That's right, versioning is a default and not mentioned anywhere in new UI. The "Change Management" (better term needed) - is something that you can toggle on/off - and it's the part where publish is not automatic but instead needs to go through approval process first (how our versioning works now)

docs/prd/updated-canvas-editing.md
AleksandarCole: Exit sucks - but I just use that term in prototype - Petar will find a better UI flow/copy. 

@shiroyasha There's a link to the video at the bottom.

docs/prd/updated-canvas-editing.md
AleksandarCole: Change management right now is rudimentary. I would discuss the functionality of that in future PRDs (separate track). 
I'm happy with the 0 to 1 solution we currently have - being able to enforce approval first. 
But for a `MLP` solution we need to invest some time and design it from ground up.

docs/prd/updated-canvas-editing.md
AleksandarCole: @shiroyasha Are we doing detailed specs? If so, can I get example of the ones you think are good.

shiroyasha: Not really, I was just curious what else is there because of this line.

AleksandarCole:
> **Open questions:**
> 
> * In the video, I noticed a **“Save and Exit”** option in the dropdown. Do we always need to click that to update the canvas?
>   In the past, it was easy to forget to save and lose work, so I’d strongly prefer **autosaving drafts**. Let me know if I misunderstood how this works.
> 
> **Ideas / things we discovered while proposing changes:**
> 
> * What if we could have **multiple drafts**?
>   This might be out of scope, but it would be useful to keep several change proposals without losing previous versions, so we can compare and decide which one is best.

@forestileao Auto-save is always happening in the background (for the draft), so if you close the tab, or power goes out - your draft is saved it's not lost. Petar should work a bit on this interaction though - not sure about having "discard" and "save" as options is the best approach either. 

As per multiple drafts - it's a good idea but I would keep it simple for now. If you have `change management` enabled, nothing is stopping you from submitting multiple and only merging one. Let's have a clean and easy to understand flow first though and when use-cases pop up for extended functionality we can add it.

docs/prd/updated-canvas-editing.md
AleksandarCole: 👍 all clear. It was just a clumsy note that `API and Backend Requirements` and similar things are excluded.

docs/prd/updated-canvas-editing.md
shiroyasha: Watched the video. 

Lets review to @ropsii to review it / polish the POC.
I feel that one level of UX polish would do wonders to simplify it.

Current weirdness:

1/ The "exit" terminology. Collides with "delete" which makes it hard to think.
2/ When I click on publish I get a big modal with things I don't understand: I didn't write yaml and yet you are showing me diffs on yaml as a major part of that dialog.
3/ The edit/publish flow needs to be lighter if we don't want to piss off maintainers. 1 click edit -> 1 click publish, any modal/dialog/question in between is pain and friction.

docs/prd/updated-canvas-editing.md
shiroyasha: I propose hiding the change management part completely until we get back to it with more love.

shiroyasha:
@forestileao can you extend this document to cover the high-level architecture? 
Database schemas, apis, where would things be storred, any relevant performance details.

forestileao:
@AleksandarCole @shiroyasha Added a techincal proposal to cover everything above it. I hope I covered everything.

I described API and DB changes mainly. UI and CLI changes are more abstract

docs/prd/updated-canvas-editing.md
AleksandarCole: Agreed on points 1-2. 

Regarding 3 - having a step of setting a name is not a big deal (especially if it's prefilled). It's a common pattern in similar tools. 

Not having it gets us into that "let's figure out the good auto-naming strategy that works for everyone" category.

markoa:
Overall we can merge this but as discussed in several sessions:
- in the future don't need technical spec so soon before final prototype
- we will prototype edit mode + saving in much smaller scope first -- we need to move fast atm and don't need the ultimate solution immediately

PR #3935, opened 2026-04-03
prd: canvas runs view

AleksandarCole:


forestileao (commented):
I think there’s a gap around node inspection UX.

From what I see, the PRD defines node inspection in **Run View** (double-click with run-scoped data), but it doesn’t clearly define how this relates to **Live node inspection** or whether we reuse the same UI vs create separate experiences. 

My suggestion:

* Clicking a node in **Live Canvas** should always show **latest state**
* **Run View** should be the only place for **run-specific (historical) inspection**
* Reuse the **same node inspection UI** everywhere (just change the data source) (NOT SURE IF THIS IS GOOD UX, just something that popped in my head)

This keeps the mental model simple:

* Live = “what’s happening now”
* Run View = “what happened in that run”

and avoids having multiple sidebars / duplicated inspection patterns.

let me know your thoughts.

AleksandarCole:
Hey @forestileao good points - this was WIP - i have updated it now and I hope it covers details better. 
Your instinct is right - imo we have to have similar UI/UX between Runs/Live or have completely different UI/UX - keeping it in some weird `same same, but different` state will be confusing. For now, I suggested that we go with the approach of keeping them very similar. This means that we would ditch the `runs in sidebar` on the live view completely.

lucaspin:
@AleksandarCole Can we close this now that runs are merged?

PR #3936, opened 2026-04-03
chore: Better specify several any types

shiroyasha:


shiroyasha:
unknown is a similar shit as any, but at least it raises an error on the call site if not dealt with it properly.

PR #3938, opened 2026-04-03
fix: Canvas nodes teleporting during autosave + dragging

forestileao:
**Problem**

Two race conditions caused nodes to snap back to old positions during autosave:

- **Versioned mode (edit mode):** After a save, the mutation's onSuccess handler updated the version detail cache with raw server positions. When hasPendingLocalCanvasState became false, the version sync effect replaced the canvas spec with the stale version data — undoing the position merge that syncCurrentCanvasWithSavedVersion had just applied.
- **Mid-drag refetch:** During an active drag, node positions only exist in ReactFlow's internal state — the query cache and pendingPositionUpdatesRef aren't updated until drag ends. If a cache refetch happened mid-drag (e.g. from a WebSocket echo), the prop sync effect in useCanvasState overwrote the in-progress drag position with stale server data.

**Fix**

- Version snapshot ref: syncCurrentCanvasWithSavedVersion now updates lastAppliedVersionSnapshotRef so the version sync effect recognizes the version as already applied and skips the spec replacement.
- Drag position preservation: The node sync effect in useCanvasState now preserves the position of any node with dragging: true, preventing cache refetches from interrupting active drags.

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -1175,6 +1175,15 @@ export function WorkflowPageV2() {
         return;
       }
 
+      // Mark the saved version as already applied so the version sync effect
+      // (which replaces canvas spec with loadedCanvasVersion.spec) doesn't
+      // overwrite the merged positions we set below.
+      const versionId = version.metadata?.id;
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #3939, opened 2026-04-03
fix: draft resetting on tab switch

forestileao:
## Problem
Switching to another browser tab and returning to Superplane triggered a refetch of the canvas detail query. This reset the local canvas state — including in-progress draft edits and node positions — to the server's last saved state.

## Fix
Added refetchOnWindowFocus: false to the useCanvas query, matching every other canvas query in the file. Canvas state is already kept up to date via WebSocket lifecycle events.


forestileao:
👍

PR #3940, opened 2026-04-03
fix: merge realtime updates not working

lucaspin:
Merge component is not returning its current execution ID in ProcessQueueItem() while the execution is waiting, only on completion. That prevents the worker of sending WebSocket updates to the UI.

#### Before

https://github.com/user-attachments/assets/1b63e00e-57ef-41d9-86d0-22316ace6535

#### After

https://github.com/user-attachments/assets/ce6891fb-dac2-4e3f-a1d4-2a735730852d


lucaspin:
👍

PR #3941, opened 2026-04-03
chore: cleanup expression execution inside of components

lucaspin:
### Issue

Right now, we have an ctx.ExpressionEnv() which is exposed to components so they can execute expressions. There are a couple of problems with this approach:
- Since the core only exposes the expression environment, the components themselves are still required to provide implementations for functions exposed inside of expressions (root, previous). That leads to duplication. Right now, if, filter, and merge components have duplicated code for compiling and running these expressions.
- Expression evaluation is split, but both parts are tightly coupled. I cannot modify one without modifying the other.

### Solution

Remove ExpressionEnv, and add ExpressionContext:

```golang
type ExpressionContext interface {
    Run(expression string) (any, error)
}
```

With this, all the logic about how expressions are compiled, and evaluated is inside of the core, and components can just do this:

```golang
output, err := ctx.Expressions.Run(spec.Expression)
if err != nil {
    return fmt.Errorf("expression evaluation failed: %w", err)
}
```

### DIFF
--- a/pkg/components/merge/merge.go
+++ b/pkg/components/merge/merge.go
@@ -253,7 +253,11 @@ func (m *Merge) ProcessQueueItem(ctx core.ProcessQueueContext) (*uuid.UUID, erro
 				return nil, err
 			}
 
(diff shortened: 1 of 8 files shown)


lucaspin:
👍

PR #3942, opened 2026-04-03
chore: cleanup mappers for if / filter

lucaspin:
Expressions used in filter/if components are not simple left/operator/right expressions. Trying to treat them as such introduced a lot of slop code. Here, we are cleaning that up. Check the [Discord message](https://discord.com/channels/1409914582239023200/1488540155357696213/1489423953591140362) for more details.


lucaspin:
👍

PR #3951, opened 2026-04-04
chore: remove CanvasNodeExecution.input field

lucaspin:
Now that expressions can access data from any node, CanvasNodeExecution.input is no longer needed in the UI, so that extra query every time we serialize node executions is useless.


lucaspin:
👍

PR #3952, opened 2026-04-04
chore: update API listing limits

lucaspin:
Nowhere in the UI we need these limits to be this big. Safer to be more restrictive right now and open up later on than the opposite


lucaspin:
+1

PR #3953, opened 2026-04-04
chore: remove __type=semaphoreBlocks from ChainItem

lucaspin:
Cleaning up ChainItem to not have logic about component-specific displays. Starting with semaphore.runWorkflow component. Took the opportunity to also clean up the mapper for that component, which got rid of 50+ eslint warnings.


lucaspin:
👍

PR #3954, opened 2026-04-04
chore: do not use ApprovalTimelineEntry in merge mapper

lucaspin:
In order to remove ApprovalTimelineEntry from ChainItem, we need to stop using it in approval and merge component mappers. This is the first step in that direction.

While doing that, I realized the merge mapper had way too much logic inside of it, and required an additionalDataMapper. I also want to get rid of that, so the merge implementation was updated to save all the information required in its execution metadata. With that, the merge mapper no longer needs any more additional data that is built just on the UI side. All the information is in the execution metadata.

### Core change

ProcessQueueContext.CountDistinctIncomingSources() becomes ProcessQueueContext.DistinctIncomingSources(). New method returns the same data, but it also allows us to include more information in the future.


lucaspin:
👍

PR #3971, opened 2026-04-06
fix: harden canvas editor runtime and add frontend regression coverage

forestileao:
## Summary

This PR fixes a set of production frontend issues affecting the canvas editor and admin/auth flows.

### What changed

- Replaced Monaco's default CDN worker loading with bundled local workers
- Moved auth redirect navigation out of render and into `useEffect`
- Improved handling of generic browser network failures like `Failed to fetch`
- Removed handled `console.error` noise from canvas save and emit-event flows
- Added a fallback hidden dialog title to satisfy Radix dialog accessibility requirements
- Added targeted Vitest coverage for the reported regressions

### Monaco worker script load failure
Monaco was still using the default `@monaco-editor/react` loader behavior, which falls back to jsDelivr in production. This PR bootstraps Monaco with bundled workers so the app no longer depends on the CDN for editor worker scripts.

### `navigate()` called during render
`AuthGuard` was calling `navigate()` during render for unauthenticated users. This PR moves that redirect into `useEffect`, which removes the React lifecycle warning.

### `TypeError: Failed to fetch`
Some canvas save and manual emit-event failures were already handled in the UI, but still surfaced noisy generic browser errors and `console.error` entries. This PR:
- treats generic browser network messages as fallback cases
- keeps user-facing operation-specific error messages
- avoids extra console noise for handled failures

### Dialog accessibility warning
Some dialogs could render without a `DialogTitle`, which causes Radix accessibility errors. This PR adds a hidden fallback title in the shared dialog components when no explicit title is present.

## Tests

Added targeted Vitest coverage for:
- Monaco worker bootstrap configuration
- render-safe auth redirect behavior
- dialog fallback title behavior
- handled network failure behavior in `EmitEventModal`

## Notes

The current frontend Docker build path appears to need a larger Node heap for production bundling. The application build itself is valid; the failure is due to container memory limits during the make target.

### DIFF
--- a/web_src/src/components/ui/dialog.tsx
+++ b/web_src/src/components/ui/dialog.tsx
@@ -2,41 +2,8 @@ import * as React from "react";
 import * as DialogPrimitive from "@radix-ui/react-dialog";
 import { XIcon } from "lucide-react";
 
+import { hasDialogTitle } from "@/lib/dialogTitle";
 import { cn } from "@/lib/utils";
-
-function getReactDisplayName(childType: string | React.JSXElementConstructor<unknown>): string | undefined {
-  if (typeof childType === "string") {
-    return childType;
-  }
-
-  if ((typeof childType === "function" || typeof childType === "object") && "displayName" in childType) {
-    return typeof childType.displayName === "string" ? childType.displayName : undefined;
-  }
-
-  return undefined;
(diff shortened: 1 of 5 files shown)


forestileao:
👍

PR #3972, opened 2026-04-06
feat: track token usage per chat and per organization

0vertake:
## Summary

- Record per-run token usage (input, output, cache read/write, total, estimated cost) into a new agent_chat_usage_records detail table, with denormalized cumulative totals on agent_chats
- Expose per-chat usage in existing ListAgentChats / DescribeAgentChat responses and add a new DescribeAgentUsage RPC for org-level aggregation
- Display per-chat token counts in the AI Builder conversation list and org-level totals with estimated cost on the Settings > Usage page

## Details

Agent (Python): _record_usage() is called after each AgentRunResultEvent, writing a detail row and incrementing cumulative columns within a single transaction (with FOR UPDATE lock). Cost estimation covers Claude 4.5/4.6 models only; non-Claude runs store NULL in detail and 0.0 in cumulative. The org-level query reads from the denormalized agent_chats columns for performance.

Backend (Go): New DescribeAgentUsage action proxies the internal DescribeOrganizationAgentUsage gRPC call. Per-chat usage is mapped into AgentChatInfo in the existing list/describe actions. Authorization scoped to agents:read at org level.

Frontend: Token counts appear below each chat title in the conversation list, refreshed after each stream completes. The Usage page shows an "AI Agent Tokens" card with loading/error/data states. Cost subtitle clarifies "Claude models only".

Demo video: https://www.youtube.com/watch?v=Znq2BKi0RtY


agent/src/ai/session_store.py
```diff
@@ -496,6 +512,91 @@ def set_initial_chat_message_if_missing(self, chat_id: str, initial_message: str
...
+        input_tokens: int,
+        output_tokens: int,
+        cache_read_tokens: int,
+        cache_write_tokens: int,
+        total_tokens: int,
+        estimated_cost_usd: float | None,
+    ) -> None:
+        now = _utcnow()
+        record_id = str(uuid.uuid4())
+        cost = estimated_cost_usd if estimated_cost_usd is not None else 0.0
```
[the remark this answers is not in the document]
0vertake: This is intentional as the frontend shows "Claude models only" next to the cost so users know it's partial.

web_src/src/ui/BuildingBlocksSidebar/AiBuilderChatPanel.tsx
```diff
@@ -353,6 +358,16 @@ function formatSessionDate(value: string): string {
...
 
+function formatTokenCount(tokens: number): string {
+  if (tokens >= 999_950) {
+    return `${(tokens / 1_000_000).toFixed(1)}M tokens`;
+  }
+  if (tokens >= 1_000) {
+    return `${(tokens / 1_000).toFixed(1)}k tokens`;
+  }
+  return `${tokens} tokens`;
+}
```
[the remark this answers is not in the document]
0vertake: This isn't a concern for this PR.

lucaspin:
@0vertake seems like the connection leak was fixed in https://github.com/superplanehq/superplane/pull/3974. You'll need to update your branch

0vertake:
@lucaspin Branch updated, should all be good now.

agent/db/migrations/20260403134227_add-chat-token-usage.up.sql
```diff
@@ -0,0 +1,24 @@
+BEGIN;
+
+CREATE TABLE agent_chat_usage_records (
```
lucaspin: I think agent_chat_messages record should have a foreign key pointing to records on this table. What if we renamed this table to `agent_chat_runs` instead? Then, agent_chat_messages can have a run_id referencing this. Then, the schema would look like this:

```
agent_chats (
  id uuid PRIMARY KEY,
  org_id uuid NOT NULL,
  user_id uuid NOT NULL,
  canvas_id uuid NOT NULL,
  initial_message TEXT,
  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
)

agent_chat_runs (
  id uuid PRIMARY KEY,
  chat_id uuid NOT NULL REFERENCES agent_chats(id) ON DELETE CASCADE,
  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
  model text NOT NULL DEFAULT '',
  input_tokens bigint NOT NULL DEFAULT 0,
  output_tokens bigint NOT NULL DEFAULT 0,
  cache_read_tokens bigint NOT NULL DEFAULT 0,
  cache_write_tokens bigint NOT NULL DEFAULT 0,
  total_tokens bigint NOT NULL DEFAULT 0,
  estimated_cost_usd double precision,
)

agent_chat_messages (
  id uuid PRIMARY KEY,
  chat_id uuid NOT NULL REFERENCES agent_chats(id) ON DELETE CASCADE,
  run_id uuid NOT NULL REFERENCES agent_chat_runs(id) ON DELETE CASCADE,
  message_index INTEGER NOT NULL,
  message JSONB NOT NULL,
  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
)
```

What do you think?

cc @shiroyasha

0vertake: Makes sense, we get a nicer hierarchy with this. I can start working on it or we can wait for Igor's approval?

lucaspin: Go ahead. @shiroyasha will scream if we're doing something wrong

protos/private/agents.proto
```diff
@@ -73,4 +74,20 @@ message ChatInfo {
   string id = 1;
   string initial_message = 2;
   google.protobuf.Timestamp created_at = 3;
+  ChatUsage usage = 4;
+}
+
+message ChatUsage {
```
lucaspin, line 79: I don't like that we are showing the token usage in here:

Image

I think that's unnecessary information for the user. Seeing token usage at the chat level is useful for us only, I think. So, let's remove this information from this API response and from the UI

agent/db/migrations/20260403134227_add-chat-token-usage.up.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE agent_chat_runs (
+    id uuid PRIMARY KEY,
+    chat_id uuid NOT NULL REFERENCES agent_chats(id) ON DELETE CASCADE,
+    model text NOT NULL DEFAULT '',
+    input_tokens bigint NOT NULL DEFAULT 0,
+    output_tokens bigint NOT NULL DEFAULT 0,
+    cache_read_tokens bigint NOT NULL DEFAULT 0,
+    cache_write_tokens bigint NOT NULL DEFAULT 0,
+    total_tokens bigint NOT NULL DEFAULT 0,
+    estimated_cost_usd double precision,
```
lucaspin: Not so sure about storing the cost here too. Seems like it would be better to just store the amount of tokens, since cost is derived from that. What do you think @shiroyasha?

protos/agents.proto
```diff
@@ -81,6 +81,17 @@ service Agents {
       tags: "Agent";
     };
   }
+
+  rpc DescribeAgentUsage(DescribeAgentUsageRequest) returns (DescribeAgentUsageResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/agents/usage"
```
lucaspin: I don't think this is the best place for this API. I think the [organizations.proto](https://github.com/superplanehq/superplane/blob/main/protos/organizations.proto#L188) is a better place for that. @forestileao will be working on applying limits on this, so let's remove this API from this pull request.

forestileao: yes, use the same from organizations.proto

agent/db/migrations/20260403134227_add-chat-token-usage.up.sql
shiroyasha: Agreed. Store only tokens.

protos/private/agents.proto
shiroyasha: Ditto. Only noise for users.

protos/private/agents.proto
```diff
@@ -73,4 +74,19 @@ message ChatInfo {
   string id = 1;
   string initial_message = 2;
   google.protobuf.Timestamp created_at = 3;
+  ChatUsage usage = 4;
```
lucaspin: Since we are not showing the usage on the chat level anymore, we don't need this field here

lucaspin:
CI is failing due to a flaky test, which will be handled in https://github.com/superplanehq/superplane/issues/4065

PR #3976, opened 2026-04-06
fix: 500 error on canvas version update when canvas is deleted

forestileao:
## Summary
- `updateLiveCanvasWithoutVersioning` returned a raw 500 when the canvas had been soft-deleted (e.g. user has editor open, another user deletes the canvas, user saves)
- Added `ErrRecordNotFound` check to return `NotFound` instead of `Internal`

## Sentry Issues
- HTTP 500 PUT /api/v1/canvases/{canvasId}/versions (Issue ID: 7340134055)

## Test plan
- [x] Test: deleted canvas returns NotFound instead of Internal error


forestileao:
👍

PR #3977, opened 2026-04-06
fix: 500 error on list group users for non-existent groups

forestileao:
## Summary
- `ListGroupUsers` returned 500 Internal when queried with a group name that doesn't exist (e.g. `devops_team`, `security_and_compliance`)
- Added group existence validation in `GetGroupUsers` that returns a proper error before querying users
- `ListGroupUsers` now returns `NotFound` instead of `Internal` for missing groups
- Added nil check for role metadata to prevent nil pointer dereference

## Sentry Issues
- HTTP 500 /api/v1/groups/devops_team/users (Issue ID: 7376816811)
- HTTP 500 /api/v1/groups/security_and_compliance/users (Issue ID: 7359654997)

## Test plan
- [x] Test: missing group returns NotFound instead of Internal
- [x] Test: delete group test updated to expect error on GetGroupUsers for deleted group

### DIFF
--- a/pkg/authorization/service.go
+++ b/pkg/authorization/service.go
@@ -283,26 +283,18 @@ func (a *AuthService) UpdateGroup(domainID string, domainType string, groupName 
 
 func (a *AuthService) AddUserToGroup(domainID string, domainType string, userID string, group string) error {
 	domain := prefixDomain(domainType, domainID)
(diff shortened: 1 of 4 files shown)


forestileao:
👍

PR #3978, opened 2026-04-06
chore(ui): rm pages/custom-component

lucaspin:
### Why?

- Custom components are disabled in the UI for a while
- Using CanvasPage component in there was a mistake. Building new page from scratch would give us a better output
- CanvasPage updates requires you to consider this usage path, which is a friction point. Since this needs to be built from ground up again, it's a lot better to remove it now.


lucaspin:
👍

PR #3980, opened 2026-04-06
chore(ui): rm pages/node-run

lucaspin:
Same reason as https://github.com/superplanehq/superplane/pull/3978. Also, removing storybooks for the canvas page, since they are dead weight now too.


lucaspin:
👍

PR #3987, opened 2026-04-06
fix: reduce UI build memory usage by narrowing tree-shaking icon resolution

forestileao:
## Summary

This PR fixes `make check.build.ui` failing with a Node.js heap out-of-memory error during the Vite production build.

It Removes runtime wildcard icon resolution from `lucide-react`, replacing it with a tree-shakeable icon registry.

## Changes

- narrow Monaco bootstrap import to `monaco-editor/esm/vs/editor/editor.api.js`
- keep bundled Monaco workers configured through `MonacoEnvironment`
- replace wildcard `lucide-react` namespace lookups with an explicit shared icon registry
- remove duplicated local icon resolution logic in UI components
- preserve compatibility aliases for legacy icon slugs
- update Monaco and icon resolver tests

## Result

`make check.build.ui` now completes successfully without requiring a `NODE_OPTIONS=--max-old-space-size=4096` override.


forestileao:
👍

PR #3990, opened 2026-04-06
chore: ListUsers returns roles only when include_roles is used

lucaspin:
ListUsers always returns role assignments, even though we only need it in the members page. That increases the time it takes to load the canvas page. Here, we address it by including a new `include_roles` query param in the ListUsers endpoint. By default, no role assignments are returned.

### Other changes

- Fetch account providers for users in batch, to avoid doing a single separate query per user when serializing account providers.
- The existing include_service_accounts is removed, since it's not used, and if we want service accounts, we should use the service accounts endpoints.
- We also update ListGroupUsers to not return roles
- Update me.proto to return user roles too. This is required because the approval mapper needs that information


lucaspin:
👍

PR #3994, opened 2026-04-07
chore: remove ListUserRoles messages from API

lucaspin:
https://github.com/superplanehq/superplane/pull/3969 forgot to remove the actual messages and the endpoint implementation.


lucaspin:
👍

PR #3996, opened 2026-04-07
chore: refactor approval to not rely on ComponentAdditionalDataBuilder

lucaspin:
This pull request is still not ready for review. I just opened it to get a review from BugBot.

### Main changes

- Approval implementation was updated to store role and group refs in the node metadata. That allows us to update the approval mapper to not rely on ComponentAdditionalDataBuilder. ComponentAdditionalDataBuilder now becomes something no components use, so we can safely remove it.
- In order to control when showing approve/reject buttons easily, we extend /api/v1/me response to include user name and groups the user belongs to. That is better than relying on loading groups and then checking if the user belongs to the group used in the approval requirement. This also helps us with initial canvas loading since now we don't need to load groups there anymore


lucaspin:
👍

Issue #3963, opened 2026-04-05
Can't close the canvas config panel


shiroyasha:
<img width="2946" height="1768" alt="Image" src="https://github.com/user-attachments/assets/4d65c0b9-ebb4-4473-920e-627d6a15e07d" />

PR #4004, opened 2026-04-07
fix: be able to close building blocks sidebar when rest of it is disabled

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/3963

## Summary

  - Fix building blocks sidebar close button being unclickable when the sidebar
  is in disabled/read-only state (e.g. viewing a live canvas with versioning
  enabled but not in edit mode)
  - The disabled overlay (z-30) covered the entire sidebar including the close
  button — fixed by giving both close buttons z-40 so they remain clickable
  above the overlay
  - Added data-testid="close-sidebar-button" to the close buttons for
  testability

##  Root Cause

  When versioning is enabled and the user is viewing the live canvas (not in
  edit mode), isReadOnly becomes true, which passes disabled={true} to the
  BuildingBlocksSidebar. A full-coverage overlay with absolute inset-0 z-30
  renders to block component interactions, but it also blocks the close button
  since the sidebar container is z-21 and the close buttons had no explicit
  z-index.

### DIFF
--- a/test/e2e/canvas_sidebar_close_test.go
+++ b/test/e2e/canvas_sidebar_close_test.go
@@ -1,0 +1,103 @@
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	pw "github.com/playwright-community/playwright-go"
+	"github.com/stretchr/testify/require"
+
+	"github.com/superplanehq/superplane/pkg/database"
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #4005, opened 2026-04-07
fix: TypeError: z.substring is not a function

forestileao:
## Summary

  - Fix TypeError: z.substring is not a function crash caused by calling
  .substring() on non-string configuration/metadata values in canvas node
  mappers
  - Add a safe truncate(value: unknown, maxLen, ellipsis) utility to
  safeMappers.ts that coerces any value to a string before truncating
  - Replace all 20+ unsafe .substring() calls across 18 mapper files with the
  safe truncate() utility
  - Harden the existing truncateForDisplay() functions in statuspage/utils.ts
  and aws/ecs/common.ts to accept unknown instead of string

 ## Root Cause

  Mapper subtitle() and metadata-building functions call .substring() on
  configuration values (e.g., configuration.query, source.nodeId,
  configuration.checkRule) assuming they're strings. When the API returns a
  non-string value (number, object, array), .substring() throws TypeError:
  z.substring is not a function. While the createSafeComponentMapper wrapper
  catches these errors, it degrades the node to a "Can't display" fallback — the
   proper fix is to prevent the error entirely

### DIFF
--- a/web_src/src/pages/workflowv2/mappers/safeMappers.spec.ts
+++ b/web_src/src/pages/workflowv2/mappers/safeMappers.spec.ts
@@ -496,7 +496,7 @@ describe("safe mapper recovers from substring on non-string config values", () =
     };
     const safe = createSafeComponentMapper(underlying, "broken-substring");
     const ctx = makeSubtitleContext();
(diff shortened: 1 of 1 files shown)


forestileao:
👍

Issue #4027, opened 2026-04-08
feat: add agent tools for inspecting node details and runs


re-pixel:
AI Builder currently doesn't have a way to see past runs and event payloads and cannot inspect node details.

`superplane_client.py` contains methods  `get_node_details` and `list_node_events`. We can define agent tools that use these methods to improve AI Builder's ability to reason about the canvas.

PR #4030, opened 2026-04-08
feat: add agent tools for inspecting node details, runs and events

re-pixel:
### Description

This PR improves the agent’s ability to inspect workflows and makes failed tool calls easier for the model to handle.

### Canvas and node tooling

Registers `get_canvas_shape`, `get_node_details`, `list_node_events`, and `list_node_executions` so the agent can answer topology, per-node config, activity, and execution outcomes without relying only on the high-level canvas summary.

Fixes `get_node_details` so configuration and node health fields (validation messages, pause state, integration binding) come from the describe-canvas API instead of always being empty.

Adds a `SuperplaneClient.list_node_executions` wrapper and wires it through the agent.

Updates the system prompt and UI/session tool labels for the new tools (and adjusts one session-store test for the `get_canvas` label).

### Consistent tool errors

Introduces `_tool_failure` / `_tool_error_entry` so failures use a single shape: `__tool_error__`, `__tool_name__`, optional `__tool_error_code__`, plus context fields where useful (e.g. `pattern_id`, `name`, `node_id`). List tools still return a one-element list on failure.

Replaces older `{"error": ...}`-style returns in decision-pattern and describe-catalog paths.

Documents the contract in `system_prompt.txt` and adds unit tests for the helper shape.

### Related Issues

https://github.com/superplanehq/superplane/issues/4027

### DIFF
--- a/agent/src/ai/agent.py
+++ b/agent/src/ai/agent.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from pydantic_ai import Agent, RunContext
 from pydantic_ai.models.test import TestModel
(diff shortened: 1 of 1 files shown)


shiroyasha:
Code Looks ok 👍 

How does this affect eval's price and speed?

re-pixel:
> Code Looks ok 👍
> 
> How does this affect eval's price and speed?

On few manual runs I noticed slightly more tokens used (price goes up by 0.1$-0.2$) and a bit less time (1-2s per eval). Relatively small changes.

Also noticed when running evals multiple times that specifically on this branch evals became flaky - most of the times they all run successfully but ocassionally fail silently (showing 24/24 passed). Have to look into that.

PR #4032, opened 2026-04-08
feat: allow running a subset of agent evals by case name

re-pixel:
Developers can now run one or more eval cases without running the full suite.

- `python -m evals.runner --cases name1,name2` — comma-separated names.
- `--list-cases` — prints valid case names and exits.
- `make test.agent.evals` — accepts `CASES` into the agent container; default behavior (all cases) is unchanged.

Example (with make): `make test.agent.evals CASES=github_and_slack,github_issue_opened_to_discord`

Docs: agent/README.md

### DIFF
--- a/agent/evals/runner.py
+++ b/agent/evals/runner.py
@@ -74,6 +74,7 @@ async def runner(*, selected_case_names: list[str] | None) -> None:
             model=env["model"],
             run_usages=run_usages,
             evaluate_wall_seconds=evaluate_wall_seconds,
+            case_names=case_names,
             interaction_log_paths_by_case_name=case_logger.display_paths_by_case_name,
         ).render()
     finally:
--- a/agent/evals/report.py
+++ b/agent/evals/report.py
@@ -24,12 +24,14 @@ class ReportBuilder:
         model: str,
         run_usages: dict[str, RunUsage],
         evaluate_wall_seconds: float,
+        case_names: list[str],
         interaction_log_paths_by_case_name: dict[str, str] | None = None,
     ) -> None:
         self.report = report
         self.model = model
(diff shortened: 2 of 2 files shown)


shiroyasha:
Lets make it simpler for the user:

```
make test.agent.evals <-- runs everything
make test.agent.evals NAME="<eval-name>" <- runs just that one eval
```

re-pixel:
> Lets make it simpler for the user:
> 
> ```
> make test.agent.evals <-- runs everything
> make test.agent.evals NAME="<eval-name>" <- runs just that one eval
> ```

@shiroyasha is this acceptable:

`make test.agent.evals CASES="case1[,case2,...]"` so someone can run both just one eval, and a couple of them

shiroyasha:
If I can also do: `make test.agent.evals CASES="case1"` then yes 👍

agent/evals/runner.py
```diff
@@ -78,8 +109,31 @@ async def runner() -> None:
...
+    )
+    parser.add_argument(
+        "--list-cases",
+        action="store_true",
+        help="Print eval case names and exit.",
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> None:
```
shiroyasha, line 84: It feels like 90% of this file is now dedicated to parsing and understanding case filters.

Lets extract all of this to a dedicated separate file that deals only with case filtering and simply call:

```
selected_case_names = case_filter(argv)
```

and keep the main runner simpler.

shiroyasha:
Nice! Merging it. 🚢

shiroyasha:
Clicked the wrong button 🤦

PR #4039, opened 2026-04-08
feat: add retention policy for agent chats

### LINKED ISSUE #3818: Retention policy for agent chats
We shouldn't retain agent chats forever. Good way to start this is to remove agent chats after 14 days of no updates.

0vertake:
## Summary

- Hourly background task in the agent service deletes expired chats per organization, using each org's retention_window_days from the usage service
- Falls back to AGENT_CHAT_RETENTION_DAYS env var (default 14) when the usage service is unavailable or returns no policy for an org
- Setting AGENT_CHAT_RETENTION_DAYS=0 disables the fallback but still respects per-org policies from the usage service
- Deletes are batched (500 rows per transaction) with a race-safe outer WHERE clause for READ COMMITTED isolation
- Adds composite index on (org_id, updated_at) for efficient per-org cleanup queries
- Fixes missing await on UsageLimitChecker.close()

Closes #3818


lucaspin:
@0vertake now that the agent service is talking with the usage service, I think the retention policy configuration should be based on the limit for retention policy for the organization. @forestileao can you help here?

forestileao:
> @0vertake now that the agent service is talking with the usage service, I think the retention policy configuration should be based on the limit for retention policy for the organization. @forestileao can you help here?

I will need to add a retention policy limit to agent chats on usage service. Then he will only need to grab this value to cleanup stuff.

lucaspin:
@forestileao can't we just use the same retention policy we have there for everything else already?

forestileao:
> @forestileao can't we just use the same retention policy we have there for everything else already?

We can, no problem. Unless we want to make different policies, but I don't think we need

agent/src/ai/chat_retention.py
```diff
@@ -0,0 +1,52 @@
...
+) -> None:
+    fallback_days = config.chat_retention_days
+
+    logger.info(
+        "agent chat retention started: fallback %d days, checking every %ds",
+        fallback_days,
+        CLEANUP_INTERVAL_SECONDS,
+    )
+
+    while True:
```
shiroyasha: Feels way too low level. Wdyt if we use:

```python
from apscheduler.schedulers.asyncio import AsyncIOScheduler

scheduler = AsyncIOScheduler()

scheduler.add_job(
    run_chat_retention_loop,
    "interval",
    seconds=3600,
    args=[store, limit_checker],
)

scheduler.start()
```

agent/src/ai/session_store.py
```diff
@@ -576,6 +576,36 @@ def set_initial_chat_message_if_missing(self, chat_id: str, initial_message: str
...
+                        SELECT id FROM agent_chats
+                        WHERE org_id = %s
+                          AND updated_at < NOW() - make_interval(days => %s)
+                        LIMIT %s
+                    )
+                    AND updated_at < NOW() - make_interval(days => %s)
+                    """,
+                    (org_id, retention_days, batch_size, retention_days),
+                )
+                deleted = int(cur.rowcount)
```
shiroyasha: I'm starting to hate that we don't have some better sql handler in the agent.
How hard would it be to use SQLAlchemy, seems to be the default in Python?

cc @lucaspin @0vertake

lucaspin: Not related to this pull request, but it does seem like something we should do

lucaspin: Created https://github.com/superplanehq/superplane/issues/4106 for this

agent/src/ai/chat_retention.py
0vertake: Sure, makes sense to simplify it. I'll add the dependency and implement it

PR #4067, opened 2026-04-09
feat: implement usage tracking queue to superplane agent

forestileao:
### Changes

Add token usage report through RabbitMQ message


forestileao:
adding tests...

agent/src/ai/usage_publisher.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import annotations
+
+import pika  # type: ignore[import-untyped]
+import pika.exceptions  # type: ignore[import-untyped]
+
+from ai.config import config
+from private import agents_pb2
+
+AGENT_EXCHANGE = "superplane.agent-exchange"
+AGENT_TOKENS_USED_ROUTING_KEY = "agent-tokens-used"
```
lucaspin: Since we send this when the agent run finishes, using a `agent-run-finished` routing key and message would probably be better than just sending the tokens.

forestileao: makes sense, let me update it

agent/src/ai/usage_publisher.py
```diff
@@ -0,0 +1,59 @@
...
+import pika.exceptions  # type: ignore[import-untyped]
+
+from ai.config import config
+from private import agents_pb2
+
+AGENT_EXCHANGE = "superplane.agent-exchange"
+AGENT_RUN_FINISHED_ROUTING_KEY = "agent-run-finished"
+
+
+def publish_agent_run_finished(
```
lucaspin: Having a Publisher class that reuses a connection would be better here

agent/src/ai/usage_publisher.py
```diff
@@ -46,50 +51,64 @@ def publish_agent_run_finished(
...
+        self._thread.join(timeout=5.0)
+
+    def _run(self) -> None:
+        """Dedicated publisher thread — owns the pika connection."""
+        connection: pika.BlockingConnection | None = None
+        channel: pika.adapters.blocking_connection.BlockingChannel | None = None
 
-    def _publish(self, body: bytes) -> None:
-        with self._lock:
+        while True:
```
forestileao: Uh..

lucaspin (approved):

PR #4069, opened 2026-04-09
fix: improve agent tool call animations for smoother UI

0vertake:
## Summary

- Stagger tool call appearances with a 150ms delay between each, preventing multiple tools from popping into the DOM at once
- Add a height + opacity CSS animation so each tool smoothly expands into place instead of causing a layout jump
- Collapse fast `tool_started`/`tool_finished` pairs in the queue so tools that complete before rendering skip the "running" flash
- Fix `tool_finished` events failing to match their `tool_started` message when the backend omits `tool_call_id` (`Date.now()` fallback produced mismatched IDs)
- Clean up elapsed time display: "326ms" instead of "326.0ms", "1.2s" for longer durations

## Demo Video

https://github.com/user-attachments/assets/2447a578-d294-49ab-920c-e89789571427

### DIFF
--- a/web_src/src/App.css
+++ b/web_src/src/App.css
@@ -244,3 +244,18 @@
   color: transparent;
   animation: sp-ai-thinking-shimmer 1.4s linear infinite;
 }
+
+@keyframes sp-tool-enter {
+  from {
+    opacity: 0;
+    transform: translateY(-4px);
+  }
+  to {
+    opacity: 1;
+    transform: translateY(0);
+  }
+}
+
+.sp-tool-enter {
+  animation: sp-tool-enter 0.2s ease-out both;
+}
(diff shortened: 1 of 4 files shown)


lucaspin:
@0vertake seems like the first tool calls that appear still have the problem?

0vertake:
@lucaspin Okay, turns out I misinterpreted the issue. I though the problem was the constant *back-and-forth* "jumping" of the tool calls. So the problem for this concrete prompt I shared is how when the first tool is called, the rest are also immediately called with their respective animations, causing the "jump". Is this right? I apologize in advance, it's kinda tough to put UI into words 😅

lucaspin:
@0vertake I think what you did with the animations is good, but it seems like for the first tool calls we show, there's still some jumps happening - the tool name is shown first at the bottom, then it moves to the right. For the subsequent tool call messages, that doesn't happen.

0vertake:
@lucaspin Is this good?

https://github.com/user-attachments/assets/de1bfb61-7da5-4264-b495-d4b59d0c4414

0vertake:
@lucaspin  Here is the final version after tool call messages were refactored

https://github.com/user-attachments/assets/0551c62d-560d-478d-a5f8-5e16360f1eab

shiroyasha (approved):

PR #4070, opened 2026-04-09
chore(refactor): use LayoutEngine in CanvasBuilder 

lucaspin:
Right now, after calling CanvasBuilder.apply(operations), caller still has to check if new nodes were added, and if so, apply layout changes. I want to move layout changes as part of CanvasBuilder.apply() automatically, so the caller doesn't worry about it.

### Main changes

- Move `workflowv2/autoLayout.ts` into `lib/layout`, putting layout logic behind a proper LayoutEngine interface.
- With that, we update `CanvasBuilder.apply()` signature to be accept a new optional LayoutEngine. If that is specified, and new nodes are added, builder itself will call the layout engine.


lucaspin:
👍

PR #4071, opened 2026-04-09
feat: add token limit

forestileao:
## Summary                                                                                 
                                                                                               
  - Python agent checks the token bucket against the usage service before each run,       
  blocking with a clear error when the limit is exceeded                                       

## The flow:

So the flow is: throw new Error("organization agent token limit exceeded") → caught at line  
  452 → applyChatPromptFailure which shows the error to the user in the chat UI as a failed
  message.                                                                                     
                                                                                               
  The user sees the error message "organization agent token limit exceeded" displayed in the   
  chat as a failure. The chat remains usable — they just can't send new messages until their   
  token bucket decreases (daily leak).


pkg/grpc/actions/agents/agent_token_limits_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	usagepb "github.com/superplanehq/superplane/pkg/protos/usage"
+	"github.com/superplanehq/superplane/pkg/usage"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type fakeUsageService struct {
```
lucaspin: Let's not have fake implementations of interfaces buried in the files like this, and instead follow the same pattern we use for [test/support/contexts.go](https://github.com/superplanehq/superplane/blob/main/test/support/contexts/contexts.go)

protos/usage.proto
```diff
@@ -89,6 +89,7 @@ message OrganizationLimits {
   int32 retention_window_days = 4;
   int64 max_events_per_month = 5;
   int32 max_integrations = 6;
+  int64 max_agent_tokens_per_month = 7;
```
lucaspin, line 92: This is not particularly related to this pull request, but I don't like that we are defining these limits like this, because if we wanted to change the period we are applying on a limit, we would need to change the API. We should have something like this instead:

```
type Limit {
  type: "rate" | "absolute" | "...";
  rate?: { value: number, period: number }
  absolute?: { value: number }
  // ... other types of limits
}
```

This allow me to change my limits without changing the API

pkg/grpc/actions/messages/agent_run_finished.go
```diff
@@ -0,0 +1,31 @@
...
+const (
+	AgentExchange              = "superplane.agent-exchange"
+	AgentRunFinishedRoutingKey = "agent-run-finished"
+)
+
+type AgentRunFinishedMessage struct {
+	message *pb.AgentRunFinishedMessage
+}
+
+func NewAgentRunFinishedMessage(organizationID, chatID, model string, inputTokens, outputTokens, totalTokens int64) AgentRunFinishedMessage {
```
lucaspin: Why do we have this here?

pkg/grpc/agent_service.go
```diff
@@ -71,6 +78,10 @@ func (s *AgentsService) ResumeAgentChat(ctx context.Context, req *pb.ResumeAgent
 		return nil, err
 	}
 
+	if err := agents.EnsureAgentTokensWithinLimits(ctx, s.usageService, organizationID); err != nil {
```
lucaspin: Shouldn't we enforce this limit on the agent service itself? Doing it here requires an extra gRPC call every time

protos/usage.proto
forestileao: We are using leaking bucket algorithm for this. PER MONTH is just a time reference to how much the "bucket leaks" (decreases).

It would be approximately the same as using a per week with a lower limit.

The only thing that is fixed is that we are leaking it daily (not sure it is the best way)

lucaspin: > PER MONTH is just a time reference to how much the "bucket leaks"

What do you mean? The field in your API message is called `max_agent_tokens_per_month`. If you change the period to weekly, you have to rename the field, no? And if you rename the field to just `max_agent_tokens`, then from the caller side, it's not clear what the period being used is

forestileao: Yes it would. what I wanted to say is:

100000 tokens limit PER MONTH ~= 25000 tokens limit per week
Saying that the period is just a time reference and we can change that in another PR.

pkg/grpc/agent_service.go
forestileao: I'll move it to the agent. It will be way better

pkg/grpc/actions/messages/agent_run_finished.go
forestileao: I'll remove it

shiroyasha:
Sanity check: Will this approach work in self-hosted installations?

forestileao:
@shiroyasha The same way it works for other limits today. It would require self hosted to implement their usage service.

if USAGE_GRPC_URL is not set, the lifespan creates NoopUsageLimitChecker which does  nothing.

There is another task though, that is allowing to use their own key. Then it would use their own limits, unless they use their own usage service.

lucaspin (approved):

PR #4083, opened 2026-04-10
chore(ui): cleanup BuildingBlock type

lucaspin:
The BuildingBlock type we have in the UI doesn't really match what we have in the backend, and it has some unused things:
- deprecated, exampleData, and exampleOutput are unused
- componentSubtype is used, but it is confusing. Backend has no concept of a subtype, and the subtype used in the UI is equal to the type in 2/3 of the cases. Makes no sense.

These problems seem small, but when you are trying to come up with a better registry API, they confuse you, so I'm just removing it.


lucaspin:
👍

PR #4089, opened 2026-04-11
chore: Remove deprecated agent repl

shiroyasha:
Just a round of code cleanup.
The repl was useful in the start, it is no longer useful and probably broken.
Also, we were calling out main server repl_web, instead of just web.py.


shiroyasha:
👍

PR #4105, opened 2026-04-13
fix: retry transient LLM errors and show friendly message

0vertake:
## Summary

- Retries transient LLM errors (429, 502, 503, 504, 529) up to 3 times with jittered exponential backoff before surfacing a failure
- Maps LLM provider errors to clean, user-friendly messages instead of exposing raw Python exception reprs
- Adds a frontend sanitization layer as defense-in-depth to catch any unsanitized errors
- Always logs the original error server-side for production debugging

## Test plan
- 26 backend unit tests covering _extract_status_code, _is_transient_error, _friendly_error_message, and _run_stream_events retry behavior (including the mid-stream guard)
- 7 frontend unit tests covering sanitizeErrorMessage for all rejection conditions
- Manual smoke test: injected a fake 529 error inside _run_stream_events, confirmed retries fire with backoff in agent logs and only a clean message reaches the UI


0vertake:
@lucaspin So I came across 2 possible issues that are not introduced by this PR:
1. Error flash on new chats: When the **first** prompt in a new chat fails, the error briefly flashes then disappears.
2. Duplicate error display: On subsequent prompts, both a generic assistant reply ("I couldn't generate changes right now") and the red `aiError` text are shown. These could be possibly be squashed into one message.


https://github.com/user-attachments/assets/0d1eb29c-ddd9-40eb-bbff-b2edfe3a98b9

lucaspin:
@0vertake how are you testing this change?

0vertake:
> @0vertake how are you testing this change?

My bad I should've included the test plan in the PR description, since this is a tricky one. Besides the unit tests, I tested this by hardcoding a 529 error inside `web.py`. Then I tried using the agent and confirmed the retries happened in the agent logs, as shown here:
```
agent-1  | [web] transient LLM error, retrying attempt=1 delay=1.4438241113218107 error=simulated overload
agent-1  | [web] transient LLM error, retrying attempt=2 delay=2.235779114462156 error=simulated overload
agent-1  | [web] stream failed chat_id=263a4cb2-0720-47b7-a783-5ec5540f67c4 error=simulated overload
```

lucaspin (approved):

PR #4107, opened 2026-04-13
chore: remove canvas user drafts table in favor of version state column

forestileao:
## Summary

  - Adds a `state` column (`draft`/`published`) to the
  `workflow_versions` table with a partial unique index
  on `(workflow_id, owner_id) WHERE state = 'draft'`,
  enforcing one draft per user per canvas at the database
   level
  - Drops the `workflow_user_drafts` table — draft
  tracking now lives entirely on the version row
  - Simplifies model functions (`FindCanvasDraft*`,
  `SaveCanvasDraft*`, `PublishCanvasDraft*`) to
  query/update versions directly instead of joining
  through a separate table
  - Simplifies action code by removing the extra version
  lookup after finding a draft (the draft *is* the
  version now)

  ## Motivation

  The `workflow_user_drafts` table was a join table
  mapping `(workflow_id, user_id) → version_id`. This
  added indirection and complexity for something that can
   be expressed as a column + partial unique index on the
   versions table itself.


db/migrations/20260413120000_remove-canvas-user-drafts.up.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+-- Add state column to workflow_versions.
+-- Existing rows default to 'published'.
+ALTER TABLE public.workflow_versions
+  ADD COLUMN IF NOT EXISTS state character varying(32) NOT NULL DEFAULT 'published';
```
forestileao: default should never be published

pkg/models/canvas_version.go
```diff
@@ -218,111 +215,35 @@ func SaveCanvasDraftInTransaction(
...
-	nodes []Node,
-	edges []Edge,
-) (*CanvasVersion, error) {
-	canvas, err := lockCanvasForVersioningInTransaction(tx, workflowID)
-	if err != nil {
-		return nil, err
-	}
-
-	if canvas.LiveVersionID == nil {
-		return nil, gorm.ErrRecordNotFound
```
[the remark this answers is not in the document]
forestileao: we should remove published_at + is_published

protos/canvases.proto
```diff
@@ -569,14 +569,19 @@ message Canvas {
...
     string id = 1;
     string canvas_id = 2;
     UserRef owner = 4;
-    bool is_published = 6;
-    google.protobuf.Timestamp published_at = 7;
-    google.protobuf.Timestamp created_at = 8;
-    google.protobuf.Timestamp updated_at = 9;
+    State state = 5;
+    google.protobuf.Timestamp created_at = 6;
+    google.protobuf.Timestamp updated_at = 7;
```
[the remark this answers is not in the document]
forestileao: not an issue

web_src/src/hooks/useCanvasData.ts
```diff
@@ -267,7 +266,7 @@ type PositionedNode = ComponentsNode & {
 };
 
 const versionSortTimestamp = (version: CanvasesCanvasVersion): number => {
-  const raw = version?.metadata?.updatedAt || version?.metadata?.createdAt;
+  const raw = version?.metadata?.publishedAt || version?.metadata?.updatedAt || version?.metadata?.createdAt;
```
shiroyasha: Do we need all three fallbacks here? Is it even possible to have publishedAt null in this context?

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -256,8 +256,8 @@ export function WorkflowPageV2() {
   const liveVersions = useMemo(
     () =>
       visibleCanvasVersions
-        .filter((version) => version.metadata?.state === "STATE_PUBLISHED")
-        .sort((a, b) => versionSortValue(b.metadata?.updatedAt) - versionSortValue(a.metadata?.updatedAt)),
+        .filter((version) => !!version.metadata?.publishedAt)
+        .sort((a, b) => versionSortValue(b.metadata?.publishedAt) - versionSortValue(a.metadata?.publishedAt)),
     [visibleCanvasVersions],
   );
```
shiroyasha: Lets extract this to some other file. 
Complex logic, doesn't need to live in this big component.

shiroyasha (dismissed):
Overall, looks good. Left some minor non-blocker comments.

forestileao:
> Overall, looks good. Left some minor non-blocker comments.

Thanks a lot @shiroyasha . Going to solve them and merge

forestileao:
👍

forestileao:
@shiroyasha I will need another approval (mine is not enough)

shiroyasha (approved):

PR #4112, opened 2026-04-14
feat: add prompt aborting in agent chat

### LINKED ISSUE #4076: Agent: Allow me to stop a running prompt

abtonc:
## What Changed

This PR adds prompt cancellation support to the Agent chat UI in the workflow sidebar.

- Added a "Stop" action while a response is generating (square icon replaces send button while running, also kind of works like a "Steer" feature.).
- Introduced request cancellation via `AbortController` for the agent stream request and added explicit abort handling so canceled runs do not show an error and instead show `Stopped.`.
- Prevented new chat session selection updates when a run is aborted.
- Added cleanup to cancel in-flight agent streams on canvas and component unmount.

## How

- Added an `onStopPrompt` callback next to `onSendPrompt`.
- Added `chatAbortControllerRef` state in sidebar scope and passed it into `sendChatPrompt`.
- Added abort-aware error detection (`AbortError`) and a dedicated cancellation UI path.
- Added `stopRunningChatPrompt(...)` helper:
  - explicit stop button click
  - canvas change reset
  - component unmount cleanup

## Related Issues

Closes #4076 

Let me know if anything needs to be changed. Much love.

### DIFF
--- a/web_src/src/components/AiBuilderInputForm.tsx
+++ b/web_src/src/components/AiBuilderInputForm.tsx
@@ -1,6 +1,6 @@
 import { Textarea } from "@/components/ui/textarea";
 import { cn } from "@/lib/utils";
-import { ArrowUp } from "lucide-react";
+import { ArrowUp, Square } from "lucide-react";
 import type { FormEvent, KeyboardEvent, RefObject } from "react";
 
 export type InputFormProps = {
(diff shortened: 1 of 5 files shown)


lucaspin:
This is about the old agent implementation. Closing it

PR #4114, opened 2026-04-14
fix: don't crash public API when usage client init fails

super-stackweaver:
Prevent the public API from crashing when the usage gRPC client can’t be initialized.
Fall back to a disabled usage service and log a warning instead.


shiroyasha:
Thanks for opening the PR.
This is not the direction we would like to take this fix.
It is better for the app to fail at start than to run with a wrong service.

PR #4120, opened 2026-04-14
feat: wire Sentry to track agent-related errors

re-pixel:
- Add Sentry exception tracking to the Python agent service, following the same pattern used by the Go backend (`pkg/telemetry/sentry.go`): reads **SENTRY_DSN** and **SENTRY_ENVIRONMENT**, no-op when DSN is empty, flushes on shutdown.
- Exceptions in the SSE stream handler are captured via `sentry_sdk.capture_exception()`.
- Sentry is disabled by default in dev (empty DSN in docker-compose).


shiroyasha:
👍

PR #4121, opened 2026-04-14
feat: new endpoint for applying a changeset on a canvas version

lucaspin:
New `PATCH /api/v1/canvases/{canvas_id}/versions/{version_id}` for applying a changeset on a canvas version. It supports a dry_run flag, for validation purposes.

### Main changes

Most of the changes here are protos and SDKs being updated. The main change is:
- New endpoint definition in [canvases.proto](https://github.com/superplanehq/superplane/pull/4121/changes#diff-4a00843d6f945de3d8e45e54c2da4606937114ea29666432a22f9d2afa90aff1)
- Its implementation in [pkg/grpc/actions/canvases/apply_canvas_version_changeset.go](https://github.com/superplanehq/superplane/pull/4121/changes#diff-97bf3dc5af0a8fa0ad42fe2bb91399d718de4e72bce222de88391bebf015f3dcR20)
- New `grpc/actions/canvases/changesets` package. All the logic for applying a changeset on a canvas version is on the new [CanvasPatcher](https://github.com/superplanehq/superplane/pull/4121/changes#diff-86cc3f4055bd17caa930cfec1b4e8669323bb6280dcf4e8bbd51c5206ccef23cR13).


pkg/grpc/actions/canvases/changesets/canvas_patcher.go
```diff
@@ -0,0 +1,375 @@
...
+
+func (p *CanvasPatcher) findEdge(sourceID, targetID, channel string) (int, bool) {
+	index := slices.IndexFunc(p.canvas.Edges, func(edge models.Edge) bool {
+		return edge.SourceID == sourceID &&
+			edge.TargetID == targetID &&
+			edge.Channel == channel
+	})
+
+	return index, index >= 0
+}
```
forestileao: I think this could become a performance bottleneck as the dataset grows. We're doing linear scans (IndexFunc) for every lookup, which can lead to O(n²) behavior across multiple operations.

Since we can update many nodes and that may delay a bit the endpoint. (Not so much but it can)

It might be worth introducing maps to index nodes and edges (e.g. by ID or a composite key) so lookups are O(1) and the overall complexity stays closer to O(n).

like:

```go

type CanvasPatcher struct {
    canvas    *models.Canvas
    nodeIndex map[string]int
    edgeIndex map[string]int
}

func edgeKey(sourceID, targetID, channel string) string {
    return sourceID + "|" + targetID + "|" + channel
}

func (p *CanvasPatcher) findNode(nodeID string) (int, bool) {
    idx, ok := p.nodeIndex[nodeID]
    return idx, ok
}

func (p *CanvasPatcher) findEdge(sourceID, targetID, channel string) (int, bool) {
    idx, ok := p.edgeIndex[edgeKey(sourceID, targetID, channel)]
    return idx, ok
}
```

pkg/grpc/actions/canvases/changesets/canvas_patcher.go
```diff
@@ -0,0 +1,375 @@
...
+			return fmt.Errorf("duplicate node id: %s", node.ID)
+		}
+
+		nodeIDs[node.ID] = true
+		adjacency[node.ID] = []string{}
+		inDegree[node.ID] = 0
+	}
+
+	for _, edge := range p.canvas.Edges {
+		if edge.SourceID == "" || edge.TargetID == "" {
```
forestileao: Don't we have this logic in the current update_canvas_version? when validating the nodes, etc? 

Just to check if it is a validation that could be shared to avoid duplicated logic.

pkg/grpc/actions/canvases/changesets/canvas_patcher.go
```diff
@@ -0,0 +1,412 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		return widget.Configuration(), nil
+
+	default:
+		return nil, fmt.Errorf("unknown node type: %s", nodeType)
+	}
+}
```
shiroyasha, line 275: non-blocking comment:

Feels like this could be extracted to the registry.GetNode(type, name).

shiroyasha (dismissed):

pkg/grpc/actions/canvases/changesets/canvas_patcher.go
```diff
@@ -0,0 +1,412 @@
...
+		nodeConfiguration = node.GetConfiguration().AsMap()
+	}
+
+	p.canvas.Nodes = append(p.canvas.Nodes, models.Node{
+		ID:            nodeID,
+		Name:          node.GetName(),
+		Type:          nodeType,
+		Ref:           *nodeRef,
+		Configuration: nodeConfiguration,
+	})
```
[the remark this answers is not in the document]
lucaspin: addNode does not need to consider metadata. Positioning will be handled in separate pull request, because using the layout logic we already have in the patcher requires refactoring it.

pkg/grpc/actions/canvases/changesets/canvas_patcher.go
lucaspin: Good catch. There was a CheckForCycles(). Re-used it

lucaspin:
👍

PR #4133, opened 2026-04-14
chore: Remove dead code associated with disabled auto-save state

shiroyasha:
Disabled auto-save no longer exists as a concept in the app.

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -603,7 +603,6 @@ export function WorkflowPageV2() {
   const [isPositionAutoSaveQueued, setIsPositionAutoSaveQueued] = useState(false);
   const [isAnnotationAutoSaveQueued, setIsAnnotationAutoSaveQueued] = useState(false);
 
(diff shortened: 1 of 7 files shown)


web_src/src/pages/workflowv2/index.tsx
```diff
@@ -1433,8 +1423,8 @@ export function WorkflowPageV2() {
 
   /**
    * Debounced auto-save function for node position changes.
-   * Waits 100ms after the last position change when auto-save is enabled,
-   * or 5s when auto-save is disabled to avoid disrupting editing.
+   * Uses a short delay while the canvas is editable; a longer delay when read-only
+   * so stray timers from a mode switch do not fire too aggressively.
```
shiroyasha, line 1427: tbh, I don't understand neither this, not the previous.

forestileao: that was one of the fixes for that flickering while dragging nodes + auto-saving. 
As far as I remember

forestileao (approved):

PR #4134, opened 2026-04-14
feat: add publish draft canvas version endpoint

forestileao:
## Summary

  - Adds PATCH /api/v1/canvases/{canvas_id}/versions/{version_id}/publish to
  publish a draft version directly to live without going through the change
  request flow.
  - Extracts shared node-application logic (applyCanvasSpecInTransaction) to
  avoid duplication between this endpoint and the existing change request publish
   flow.

## Changes

  New endpoint — publish_canvas_version.go
  - Validates the version is a draft owned by the caller and that versioning is
  enabled
  - Applies the draft spec to the live canvas (remap, expand, upsert, setup,
  delete nodes) — same logic used by change request publish
  - Creates a new published version and deletes the old draft
  - Uses SELECT FOR UPDATE via FindCanvasVersionForUpdateInTransaction to prevent
   race conditions

 ### Extracted shared logic — apply_canvas_spec.go
  - applyCanvasSpecInTransaction — the common node-application pipeline extracted
   from PublishCanvasChangeRequest
  - Broken into small helpers (applyNode, setParentNodeError, markNodeSetupError)
   to reduce nesting

 ### Refactored publish_canvas_change_request.go
  - Replaced ~80 lines of inline node-application code with a call to
  applyCanvasSpecInTransaction


pkg/grpc/actions/canvases/publish_canvas_version.go
```diff
@@ -0,0 +1,147 @@
...
+		var createErr error
+		publishedVersion, createErr = models.CreatePublishedCanvasVersionInTransaction(
+			tx, canvasUUID, version.OwnerID, nodes, edges,
+		)
+		if createErr != nil {
+			return createErr
+		}
+
+		return tx.Delete(version).Error
+	})
```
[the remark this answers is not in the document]
forestileao: well pointed

pkg/grpc/actions/canvases/publish_canvas_version.go
```diff
@@ -0,0 +1,151 @@
...
+
+		var createErr error
+		publishedVersion, createErr = models.CreatePublishedCanvasVersionInTransaction(
+			tx, canvasUUID, version.OwnerID, nodes, edges,
+		)
+		if createErr != nil {
+			return createErr
+		}
+
+		if err := tx.Delete(version).Error; err != nil {
```
lucaspin: Why are we deleting the version?

forestileao: this is wrong and it was using the same CR behavior or duplicating a version to continue editing. That is not what we want. Thanks for noticing it

forestileao: @lucaspin Now this is fixed 

Direct publish now transitions the draft version in place (same row) instead of creating a new published version and deleting the draft

lucaspin (approved):

shiroyasha (approved):

PR #4139, opened 2026-04-14
feat: split changeset validation to a new endpoint

lucaspin:
Having the changeset validation behind a dry_run flag makes it impossible for a read-only agent to use it for validating its answer. Here, we split the changeset validation into a new endpoint:
- `PATCH /api/v1/canvases/{canvas_id}/versions/{version_id}` -> remains as the endpoint for applying a changeset, requires `canvases:update` permission
- `PATCH /api/v1/canvases/{canvas_id}/versions/{version_id}/validate` -> new endpoint for validating a changeset, requires `canvases:read` permission, so the agent can use it for validating its answer


lucaspin:
👍

forestileao (approved):

PR #4140, opened 2026-04-14
feat: Agent is aware on which screen are we using it

shiroyasha:
- On the live canvas = agent is in inspect mode
- On the editor = agent is in builder mode

In this PR I focus only on sending down this info. Further prompt optimizations will be necessary.

### DIFF
--- a/agent/src/ai/tools/get_canvas.py
+++ b/agent/src/ai/tools/get_canvas.py
@@ -15,10 +15,12 @@ class GetCanvas:
     @staticmethod
     def run(ctx: RunContext[AgentDeps]) -> CanvasSummary:
         resolved_canvas_id = ctx.deps.canvas_id
-        cached_summary = ctx.deps.canvas_cache.get(resolved_canvas_id)
+        version_id = ctx.deps.editing_version_id
+        cache_key = f"{resolved_canvas_id}:{version_id or 'live'}"
+        cached_summary = ctx.deps.canvas_cache.get(cache_key)
         if cached_summary is not None:
             return cached_summary
 
-        summary = ctx.deps.client.describe_canvas(resolved_canvas_id)
-        ctx.deps.canvas_cache[resolved_canvas_id] = summary
+        summary = ctx.deps.client.describe_editing_canvas(resolved_canvas_id, version_id)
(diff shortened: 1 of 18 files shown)


agent/src/ai/tools/get_canvas.py
```diff
@@ -15,10 +15,12 @@ def label(ctx: RunContext[AgentDeps]) -> str:
     @staticmethod
     def run(ctx: RunContext[AgentDeps]) -> CanvasSummary:
         resolved_canvas_id = ctx.deps.canvas_id
-        cached_summary = ctx.deps.canvas_cache.get(resolved_canvas_id)
+        version_id = ctx.deps.canvas_version_id
+        cache_key = f"{resolved_canvas_id}:{version_id or 'live'}"
+        cached_summary = ctx.deps.canvas_cache.get(cache_key)
```
shiroyasha, line 20: not for this PR but writing it down:
Do we even need this cache?

agent/src/ai/web.py
```diff
@@ -338,7 +361,9 @@ async def _stream_agent_run(
         "canvas_id": resolved_canvas_id,
     }
 
-    run_kwargs: dict[str, Any] = {"user_prompt": payload.question}
+    _, prompt_surface = derive_agent_editor_fields(payload.agent_context)
```
shiroyasha: Wtf is prompt surface?

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -122,18 +114,31 @@ import {
...
+  buildUserInfo,
+  generateNodeId,
+  generateUniqueNodeName,
+  mapCanvasNodesToLogEntries,
+  mapExecutionsToSidebarEvents,
+  mapQueueItemsToSidebarEvents,
+  mapTriggerEventsToSidebarEvents,
+  mapWorkflowEventsToRunLogEntries,
+  summarizeWorkflowChanges,
+} from "./utils";
```
shiroyasha, line 141: My editor auto-reorganizes these imports.
Sucks that make format.js is ignoring this, it should be the same for everyone.

forestileao (dismissed):

lucaspin (dismissed):

PR #4143, opened 2026-04-15
chore: replace versioning_enabled with change_management_enabled

forestileao:
## Summary

  - **Versioning is now always enabled** — every      
  canvas uses the draft/edit/publish flow. The
  `versioning_enabled` flag is removed from both      
  organizations and canvases.                       
  - **Change management (change requests with       
  approvals) is now optional** and controlled by a new
   `change_management_enabled` flag, disabled by
  default.                                            
  - When change management is **disabled**: users edit
   drafts and publish directly via a "Publish" button.
  - When change management is **enabled**: users edit
  drafts and submit change requests for approval via a
   "Propose Change" button.                         
                                                      
  ## Changes                                        
                                                    
  ### Database
  - Migration renames `versioning_enabled` →
  `change_management_enabled` in `organizations` and  
  `workflows` tables.
                                                      
  ### Backend (Go)                                  
  - Models: `VersioningEnabled` →                   
  `ChangeManagementEnabled` on `Organization` and     
  `Canvas`.
  - Removed versioning checks from                    
  `publish_canvas_version`, `create_canvas_version`,  
  and `update_canvas_version` (versioning is always 
  on).                                                
  - Removed `updateLiveCanvasWithoutVersioning` —   
  version ID is always required.                      
  - Change request endpoints (`create`, `act_on`,
  `publish`, `resolve`) now check                     
  `change_management_enabled` instead of            
  `versioning_enabled`.                               
  - Updated `versioning.go` helpers to              
  `isChangeManagementEnabledForCanvas`.               
  - Updated serialization, canvas creation, org
  update, and CLI commands.                           
                                                    
  ### Frontend                                        
  - Removed `versioning-disabled` header mode — header
   always shows version-live/version-edit.            
  - Header publish button label is dynamic: "Publish"
  (direct) or "Propose Change" (change requests).     
  - Org settings: "Canvas Versioning" toggle → "Change
   Management" toggle.                                
  - Canvas settings page: same rename, approver config
   only shown when change management is effective.    
  - Removed obsolete `CanvasSettingsView.tsx`       
  (replaced by `pages/canvas/settings/`).             
                                                    
  ### Proto                                           
  - `versioning_enabled` → `change_management_enabled`
   in `Canvas.Metadata`, `UpdateCanvasRequest`, and   
  `Organization.Metadata`.
                                                      
  ### Tests                                           
  - Updated all unit and E2E tests for the rename.  
  - Removed tests for disabled-versioning paths that  
  no longer exist.


db/migrations/20260414233443_rename-versioning-enabled-to-change-management-enabled.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+ALTER TABLE workflows DROP COLUMN versioning_enabled;
+ALTER TABLE organizations DROP COLUMN versioning_enabled;
+
+ALTER TABLE organizations ADD COLUMN change_management_enabled boolean NOT NULL DEFAULT false;
+ALTER TABLE workflows ADD COLUMN change_management_enabled boolean NOT NULL DEFAULT false;
+
+COMMIT;
```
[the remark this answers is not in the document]
forestileao: that is the idea

protos/canvases.proto
```diff
@@ -433,7 +433,7 @@ message UpdateCanvasRequest {
   string id = 1;
   optional string name = 2;
   optional string description = 3;
-  optional bool versioning_enabled = 4;
+  optional bool change_management_enabled = 4;
```
lucaspin: These things shouldn't be in the metadata, right? They are things you can change, so should be on the spec.

Also, we should collapse configuration for related things under one common field. Right now, YAML would look like this:

```yaml
metadata:
  changeManagementEnabled: true
  changeRequestApprovalConfig:
    items:
    - type: TYPE_USER
      userId: "123"
    - type: TYPE_ROLE
      roleName: "admin"
```

I think this would look a bit better, for example:

```yaml
spec:
  changeManagement:
    enabled: true
    approvals:
      - type: TYPE_USER
        userId: "123"
      - type: TYPE_ROLE
        roleName: "admin"
```

So, we could have this proto definition:

```
message Canvas {
  message ChangeManagement {
    message ApproverConfig {
      enum Type {
        TYPE_UNSPECIFIED = 0;
        TYPE_ANYONE = 1;
        TYPE_USER = 2;
        TYPE_ROLE = 3;
      }

      Type type = 1;
      string user_id = 2;
      string role_name = 3;
    }

    bool enabled = 1;
    ApproverConfig approvals = 2;
  }

  message Spec {
    ChangeManagement change_management = 1;
  }
}
```

forestileao: Excelent review. Doing it right now :)

forestileao: @lucaspin done!

test/e2e/shared/canvas_steps.go
```diff
@@ -29,11 +29,67 @@ func NewCanvasSteps(name string, t *testing.T, session *session.TestSession) *Ca
 }
 
 func (s *CanvasSteps) WaitForCanvasSaveStatusSaved() {
-	// ugly, but it is what it is for now
-	// TODO: remove this completely once edit mode is shipped
 	s.session.Sleep(2500)
```
lucaspin: You removed the comment, but the sleep is still there? cc @shiroyasha

forestileao: Let me check if we still can remove it... (we should once we click on publish)

forestileao: @lucaspin GOOD CATCH. everyting worked after the removal

lucaspin (approved):

PR #4152, opened 2026-04-15
fix: changeset contains information about integration

lucaspin:
When adding an integration-based node using a changeset, we should be able to choose the integration the node will use. See [this Discord message](https://discord.com/channels/1409914582239023200/1493187148604506252/1493943366616875028) for more details


lucaspin:
👍

PR #4160, opened 2026-04-15
feat: auto layout when applying changesets

lucaspin:
Same thing we have in UpdateCanvasVersion, but for the new endpoint that applies changesets. Moved layout code into `pkg/grpc/actions/canvases/layout` to use in CanvasPatcher too.


lucaspin:
👍

PR #4166, opened 2026-04-15
chore(refactor): AuthContext -> AuthReader

lucaspin:
Separating core interfaces between readers / writers is showing to be a better idea than having one for everything. Starting this movement with the AuthContext one, which is already a reader by its nature.


lucaspin:
👍

PR #4167, opened 2026-04-15
chore: split MetadataContext into reader and writer

lucaspin:
This allows us to control on the interface-level which level of access to metadata a component/trigger implementation method has.


lucaspin:
👍

PR #4168, opened 2026-04-15
chore: ignore generated pb files

forestileao:


forestileao:
Related to bug bot comments: 

- The `pb.gen` ownership comment is a false positive. The first three commands in [Makefile](/Users/pedroforestileao/Documents/code/superplane/Makefile:358) do run in root-owned dev images, but `gofmt -w` at [Makefile](/Users/pedroforestileao/Documents/code/superplane/Makefile:363) does not need write permission on the existing file; it rewrites via temp file + rename in the containing directory. I verified that behavior directly with a root-owned `644` Go file in a user-owned directory, and `gofmt -w` succeeded.

lucaspin (approved):

PR #4171, opened 2026-04-15
feat: Apply visuals to the secondary header row

shiroyasha:
I'm removing the CLI tab from the header.
When that was removed, the switch with just "canvas" didn't made any sense. Removed that as well.

CleanShot 2026-04-16 at 00 23 54@2x


forestileao (dismissed):

web_src/src/ui/CanvasPage/useCanvasState.ts
```diff
@@ -2,11 +2,8 @@ import type { Edge, EdgeChange, Node, NodeChange, NodePositionChange } from "@xy
 import { applyEdgeChanges, applyNodeChanges } from "@xyflow/react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import type { CanvasPageProps } from ".";
-import type { BreadcrumbItem } from "../../components/Breadcrumbs";
```
forestileao, line 5: just need to remove breadcumps dead file

PR #4172, opened 2026-04-15
fix: propagate caller context in LocalProvider.Load

JelenaMarjanovic:
## What
Forward the caller's `context.Context` from `LocalProvider.Load` to `encryptor.Decrypt`. Previously, `context.TODO()` was passed instead, dropping any cancellation, deadlines, and values (tracing, request IDs) from the caller.

## Why
`LocalProvider.Load` has the correct signature (`func (p *LocalProvider) Load(ctx context.Context)`), but the received `ctx` was ignored at the decryption boundary:

```go
decrypted, err := p.encryptor.Decrypt(context.TODO(), p.record.Data, []byte(name))
```

Consequences in production:

- A client disconnect mid-request cannot cancel an in-flight decrypt operation (wasted work, slower shutdown under load).
- Request-scoped telemetry (trace/span IDs, request IDs) carried via `context.Value` never reaches the encryptor or anything it calls.
- Future deadline-based rate limiting or timeout middleware in `crypto.Encryptor` implementations would silently be bypassed for `LocalProvider`.

## How
Single-line fix: replace `context.TODO()` with the caller's `ctx`.

Added `pkg/secrets/local_provider_test.go` (the package previously had no tests for `LocalProvider`) covering:

- **context value propagation** — asserts that a value placed in the caller's context reaches the encryptor.
- **context cancellation propagation** — asserts that a cancelled caller context is observed by the encryptor (`context.Canceled`).
- **happy path** — successful decrypt + JSON unmarshal.
- **decryption error wrapping** — verifies the existing `"error decrypting secret %s"` wrapping.
- **JSON unmarshal error wrapping** — verifies the existing `"error unmarshaling secret %s"` wrapping.

The tests use a small `contextCapturingEncryptor` mock (implements `crypto.Encryptor`) to record and assert on the context seen by the encryptor without touching the database.

## Verification
- `make test PKG_TEST_PACKAGES=./pkg/secrets` — 6/6 pass
- `make lint` — clean
- `make format.go` — no changes (file was already gofmt-clean)
- `make check.build.app` — binary builds
- `make test` (full backend suite) — 5797 pass, 0 failures

### DIFF
--- a/pkg/secrets/local_provider.go
+++ b/pkg/secrets/local_provider.go
@@ -26,7 +26,7 @@ func NewLocalProvider(tx *gorm.DB, encryptor crypto.Encryptor, record *models.Se
 
 func (p *LocalProvider) Load(ctx context.Context) (map[string]string, error) {
 	name := p.record.Name
-	decrypted, err := p.encryptor.Decrypt(context.TODO(), p.record.Data, []byte(name))
+	decrypted, err := p.encryptor.Decrypt(ctx, p.record.Data, []byte(name))
 	if err != nil {
 		return nil, fmt.Errorf("error decrypting secret %s: %v", name, err)
 	}
--- a/pkg/secrets/local_provider_test.go
+++ b/pkg/secrets/local_provider_test.go
@@ -1,0 +1,118 @@
+package secrets
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/superplanehq/superplane/pkg/crypto"
+	"github.com/superplanehq/superplane/pkg/models"
+)
+
+// contextCapturingEncryptor records the context passed to Decrypt so tests
+// can assert that LocalProvider forwards the caller's context instead of
+// substituting context.TODO() or context.Background().
+type contextCapturingEncryptor struct {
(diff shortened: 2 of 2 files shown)


pkg/secrets/local_provider.go
```diff
@@ -26,7 +26,7 @@ func NewLocalProvider(tx *gorm.DB, encryptor crypto.Encryptor, record *models.Se
 
 func (p *LocalProvider) Load(ctx context.Context) (map[string]string, error) {
 	name := p.record.Name
-	decrypted, err := p.encryptor.Decrypt(context.TODO(), p.record.Data, []byte(name))
+	decrypted, err := p.encryptor.Decrypt(ctx, p.record.Data, []byte(name))
```
shiroyasha, line 29: Nice catch!

Sanity check for us: Looking at the previous code, it feels like we intentionally left it on `context.TODO()`. Is there a deeper reason why would that be preferable? 

@lucaspin 🤔

lucaspin: I can't think of any. Probably an oversight on our part

shiroyasha:
@JelenaMarjanovic thank you for contributing to SuperPlane and congrats on first PR merged! <3

PR #4174, opened 2026-04-16
fix: agent not able to list integration resources

lucaspin:
The ListResources endpoint is a bit special, and the OpenAPI-based SDK does not handle it properly. See https://github.com/superplanehq/superplane/pull/2891 for more details. Due to that, the agent has not been able to list integration resources, getting a `400 resource type is required` error every time. To work around it, we form the URL directly instead of relying on the generated SDK method.


lucaspin:
👍

PR #4175, opened 2026-04-16
fix: image builds

forestileao:
Release jobs clone a clean checkout and build images directly. Since generated protobuf/OpenAPI artifacts are no longer committed, the image builds must generate them before building. Without this, the app image fails during go build because pkg/protos/... packages are missing, and the agent image has the same issue for generated Python client/protobuf files.

### DIFF
--- a/release/superplane-image/build.sh
+++ b/release/superplane-image/build.sh
@@ -20,12 +20,14 @@ IMAGE_REPO="${STANDARD_IMAGE_REPO:-ghcr.io/superplanehq/superplane}"
 
 echo "Building SuperPlane image (${IMAGE_REPO})"
 
+make gen.setup.backend
+make gen.setup.ui
+
 docker buildx build \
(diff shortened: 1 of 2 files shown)


forestileao:
👍

release/superplane-image/build.sh
```diff
@@ -20,8 +20,10 @@ IMAGE_REPO="${STANDARD_IMAGE_REPO:-ghcr.io/superplanehq/superplane}"
 
 echo "Building SuperPlane image (${IMAGE_REPO})"
 
+make gen.setup.backend
+make gen.setup.ui
+
 docker buildx build \
-  --platform "linux/${ARCH}" \
```
shiroyasha, line 24: Why did we delete this?

forestileao: My mistake for testing :)

shiroyasha (approved):

PR #4178, opened 2026-04-16
fix: finalize circuit-broken agent runs through usage accounting

orange-dot:
## Summary
This fixes the circuit-break finalization path so agent runs record and publish usage through the same run-level accounting surface used by the normal completion path.

## Why
Circuit-broken runs can terminate after real model usage has already accumulated. If that path does not finalize through the same run-level accounting source as normal completion, the recorded totals can drift from the actual run state.

## What changed
- read usage from `agent_run.usage()` in the web agent flow
- use that run-level usage in the circuit-break finalization path
- align the normal completion path to the same source for consistency
- keep the supporting loop-guard and discovery-tool behavior changes in the same PR because they belong to the same runtime-correctness surface
- add focused regression coverage for the circuit-break accounting path

## Scope guardrails
- no schema or API changes
- no protocol or wire-format changes
- limited to agent runtime finalization/accounting behavior and the directly related guard behavior needed to exercise it cleanly

## Validation
- isolated `agents_test` DB recreate plus agent migrations
- `docker compose -f docker-compose.dev.yml run --rm -e DB_NAME=agents_test agent uv run --group dev python -m pytest tests/test_agent_http.py tests/test_tool_loop_guard.py tests/test_list_integration_resources.py`
- `docker compose -f docker-compose.dev.yml run --rm -e DB_NAME=agents_test agent uv run --group dev python -m pytest tests`

Observed results:
- focused regression suite passed: `12 passed`
- full agent suite passed: `172 passed`
- the signed-off single-commit snapshot was rerun as-is after metadata cleanup

## Request for review
Would appreciate a close look at whether this keeps the accounting/finalization contract tight enough around circuit-broken runs without widening scope beyond the runtime paths that actually participate in usage publication.


orange-dot:
Observation from local follow-up on April 16,
  2026:

  The public Semaphore run failed on
  `TestSendEmailComponent/
  running_send_email_in_a_canvas_flow` with `timed
  out waiting for execution of node Send Email`.

  I ran the same targeted test locally in an
  isolated compose setup against:
  - PR head
  `fb75cd0e710847523b406cf3dfe274e08eb645be`
  - PR merge ref
  `e734a2d5724d0b1ea9fd212692b7437c45df59ab`
  - local `main` baseline
  `62671f252985284d6d468e07a10b58d05fe5fd86`

  All three passed locally.

  Not claiming the PR is clean from this alone,
  only noting that I could not reproduce the
  current Semaphore `Send Email` failure locally
  on either the head or merge snapshot.

PR #4183, opened 2026-04-16
chore: update bottom toolbar

forestileao:
## Summary

- remove the global compact toggle from the canvas bottom toolbar
- show auto-layout and snap-to-grid only in edit mode
- show runs controls only in live mode, while keeping errors and warnings in both modes
- move the versions control into the header action area on live mode
- keep the versions control as an icon-only toggle with tooltip and active state
- restore draft warnings in edit mode logs sidebar

## Live Canvas

image

## Edit Mode

image


forestileao:
👍

PR #4184, opened 2026-04-16
chore: remove editing draft toast when creating a new one

forestileao:
## Change

remove editing draft toast when creating a new one


forestileao:
👍

PR #4185, opened 2026-04-16
fix: keep canvas publish and discard enabled during draft autosave

forestileao:
## Summary

Fix CanvasPage version actions so local draft/save state does not disable `Publish` or `Discard`.

## Changes

- keep `Publish` enabled even while autosave or version update work is in progress
- keep `Publish` enabled even when local draft changes are present
- keep `Discard` enabled instead of disabling during local save activity
- remove the misleading `Save your version before publishing.` disabled-state path
- update publish/change-request actions to:
  - flush pending autosave work
  - attempt a silent save of the latest workflow snapshot when needed
  - wait briefly for the draft to settle before continuing

## Result

Users can publish from their latest draft changes without the button flipping disabled/enabled during autosave, and discard is no longer incorrectly blocked by local save activity.

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -114,6 +114,7 @@ import {
 import { buildChangeRequestVersionRowsForStatus } from "./lib/change-requests";
 import { getNodeIntegrationName, overlayIntegrationWarnings } from "./lib/node-integrations";
 import { renderCanvasNodeCustomField } from "./lib/render-canvas-node-custom-field";
+import { getVersionActionAvailability } from "./lib/version-action-state";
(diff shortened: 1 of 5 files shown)


web_src/src/pages/workflowv2/index.tsx
```diff
@@ -723,9 +738,9 @@ export function WorkflowPageV2() {
     }
 
     if (!lastSavedWorkflowRef.current) {
-      lastSavedWorkflowRef.current = JSON.parse(JSON.stringify(canvas));
```
forestileao, line 726: Not a way that I like to make a deep copy but it's faster than other things... (spreads {...object} are just shallow copies)

forestileao:
👍

PR #4187, opened 2026-04-16
fix: use edit-mode empty states in canvas blocks

forestileao:
## Summary
This updates canvas node empty states so draft/edit mode no longer shows live-runtime messaging like `Waiting for the first run` or the rabbit icon.

## Changes
- Make node empty states mode-aware in `CanvasPage`
- Show a neutral edit-mode message for runtime-only empty states
- Keep setup and fallback empty states unchanged
- Default the node sidebar to `Settings` in edit mode instead of `Latest`
- Add coverage for edit-mode empty-state behavior in `Block.spec.tsx`

image


forestileao:
👍

PR #4188, opened 2026-04-16
fix: do not process all nodes when publishing version

lucaspin:
One issue with the current way we are publishing a new version is that we make a lot of unnecessary operations. Even when a node didn't change, we update it in the database and run Setup() for it. This pull request aims to make that process more efficient, by introducing a CanvasPublisher. CanvasPublisher does the following:
- Produces a changeset comparing the draft version being published with the current live version, and with it, it inserts / updates / delete workflow_node records only when needed.
-  Runs Setup() only when nodes are added / updated.
- Handle conflicting node IDs only for nodes that are being added.


pkg/grpc/actions/canvases/changesets/builder.go
```diff
@@ -0,0 +1,306 @@
...
+	currentNodes  map[string]models.Node
+	currentEdges  map[string]models.Edge
+	proposedNodes map[string]models.Node
+	proposedEdges map[string]models.Edge
+}
+
+func NewChangesetBuilder(currentNodes []models.Node, currentEdges []models.Edge, proposedNodes []models.Node, proposedEdges []models.Edge) *ChangesetBuilder {
+	edgeKeyFn := func(edge models.Edge) string {
+		return edge.SourceID + "|" + edge.TargetID + "|" + edge.Channel
+	}
```
shiroyasha: Why this function is not a constant and has to be recreated on every object? 🤔

pkg/grpc/actions/canvases/changesets/builder.go
```diff
@@ -0,0 +1,306 @@
...
+			continue
+		}
+
+		changes = append(changes, &pb.CanvasChangeset_Change{
+			Type: pb.CanvasChangeset_Change_DELETE_NODE,
+			Node: &pb.CanvasChangeset_Change_Node{
+				Id: nodeID,
+			},
+		})
+	}
```
shiroyasha, line 100: Just to sanity check:
There is no possibility that leaves behind unconnected edges when we remove a node?

shiroyasha (dismissed):

pkg/grpc/actions/canvases/changesets/canvas_publisher.go
```diff
@@ -0,0 +1,437 @@
...
+
+func (p *CanvasPublisher) deleteNode(change *pb.CanvasChangeset_Change) error {
+	existingNode, exists := p.allNodes[change.GetNode().GetId()]
+	if !exists {
+		return nil
+	}
+
+	delete(p.allNodes, existingNode.NodeID)
+	return models.DeleteCanvasNode(p.tx, existingNode)
+}
```
forestileao, line 334: Are we removing edges connected on this node?

forestileao: ah, same as https://github.com/superplanehq/superplane/pull/4188#discussion_r3096038578

pkg/grpc/actions/canvases/changesets/builder.go
lucaspin: [CanvasPatcher removes edges](https://github.com/superplanehq/superplane/blob/main/pkg/grpc/actions/canvases/changesets/canvas_patcher.go#L280) when processing DELETE_NODE changes. CanvasPublisher only uses the changeset as an indication of what happened, so it can update workflow_nodes records more efficiently, so it trusts that the nodes/edges on the CanvasVersion are correct, since they were produced by CanvasPatcher. However, the UpdateCanvasVersion endpoint is not yet using the CanvasPatcher, so I added a filter before publishing it, to ensure no dangling edges.

lucaspin:
👍

PR #4189, opened 2026-04-16
fix: disable live-mode edge and handle interaction affordances

forestileao:
## Summary
This removes edit-only interaction affordances from canvas live mode so edges and handles no longer look clickable or editable.

## Changes
- Hide edge delete affordances in live mode
- Disable connection editing in live mode
- Make node handles inert in live mode
- Remove live-mode edge hover tracking
- Remove live-mode edge hover transitions and pointer cursor
- Remove live-mode handle hover styling so handles no longer glow or feel interactive


forestileao:
👍

PR #4191, opened 2026-04-16
fix: hide sidebar runs tab in canvas edit mode

forestileao:
## Summary
This removes the node sidebar `Runs` / `Latest` view when the canvas is in edit mode, so the sidebar behaves like an editor instead of a live runtime viewer.

## Changes
- Pass canvas mode into `ComponentSidebar`
- Hide the `Runs` tab in edit mode
- Do not render `LatestTab` content in edit mode
- Normalize stale `latest` tab state back to `settings` when runs are unavailable
- Skip the runs loading state in edit mode
- Add a focused sidebar test covering live vs edit tab behavior

image


forestileao:
👍

PR #4193, opened 2026-04-17
fix: when adding new node, we Save() then Setup()

lucaspin:
Issue introduced by https://github.com/superplanehq/superplane/pull/4188.

The contexts given to triggers / components can create resources that reference the `workflow_node`. For example, the schedule trigger uses `ScheduleActionCall()` to create a workflow node request record. For that reason, for new nodes, we need to insert the record into the database, before running Setup().

### Other changes

We also fix the issue of metadata not being saved into the version nodes in UpdateCanvasVersion endpoint. Now, we inject the metadata into the nodes before saving them.


lucaspin:
👍

PR #4195, opened 2026-04-17
fix: respect layout in apply changeset endpoint

lucaspin:
CanvasPatcher is not really respecting the auto layout choice sent in the request. That was OK initially when this endpoint was intended only for agent use, but now, we want to migrate the UI to use it as well, so we need to respect what is in the request.


lucaspin:
👍

PR #4196, opened 2026-04-17
chore: use changeset for edge updates, collapsed state, node deletion and positions

lucaspin:


web_src/src/hooks/useCanvasData.ts
```diff
@@ -629,6 +631,76 @@ export const useUpdateCanvasVersion = (organizationId: string, canvasId: string)
   });
 };
 
+export const useApplyCanvasVersionChangeset = (organizationId: string, canvasId: string) => {
```
shiroyasha, line 634: Quite big, deserves its own file imho.

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -1376,6 +1378,72 @@ export function WorkflowPageV2() {
     [isAutoLayoutOnUpdateEnabled, components, blueprints],
   );
 
+  const applyChangeset = useCallback(
```
shiroyasha, line 1381: Way to complex to be part of this big function. 
Extract it to a useApplyChangeset hook.

lucaspin:
Closing this one because I thought using changesets would simplify UI a lot more. We can revisit it later

PR #4197, opened 2026-04-17
fix: hide node runtime actions in canvas edit mode

forestileao:
## Summary
This removes runtime action affordances from workflow canvas nodes while editing, so edit mode no longer exposes actions that only make sense in live mode.

## Changes
- Thread canvas mode into workflow v2 node mapper contexts
- Hide start trigger `Run` actions in edit mode while keeping template content visible
- Hide wait/time gate push-through actions in edit mode
- Keep approval state visible in edit mode but disable approve/reject actions
- Centralize edit-mode custom-field behavior in `ComponentBase` with `live-only` custom field visibility
- Add focused mapper coverage for edit vs live action behavior

### Edit mode

image

### Live

image


forestileao:
👍

PR #4199, opened 2026-04-17
chore: delete group widgets

forestileao:


web_src/src/lib/canvas/normalize.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { describe, expect, it } from "vitest";
+import { normalizeCanvasNodesWithoutGroups } from "./normalize";
+
+describe("normalizeCanvasNodesWithoutGroups", () => {
```
shiroyasha: Doesn't feel right that we are removing groups, but also leaving in the code traces of it.

pkg/grpc/actions/canvases/serialization.go
```diff
@@ -193,7 +193,7 @@ func canvasChangeRequestApproverTypeToProto(value string) pb.Canvas_ChangeManage
 }
 
 func serializeCanvasNodes(canvasID uuid.UUID, nodes []models.Node) ([]*compb.Node, error) {
-	serialized := actions.NodesToProto(nodes)
+	serialized := actions.NodesToProto(normalizeCanvasNodesWithoutGroups(nodes))
```
lucaspin: Why do we need this? Isn't the migration removing all group widgets?

web_src/src/lib/canvas/normalize.spec.ts
lucaspin: Yes. The whole point of removing it is to not have bad code about it around. If the migration removes all group nodes from the system, why do we need code to handle it?

db/data_migrations/20260417035041_remove-group-widgets-from-canvas-versions.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    )
+  ),
+  '[]'::jsonb
+)
+WHERE EXISTS (
+  SELECT 1
+  FROM jsonb_array_elements(nodes) AS node(item)
+  WHERE item->>'type' = 'TYPE_WIDGET'
+    AND item->'widget'->>'name' = 'group'
+);
```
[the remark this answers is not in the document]
forestileao: ?

forestileao:
migration tested on staging 👍

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -417,7 +409,7 @@ export function WorkflowPageV2() {
   }, [activeCanvasVersionId, selectedCanvasVersion, draftVersions, pendingApprovalVersionIds]);
   const latestDraftVersion = draftVersions[0];
   const createChangeRequestNodeDiffSummary = useMemo(
-    () => buildDraftNodeDiffSummary(liveCanvasVersion, createChangeRequestVersion),
+    () => buildDraftNodeDiffSummary(liveCanvasVersion, createChangeRequestVersion || undefined),
```
shiroyasha: Why `|| undefined`? 🤔

shiroyasha (dismissed):

web_src/src/pages/workflowv2/index.tsx
forestileao: It was leftover defensive coercion from an earlier version where that value could be nullish and we were normalizing arguments at the call site.

BUT It wasn’t needed anymore because buildDraftNodeDiffSummary already accepts an optional second argument and createChangeRequestVersion is valid to pass directly, so we removed it as unnecessary noise.

forestileao:
👍

forestileao:
@shiroyasha will need your approval again

lucaspin (approved):

PR #4204, opened 2026-04-17
chore: Don't keep the components sidebar open in live mode

shiroyasha:


forestileao (dismissed):

forestileao:
test_session.go:290: asserting visibility of "testID=\"building-blocks-sidebar\"": playwright: timeout: Timeout 10000ms exceeded.

Ah, there is an e2e

PR #4206, opened 2026-04-17
fix: prevent live canvas query from overriding edit mode

forestileao:
## Summary
Disable the live canvas detail query while a draft version is active, so an in-flight `/api/v1/canvases/:id` response cannot overwrite edit mode state during a fast live/edit toggle.

## Changes
- add an `enabled` flag to `useCanvas(...)`
- stop the live canvas query while a draft version is selected in `workflowv2`
- reduce unnecessary live detail refetches on remount/reconnect

## Why
A fast switch from live to edit could leave the live canvas request in flight. When that response arrived, it updated shared canvas state and replaced the draft view.

## Testing
- verified the live canvas query is disabled in edit mode
- manually retested fast `live -> edit -> live -> edit` toggling


forestileao:
👍

PR #4210, opened 2026-04-17
fix: update CreateCanvas to use Patcher and Publisher

### LINKED ISSUE #3550: CreateCanvas skips node Setup() validation
**Description:**  
When a canvas is created via the `CreateCanvas` API (used by the CLI, SDK, and AI-assisted canvas creation), node-level `Setup()` validation is never executed. All nodes are persisted with state "ready" regardless of whether their configuration is valid (e.g., HTTP component with an empty URL).

This means warnings like "url is required" or "recipients is required" are not surfaced until the user manually opens the canvas in the UI and triggers a save, at which point the `UpdateCanvasVersion` path runs `Setup()` and catches the errors.

**Steps to reproduce:**  
1. Create a canvas via CLI or API with an HTTP component that has an empty `url` field.
2. Open the canvas in the UI.
3. Observe that the HTTP node shows no warning/error state — it appears fully configured and "ready."
4. Click on the HTTP node and hit Save (without changing anything).
5. Now the "url is required" warning appears.

**Expected / Actual:**  
Expected: Nodes with invalid or incomplete configuration should show their error/warning state immediately after canvas creation.  
Actual: All nodes are stored as "ready" on creation. Warnings only appear after a subsequent save (update) triggers `Setup()` validation.

**Context:**  
- `pkg/grpc/actions/canvases/create_canvas.go` creates `CanvasNode` records with `State: CanvasNodeStateReady` and never calls `setupNode()`.
- `pkg/grpc/actions/canvases/update_canvas_version.go` (the update/save path) does call `setupNode()`, which runs each component's `Setup()` method and catches validation errors.
- This gap does not affect the standard UI flow (where users add components one by one, each triggering an update), but it affects CLI creation, API creation, AI-assisted creation, and template cloning.

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/3550. Also, fixes an issue where users can create canvas templates that would be available to everyone else.

### DIFF
--- a/pkg/grpc/actions/canvases/change_request_actions_test.go
+++ b/pkg/grpc/actions/canvases/change_request_actions_test.go
@@ -254,7 +254,7 @@ func createCanvasWithNoopNode(ctx context.Context, t *testing.T, r *support.Reso
 			Error,
 	)
 
-	createCanvasResponse, err := CreateCanvas(ctx, r.Registry, r.Organization.ID.String(), &pb.Canvas{
+	createCanvasResponse, err := CreateCanvas(ctx, r.Registry, r.Encryptor, r.AuthService, testWebhookBaseURL, r.Organization.ID, &pb.Canvas{
 		Metadata: &pb.Canvas_Metadata{Name: canvasName},
 		Spec: &pb.Canvas_Spec{
 			Nodes: []*componentpb.Node{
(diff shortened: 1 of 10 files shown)


lucaspin:
👍

PR #4253, opened 2026-04-20
fix: add proper error messages for cli errors

### LINKED ISSUE #4230: CLI 400: surface validation error body
**Problem:** When the API returns **400 Bad Request** with **no (or unusable) response body**, the CLI only shows the status line. Coding agents and humans cannot see **what failed validation**, so they cannot self-correct without guesswork.

**Current vs proposed:** Current: Opaque `400 Bad Request` from CLI. Proposed: API returns a structured validation error (field paths, messages) in the response body for validation failures; CLI **prints that body** (or a formatted summary) on stderr alongside the status. Scope may include gateway/proxy ensuring bodies are not stripped.

**Use case:** Agent-driven and scripted CLI use (`canvases update`, etc.) where a single clear error fixes the next attempt.

Related (closed, broader): https://github.com/superplanehq/superplane/issues/3643

re-pixel:
### Problem
CLI commands that failed validation printed only `400 Bad Request`, hiding the server's actual error message. Agents and scripts had to guess what went wrong and retry blindly.

### Changes

- Fixed the `errors.As` target in `FormatCommandError` and added a body fallback so the CLI always surfaces a useful message - decoded `GooglerpcStatus` when available, raw response body (capped) otherwise.
- Added end-to-end tests driving the real SDK via `httptest`.

### Before / After
Before:
`$ superplane canvases update --name ""` ... 400 Bad Request

After:
`$ superplane canvases update --name "" `... invalid request: canvas name is required

Closes: https://github.com/superplanehq/superplane/issues/4230

### DIFF
--- a/pkg/cli/core/errors.go
+++ b/pkg/cli/core/errors.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/cli/core/errors.go
```diff
@@ -64,25 +99,102 @@ func formatGoogleRPCStatusError(status *openapi_client.GooglerpcStatus) error {
...
+	case 12:
 		return "not supported"
-	case 13: // Internal
+	case 13:
 		return "internal error"
-	case 14: // Unavailable
+	case 14:
 		return "service unavailable"
-	case 16: // Unauthenticated
+	case 16:
```
forestileao: We can create a map for this. (If it does not already exists on our RPC protobuff dependencies)

pkg/cli/core/errors.go
```diff
@@ -64,25 +99,102 @@ func formatGoogleRPCStatusError(status *openapi_client.GooglerpcStatus) error {
...
+			switch {
+			case field != "" && desc != "":
+				out = append(out, fmt.Sprintf("%s: %s", field, desc))
+			case field != "":
+				out = append(out, field)
+			case desc != "":
+				out = append(out, desc)
+			}
+		}
+	}
```
forestileao, line 145: you can extract the inner parsing logic into a small helper and normalize the key lookup to avoid branching inside the loop


(THIS CODE IS NOT FORMATTED)
```golang


func extractFieldViolations(status *openapi_client.GooglerpcStatus) []string {
	if status == nil {
		return nil
	}
	
	var out []string

    for _, detail := range status.GetDetails() {
    	if !isBadRequestDetail(detail.GetType()) {
    		continue
    	}
    
    	list := getViolations(detail.AdditionalProperties)
    	if list == nil {
    		continue
    	}
    
    	for _, entry := range list {
    		field, _ := entry["field"].(string)
    		desc, _ := entry["description"].(string)
    
    		if s := formatFieldDesc(field, desc); s != "" {
    			out = append(out, s)
    		}
    	}
  }

func getViolations(props map[string]any) []map[string]any {
	raw := props["fieldViolations"]
	if raw == nil {
		raw = props["field_violations"]
	}

	items, ok := raw.([]any)
	if !ok {
		return nil
	}

	out := make([]map[string]any, 0, len(items))
	for _, item := range items {
		if m, ok := item.(map[string]any); ok {
			out = append(out, m)
		}
	}
	return out
}

func formatFieldDesc(field, desc string) string {
	switch {
	case field != "" && desc != "":
		return fmt.Sprintf("%s: %s", field, desc)
	case field != "":
		return field
	case desc != "":
		return desc
	default:
		return ""
	}
}
```

pkg/cli/core/errors.go
```diff
@@ -64,25 +99,102 @@ func formatGoogleRPCStatusError(status *openapi_client.GooglerpcStatus) error {
...
+	return out
+}
+
+func isBadRequestDetail(typeURL string) bool {
+	trimmed := strings.TrimSpace(typeURL)
+	return trimmed == badRequestDetailType || trimmed == badRequestDetailTypeAlt
```
forestileao: Here you can verify if it ends with "BadRequest" or  with the badRequestDetailTypeAlt constant

PR #4257, opened 2026-04-20
chore: simplify error handling in deleteAllTemplateWorkflows

Svengali-tech:
## Description
Refactored error handling in `deleteAllTemplateWorkflows` function to use idiomatic Go patterns.

## Changes
- Removed unnecessary `var err error` declaration
- Moved error handling inline with each operation using `if err := ...; err != nil` pattern
- Maintains the same functionality while improving code clarity and reducing variable scope

## Type of Change
- Code quality improvement (chore)


markoa:

PR #4259, opened 2026-04-20
docs: clarify demo vs development server ports

gatshaayanda:
## Summary
Adds a short note to clarify the difference between the demo container quick start and the local development setup.

## Why
The README quick start points users to localhost:3000 for the demo container, while the development setup uses a different local server port. This clarification should reduce confusion for first-time contributors.

## Testing
Documentation-only change.


markoa:

PR #4260, opened 2026-04-20
fix: move run title to trigger definitions and root events

### LINKED ISSUE #1611: Improve UI of setting run title
Current UI:
Image

is not intuitive because 
- there's no verb in copy that would clarify what you can do
- the toggle as a pattern doesn't fit the use case
- what is even a run?

Each trigger already has a sensible default for setting run titles, so let's expose that to users. It would set an expectation and make it immediately clear that it's a dynamic field. Something like this:

Image

Syntax is for illustration purposes. The word 'expressions' should be a link to a cheat sheet throughout the app.

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/1611

## Summary

Move run title ownership out of frontend renderers and trigger config fields into the backend trigger interface.

## Changes

- Add `DefaultRunTitle()` to the trigger interface and define run title defaults per trigger in the backend.
- Remove frontend-derived trigger/event titles and use the API-provided root event `runTitle` everywhere in workflow v2.
- Stop storing run title on the DB workflow node model; keep the editable `runTitleTemplate` only in the workflow node JSON.
- Resolve the template once when creating root events and persist the final value on the root event as `runTitle`.
- Standardize run title template resolution around `root()`, so templates use root event data instead of renderer-specific payload assumptions.
- Make sidebar `Run title` edits count as real draft changes, enabling Publish/Propose like any other config change.
- Use a consistent fallback title of `Event received at <timestamp>` when the run title is empty.

image
image


pkg/grpc/actions/common.go
```diff
@@ -1068,43 +1088,32 @@ func SerializeTriggers(in []core.Trigger) []*triggerpb.Trigger {
...
 	return out
 }
 
-func AppendGlobalTriggerFields(fields []configuration.Field) []configuration.Field {
-	if slices.ContainsFunc(fields, func(field configuration.Field) bool {
-		return field.Name == "customName"
-	}) {
-		return fields
+func TriggerDefaultRunTitle(trigger core.Trigger) string {
+	if title := strings.TrimSpace(trigger.DefaultRunTitle()); title != "" {
```
forestileao: WE DONT NEED THIS FUNCTION

pkg/integrations/aws/ecr/on_image_scan.go
```diff
@@ -73,6 +73,10 @@ func (p *OnImageScan) Color() string {
 	return "gray"
 }
 
+func (p *OnImageScan) DefaultRunTitle() string {
+	return `{{ root().data.detail["repository-name"] != "" ? root().data.detail["repository-name"] : "ECR image scan" }}{{ len(root().data.detail["image-tags"]) > 0 ? ":" + root().data.detail["image-tags"][0] : "" }}`
```
forestileao: Too much ternaries

web_src/src/pages/workflowv2/mappers/aws/codeartifact/create_repository.ts
```diff
@@ -100,18 +100,13 @@ function createRepositoryMetadataList(node: NodeInfo): MetadataItem[] {
 }
 
 function createRepositoryEventSections(
-  nodes: NodeInfo[],
+  _nodes: NodeInfo[],
```
lucaspin: If parameter is not used, remove from function, instead of _* it

web_src/src/pages/workflowv2/mappers/rendererTypes.ts
```diff
@@ -0,0 +1,8 @@
+export type {
```
lucaspin: Why a separate file for this? Can't we use types.ts?

pkg/workers/contexts/event_context.go
```diff
@@ -106,3 +105,36 @@ func (s *EventContext) resolveCustomName(payload any) (*string, error) {
 
 	return &resolvedName, nil
 }
+
+func resolveRootEventRunTitleTemplate(tx *gorm.DB, node *models.CanvasNode) (string, error) {
+	if node == nil {
```
lucaspin: Is not ever gonna be nil here?

pkg/workers/contexts/event_context.go
```diff
@@ -106,3 +105,36 @@ func (s *EventContext) resolveCustomName(payload any) (*string, error) {
...
+		return "", nil
+	}
+
+	liveNodes, _, err := models.FindLiveCanvasSpecInTransaction(tx, node.WorkflowID)
+	if err != nil {
+		return "", err
+	}
+
+	for _, liveNode := range liveNodes {
```
lucaspin: Why are we doing this? Can't we just use the `node *models.CanvasNode` directly?

pkg/workers/contexts/event_context.go
```diff
@@ -23,9 +25,13 @@ func NewEventContext(tx *gorm.DB, node *models.CanvasNode, onNewEvents func([]mo
...
 
 func (s *EventContext) Emit(payloadType string, payload any) error {
+	now := time.Now()
+	eventID := uuid.New()
 	structuredPayload := map[string]any{
+		"id":        eventID.String(),
 		"type":      payloadType,
-		"timestamp": time.Now(),
+		"timestamp": now.UTC().Format(time.RFC3339Nano),
+		"channel":   "default",
```
lucaspin: Why are you changing the event payload as part of this?

pkg/workers/contexts/event_context.go
forestileao: I didnt want to update that on models.canvasNodes, since that would be stored on model.canvasNodes and on json. But on db directly might be cleaner

forestileao: Actually, runTitleTemplate intentionally lives in the versioned canvas spec JSON, not in workflow_nodes. Looking it up from the live spec avoids reintroducing duplicated state between the DB row and the versioned node JSON.

What do you think about leaving it in the json?

pkg/workers/contexts/event_context.go
forestileao: I was changing the DefaultRunTitle based on their mappers. Some of them used the Ids. But that is ugly so i removed it

web_src/src/pages/workflowv2/mappers/aws/codeartifact/create_repository.ts
forestileao: removed from all functions I found using this pattern

web_src/src/pages/workflowv2/index.tsx
```diff
@@ -2750,6 +2756,7 @@ export function WorkflowPageV2() {
             ...node,
             configuration: updatedConfiguration,
             name: updatedNodeName,
+            runTitleTemplate,
```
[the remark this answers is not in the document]
forestileao: this one doesn't seem to be true since I tested but let me check more deeply.

web_src/src/pages/workflowv2/mappers/types.ts
```diff
@@ -4,10 +4,29 @@ import type {
...
+  ComponentBaseSpec,
+  EventSection as UIEventSection,
+  EventState,
+  EventStateMap,
+} from "@/ui/componentBase";
+import { DEFAULT_EVENT_STATE_MAP } from "@/ui/componentBase";
+import type { TriggerProps as UITriggerProps } from "@/ui/trigger";
 import type { ReactNode } from "react";
 
+export type RendererEventSection = Omit<UIEventSection, "eventTitle">;
```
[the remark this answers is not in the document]
forestileao: that is the idea :)

pkg/grpc/actions/canvases/emit_node_event.go
```diff
@@ -44,9 +44,15 @@ func EmitNodeEvent(
 		CreatedAt:  &now,
 	}
 
-	customName, err := resolveCustomName(node, data)
-	if err == nil && customName != nil {
-		event.CustomName = customName
+	rootPayload := contexts.BuildRootEventPayload(data, "", now)
```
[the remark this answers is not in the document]
forestileao: This is only used for manual run payload

pkg/workers/contexts/event_context.go
lucaspin: Isn't that inconsistent? All the other fields (position, configuration, collapsed state, ...) are in both places, no?

forestileao: You are right. Well, Let me update that

forestileao:
@lucaspin Migration and event context updated. Everything continues working

image

forestileao:
Closing this PR, since we will probably change the expr lang in the future..

PR #4300, opened 2026-04-21
fix: respect --from when describing index components and triggers

hassanredaa:


markoa:

PR #4301, opened 2026-04-21
fix: auto sign-off DCO for Cursor PRs

### LINKED ISSUE #3520: DCO check should not fail when Cursor is opening the PR

harxhist:
Closes #3520 
Fix to auto sign-off cursor PRs


shiroyasha:
Interesting approach! Not yet sure if this is the right fit for our repo. I'll have to think about it.

shiroyasha:
Thank you for opening this PR. After reviewing it several times I realized that we can't use this solution. DCO signing needs to be tied to a human, not to a robot.

Issue #4249, opened 2026-04-20
Agents can't use Manual Run


shiroyasha:
<img width="1500" height="800" alt="Image" src="https://github.com/user-attachments/assets/efbadd59-e666-467a-bd3a-6ce4006ba333" />

PR #4305, opened 2026-04-21
feat: add manual run cli command

### LINKED ISSUE #4298: CLI: programmatic run and event replay for canvases
### Context
Agents iterating on canvases cannot close the verification loop autonomously today.

### Observed behavior
After editing and deploying YAML, test execution still requires a human to click **Run** in the UI and report results back.

Also, manual-trigger payloads can differ from real `github.onPush` payloads (nesting/shape differences), causing avoidable bugs (e.g. triple nesting like `root().data.data.data`).

### Suggested improvement
Add CLI command such as:

```bash
superplane canvases run <canvas-id> --trigger <trigger-id> [--template <name> | --payload <file>]
```

Bonus:

```bash
--replay <event-id>
```

to re-fire a real historical event and avoid template-vs-real payload mismatch.

### Why this matters
Enables fully agentic test loops and speeds iteration significantly.

re-pixel:
Resolves the issue: https://github.com/superplanehq/superplane/issues/4249
Partially resolves https://github.com/superplanehq/superplane/issues/4298

### Summary

Gives external agents (and users) a first-class way to fire a canvas's Manual Run (start) trigger from the CLI, instead of having to discover and call the trigger hook API directly. This is the CLI equivalent of clicking the "Run" button in the UI.

```
superplane canvases run <canvas> \
  --node <start-node-id> \
  --template <template-name> \
  [--payload '<json-or-file>']
```
### CLI (canvases run)

- Resolves the canvas by name or ID, validates the target node is a start trigger, and validates the template name exists on the node (listing available templates on mismatch).
- Invokes the trigger's run hook with the template name and optional payload override.
- Returns actionable errors for common failure modes: node not found, node is not a trigger, trigger isn't start, unknown template name, invalid JSON, array payloads.

UI (emit modal)

When running from a template button, the modal now receives the templateName, skips the channel selector, and shows a focused "Run trigger" UX with contextual title, description, and toast messages.


lucaspin:
@re-pixel this pull request is doing too much at once so it's a bit hard to review. I can see us splitting this into multiple PRs a bit:
1. Update the manual trigger to have hooks and the UI to use the hooks API to execute it.
1. Remove the EmitNodeEvent endpoint. This will require us to add a new endpoint for re-emitting a trigger event, because the UI is currently using EmitNodeEvent too.
2. Update the CLI

re-pixel:
> @re-pixel this pull request is doing too much at once so it's a bit hard to review. I can see us splitting this into multiple PRs a bit:
> 
> 1. Update the manual trigger to have hooks and the UI to use the hooks API to execute it.
> 2. Remove the EmitNodeEvent endpoint. This will require us to add a new endpoint for re-emitting a trigger event, because the UI is currently using EmitNodeEvent too.
> 3. Update the CLI

@lucaspin here is the first PR https://github.com/superplanehq/superplane/pull/4503

PR #4307, opened 2026-04-21
fix: detect integration binding changes in workflow publish state

### LINKED ISSUE #4225: Integration swap not detected for publish
**Description:**  
Changing which integration instance a component uses (e.g. GitHub) is not treated as a draft change versus live, so **Publish** stays disabled and users cannot ship the update.

**Steps to reproduce:**  
1. Add two GitHub integrations (e.g. github1, github2).  
2. Add a GitHub component bound to github1.  
3. Publish the canvas.  
4. Edit the canvas and change the component's integration from github1 to github2.  
5. Observe: Publish remains disabled even though the binding changed.

**Expected / Actual:**  
Expected: Draft differs from live → Publish enabled (or change request flow). Actual: No draft-vs-live graph diff for this edit → Publish stays disabled.

forestileao:
## Summary

Closes: https://github.com/superplanehq/superplane/issues/4225

Fixes a frontend issue where changing a node’s integration binding was not treated as a draft change, so the `Publish` button could stay disabled or fail to recognize the update correctly.

## Problem

When a user changed a component or trigger from one integration instance to another, the draft/live diff logic did not correctly detect that integration-binding change.

As a result:
- the draft was treated as unchanged
- `Publish` did not reliably reflect that there was something to publish

## Fix

- Updated workflow diff/signature comparison so integration binding changes are detected correctly.
- Normalized integration comparison to use the comparable integration identity instead of UI-only integration object differences.
- Reused the same integration comparison helper across workflow diff code.

### DIFF
--- a/web_src/src/pages/workflowv2/VersionNodeDiff.tsx
+++ b/web_src/src/pages/workflowv2/VersionNodeDiff.tsx
@@ -6,6 +6,7 @@ import { Diff, Hunk, parseDiff } from "react-diff-view";
 import { useMemo } from "react";
 import * as yaml from "js-yaml";
 import "react-diff-view/style/index.css";
+import { getComparableIntegrationId } from "./utils";
 import { WorkflowMarkdownPreview } from "./WorkflowMarkdownPreview";
 
 export function buildInitials(name?: string): string {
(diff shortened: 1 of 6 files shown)


forestileao:
👍

PR #4308, opened 2026-04-21
feat: persist agent run and restore chat state after disconnect

### LINKED ISSUE #4104: Agent: persist run when sidebar closes
**Problem:**  
If the user sends a prompt and the agent is still working, **closing the chat** (sidebar, navigating away) or **refreshing the page** appears to **abort** the run. Progress and partial results feel **lost** instead of continuing server-side.

**Current vs proposed:** **Current:** In-flight agent work is tied to the open UI session; leaving stops the experience. **Proposed:** Agent execution **continues on the server**; when the user returns, they see the same chat with **status, partial output, and final result** (or errors), same as resuming a long-running job.

**Use case:**  
Long builds, pattern search, or multi-step flows where users should not lose work when they dismiss the panel or reload accidentally.

0vertake:
## Summary

Agent execution now continues on the server after the user closes the
sidebar or refreshes the page. When they return they see the same chat
in the same state — spinner if the run is still in progress, streamed
messages as they arrive, and the canvas-change proposal once the run
finishes.

## What changed

**Python agent (`agent/`)**
- `_run_agent_to_queue` runs the agent as an independent `asyncio.Task`,
  fully decoupled from the SSE connection. Closing the browser tab no
  longer cancels the underlying AI call.
- New `latest_run_status` column on `agent_chats` (`running` →
  `completed` / `failed`). Set atomically by `create_agent_chat_run`,
  `update_run_usage`, and `mark_run_failed`.
- `mark_run_failed` is called on task error and also as a fallback if
  `update_run_usage` fails, so the DB never stays stuck at `running`.
  Guarded against overwriting a `completed` status.
- `reset_stale_running_chats()` runs at startup to flip any chats left
  `running` by a previous crash or OOM kill.
- `proposal` JSON persisted on each assistant message so the
  canvas-change approval UI survives a refresh.
- `ActiveStreamTracker` ownership transferred to the background task on
  disconnect so graceful shutdown waits for real agent completion, not
  just the SSE connection lifetime.
- SSE queue consumer changed from a busy-poll loop (100 wakeups/s) to
  `asyncio.wait_for(queue.get(), timeout=0.1)`.

**Proto / Go (`protos/`, `pkg/`)**
- `proposal` added to `AgentChatMessage` (field 7).
- `latest_run_status` added to `AgentChatInfo` / `ChatInfo` (field 4).
- Mapped through `describe_agent_chat`, `list_agent_chats`, and
  `list_agent_chat_messages`.

**Frontend (`web_src/`)**
- `extractPendingProposalFromApiMessages` parses the persisted proposal
  from loaded messages.
- `useLoadChatConversation` fetches messages and sessions in parallel;
  restores the spinner when `latestRunStatus === "running"` and the
  proposal when the run is done.
- New `usePollForMessages` hook polls every 2 s when a run is in progress
  but no SSE stream is active. Skips ticks while the tab is hidden;
  fires immediately on tab focus via `visibilitychange`.
- `currentChatId` lifted into `useAgentState` so canvas changes
  correctly reset it.

## Test plan

- [x] Send a prompt that triggers a canvas-change proposal. While the
  agent is thinking, refresh the page and navigate back to the chat.
  Verify the spinner shows and the proposal appears when the run finishes.
- [x] Send a simple prompt ("what nodes are on my canvas?"), get a
  response, refresh. Verify the conversation is restored.
- [x] Open a chat mid-run and switch to another browser tab while the
  agent works. Verify no polling requests fire while the tab is hidden,
  and messages appear promptly when you switch back.
  
  Closes #4104

### DIFF
--- a/agent/db/migrations/20260421120636_add-latest-run-status-to-agent-chats.up.sql
+++ b/agent/db/migrations/20260421120636_add-latest-run-status-to-agent-chats.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE agent_chats
+    ADD COLUMN latest_run_status TEXT NOT NULL DEFAULT '';
--- a/agent/db/migrations/20260421120636_add-latest-run-status-to-agent-chats.up.sql
+++ b/agent/db/migrations/20260421120636_add-latest-run-status-to-agent-chats.up.sql
@@ -1,2 +1,5 @@
 ALTER TABLE agent_chats
     ADD COLUMN latest_run_status TEXT NOT NULL DEFAULT '';
+
+ALTER TABLE agent_chat_messages
+    ADD COLUMN proposal TEXT;
--- a/agent/db/structure.sql
+++ b/agent/db/structure.sql
@@ -45,7 +45,8 @@ CREATE TABLE public.agent_chat_messages (
     message jsonb NOT NULL,
     created_at timestamp with time zone DEFAULT now() NOT NULL,
     updated_at timestamp with time zone DEFAULT now() NOT NULL,
-    run_id uuid
+    run_id uuid,
+    proposal text
 );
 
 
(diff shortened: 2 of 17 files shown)


shiroyasha:
The PR is good, but I'm closing everything related to the old agent system.

Issue #3624, opened 2026-03-24
Toggling between 'Fixed' and 'Expression' type values clears previously entered content


markoa:
The value should be intact.

https://github.com/user-attachments/assets/92322801-83e4-4bb5-9e5e-1af4a797d008

tracemycodes:
Hi @AleksandarCole 
I've opened this PR to resolve the ticket mentioned in the thread 

You can find a quick walkthrough of the implementation here: https://www.loom.com/share/e316f122e3d340d6b40d86709ceef7aa. I'll also be tagging a few other team members to help with the review. Thanks!

PR #4314, opened 2026-04-21
fix: toggling between 'fixed' and 'expression' type values

tracemycodes:
Addresses this [issue](https://github.com/superplanehq/superplane/issues/3624)

**Summary**

Integration resource fields with expression support now preserve user input when switching between Fixed and Expression tabs. When a user types a search query in the Fixed tab's picker, the query is automatically converted to an expression format when switching to the Expression tab, and both tab states are cached independently for seamless back-and-forth navigation.

DEMO Video
https://www.loom.com/share/e316f122e3d340d6b40d86709ceef7aa

**What Changed**
**AutoCompleteSelect Component (`AutoCompleteSelect`)**

- Added optional controlled `query` interface for external query state management.
- Implemented controlled/uncontrolled pattern: uses external `query` prop when provided (controlled mode), falls back to internal state when omitted (uncontrolled mode).
- Updated prop naming from `externalQuery` to `controlledQuery` following React controlled component conventions.
- No breaking changes to existing usage in `DayInYearFieldRenderer` (operates in uncontrolled mode by default).

**IntegrationResourceFieldRenderer Component (`configurationFieldRenderer`)**

- New `selectQueryValue` state tracks search input in the Fixed tab picker independently from selected values.
- Enhanced caching system with three distinct state variables:
   `selectQueryValue`: Ephemeral search query during typing
   `cachedFixedValue`: Persistent selected value for Fixed tab
   `cachedExpressionValue`: Persistent expression text for Expression tab
- Smart tab switching logic in `handleTabChange`:
   Fixed → Expression: Wraps search query in `{{}}` if valid (no existing braces), otherwise restores cached expression
   Expression → Fixed: Restores cached fixed selection
- Brace validation using `/[{}]/.test()]` prevents double-wrapping expressions
- Query state clears after selection to avoid confusion
- Expression cache preserved when making fixed selections (no aggressive clearing)

**Implementation Details**
**State Flow Architecture:**

1. User types in Fixed tab search → selectQueryValue updates → prepares cachedExpressionValue as {{query}}
2. User switches to Expression tab → sees wrapped query or previous expression
3. User types in Expression tab → only updates cachedExpressionValue, doesn't contaminate search query
4. User switches back to Fixed → sees previous fixed selection
5. Caches persist across multiple tab switches

**Key Technical Decisions:**

- **Controlled query pattern:** Allows parent component to access search-in-progress text without modifying AutoCompleteSelect's internal behavior
- **Separate cache states:** Fixed and Expression values stored independently to prevent cross-contamination
- **Smart wrapping:** Only wraps search query in {{}} when switching to expression mode AND query doesn't already contain braces
- **UX preservation:** Aggressive cache clearing removed - expression value persists when user selects a fixed value

**Testing Checklist**
 

- [ ] Verify search query wraps in {{}} when switching Fixed → Expression
- [ ] Verify expression value persists when editing and switching back
- [ ] Verify fixed selection restores when switching Expression → Fixed
- [ ] Verify multiple back-and-forth tab switches preserve both caches
- [ ] Verify search query with existing {} characters doesn't double-wrap
- [ ] Verify DayInYearFieldRenderer (uncontrolled mode) still works
- [ ] Verify clearing search clears expression cache appropriately

### DIFF
--- a/web_src/src/components/AutoCompleteSelect/AutoCompleteSelect.tsx
+++ b/web_src/src/components/AutoCompleteSelect/AutoCompleteSelect.tsx
@@ -18,6 +18,7 @@ export interface AutoCompleteSelectProps {
   className?: string;
   error?: boolean;
   disabled?: boolean;
+  onQueryChange?: (query: string) => void;
 }
 
 export function AutoCompleteSelect({
@@ -28,6 +29,7 @@ export function AutoCompleteSelect({
   className,
   error = false,
   disabled = false,
+  onQueryChange,
 }: AutoCompleteSelectProps) {
   const [isOpen, setIsOpen] = useState(false);
   const [query, setQuery] = useState("");
@@ -100,7 +102,9 @@ export function AutoCompleteSelect({
   };
 
   const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    setQuery(e.target.value);
+    const newQuery = e.target.value;
+    setQuery(newQuery);
+    onQueryChange?.(newQuery);
     if (!isOpen) setIsOpen(true);
   };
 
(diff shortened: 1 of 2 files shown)


markoa:

Issue #3453, opened 2026-03-13
[Oracle Cloud] Base


AleksandarCole:
### Description
Create a SuperPlane integration for Oracle Cloud Infrastructure (OCI) and two starter components (one trigger, one action) so users can build workflows across Oracle Cloud services in SuperPlane. 

### First Components 

1. On Compute Instance Created (Trigger)                                                                                                                              
                                                              
Emits when a new OCI Compute instance reaches RUNNING state. The trigger should use the appropriate OCI event source — OCI Events Service with a rule matching com.oraclecloud.computeapi.launchinstance.end — and emit the instance creation payload to start SuperPlane workflow executions.

2. Create Compute Instance (Action)                             

Provision a new OCI Compute instance and return the instance OCID, display name, lifecycle state, public/private IP addresses, shape, availability domain, and key metadata.

shiroyasha:
@AleksandarCole what is the correct state of this issue?

shiroyasha:
Based on the state of the PR, this is closed.

PR #4321, opened 2026-04-22
feat: add Oracle Cloud Infrastructure (OCI) base integration

felixgateru:
Implements: #3453

## What changed:

Added a new  Oracle Cloud Infrastructure (OCI) integration to SuperPlane, including:

An OCI integration that authenticates via API Key (RSA private key + fingerprint) and provisions a shared OCI Notifications (ONS) topic on setup
A Create Compute Instance action component that launches an OCI Compute instance 
An On Compute Instance Created trigger that fires when a compute instance launch completes in a configured compartment

## Why:

OCI is a required cloud provider. SuperPlane previously had no OCI support.

### Backend:

Created integration lifecycle: validates credentials, creates/deletes a shared ONS topic  and exposes OCI resources 
Added  async create instance action which launches instance and polls until RUNNING 
Added trigger that  creates a per-node OCI Events rule  and emits oci.computeInstanceCreated events

### Frontend:

Created mapper files for both components (create_compute_instancets and on_create_compute_instance) with node metadata, execution details, and event state rendering.
Added corresponding TypeScript interfaces in types.ts
Registered all mappers and event state registries in index.ts.

## Demo
https://youtu.be/tCNnrOvumxs

### Authentication
The integration works by using the OCI API which provides the required functionality for CRUD operations on the resources provided by OCI. For authentication the API request require to be signed with the userOCID, tenant OCID, fingerprint and secret key. These are required when setting up the integration but are easily copied from a configuration file presented by the OCI dashboard as shown below:
image
 The instructions to obtain these values is provided when setting up the integration as shown below:
 
image

### Webhooks
To receive webhooks OCI requires a custom URL subscription to an existing Topic where events are published using Rules. On integration setup a topic is created and a webhook URL is used create a subscription to the topic.
When a trigger component is created this creates a rule to publish events of the related trigger to the topic.


WashingtonKK (commented):
Image

icon is missing here

pkg/integrations/oci/on_compute_instance_created.go
```diff
@@ -0,0 +1,291 @@
...
+			return http.StatusOK, nil, nil
+		}
+	}
+
+	if err := ctx.Events.Emit(OnComputeInstanceCreatedPayloadType, envelope); err != nil {
+		return http.StatusInternalServerError, nil, fmt.Errorf("failed to emit event: %w", err)
+	}
+
+	return http.StatusOK, nil, nil
+}
```
forestileao, line 269: let's split this logic in some functions

forestileao (approved):

PR #4322, opened 2026-04-22
feat(canvas): add 'c' keyboard shortcut to open the add-component panel

### LINKED ISSUE #1613: Keyboard shortcut to open add new component panel
- press `c`
- components panel opens
- search input is autofocused
- I can type something like 'filt' and press return for a component to be dropped on the canvas

AlgusDark:
## Summary

Closes: https://github.com/superplanehq/superplane/issues/1613

Adds a keyboard-first path to add a new component to the canvas:

- Pressing **c** opens the **Add component** panel (same side-effect as clicking the `+` button).
- The filter input inside the panel is auto-focused, so the user can start typing immediately.
- Pressing **Enter** drops the first matching component at the viewport center. 

Before (mouse only):

1. Aim for the `+` button. 
2. Click. 
3. Click search box. 
4. Type. 5. Click the result.

After (keyboard):

1. Press `c`. 
2. Type. 
3. Press `Enter`.

## Design notes

- **Active only in editor mode.** The shortcut mirrors the visibility rules of the `+` button: disabled when `readOnly`, in `version-live` mode, when `hideAddControls` is true, or when the component inspector (right-side sidebar) is already open. We did _not_ fold these into a single \"is add-button visible\" prop on purpose — the shortcut and the button can reasonably diverge (e.g. we want `c` off while inspecting a node, but the `+` button is still rendered there).
- **No visible failure indicator**: empty filter, no-match filter, or an already-open panel are all silent no-ops.
- **Safe around text input.** The global listener ignores events coming from `<input>`, `<textarea>`, `<select>`, `[contenteditable=\"true\"]`, or any `.monaco-editor` descendant, so `c` never eats typed characters in forms, YAML editors, etc.
- **Modifier-free.** `Cmd/Ctrl/Alt + c` is passed through untouched (so browser copy still works).
- **No tooltip/keycap yet.** That's worth its own PR, it needs a reusable `<Kbd />` and a keyboard-shortcut discovery convention.

## Manual verification

- [x] Press \`c\` in editor mode → Add component panel opens and filter input is focused.
- [x] Type a few chars → press \`Enter\` → first matching block drops at viewport center.
- [x] Panel already open → \`c\` does nothing.
- [x] Focus an \`<input>\` / textarea / YAML editor → \`c\` types the literal character.
- [x] \`Cmd+c\` still copies; \`Ctrl+c\` still works.
- [x] In version-live / read-only → \`c\` does nothing.
- [x] Component inspector open → \`c\` does nothing (fixed in commit 6).


lucaspin (approved):

lucaspin:
@AlgusDark code looks great! I agree that we need to a bit more keyboard-friendly, and this is a very good first step. Thanks for this one!

PR #4323, opened 2026-04-22
feat: add expression and messaging skills

re-pixel:
Two focused skills that give third-party agents the reference context they need to edit canvas node configurations correctly on the first pass:

`superplane-expressions` -- Expr syntax (payload access, nil-safe patterns with ?./??, let bindings, truncation, string/array/date functions), payload structures, YAML scalar styles with {{ }}, expression vs text field types, error behavior, and tested examples from real canvas templates.

`superplane-messaging` -- Per-provider field reference for all messaging components (Discord embeds, Slack mrkdwn, Telegram Markdown, Teams, SendGrid, SMTP) with character limits, capability summary, and formatting examples including a multi-provider comparison.

Split into two skills as there are many use-cases where only one of them is relevant.


re-pixel:
Closing this pull request, its changes belong to https://github.com/superplanehq/skills

https://github.com/superplanehq/skills/pull/14

Issue #4273, opened 2026-04-21
GitHub: createPullRequest component


AleksandarCole:
### Context
GitHub integration currently exposes many components but not a direct pull-request creation action.

### Observed behavior
To open a PR from a canvas, a workflow had to route through `cursor.launchAgent` (heavy workaround: launching a full cloud agent run for a basic GitHub action).

### Why this is a problem
- Overkill for a common automation step.
- Increased latency and complexity.
- Harder to keep pure integration-driven pipelines simple.

### Suggested improvement
Add `github.createPullRequest` component with fields like:
- `repository`
- `head`
- `base`
- `title`
- `body`
- optional labels/reviewers/draft flag (if supported)

This is a high-frequency GitHub automation primitive and should be first-class.

PR #4324, opened 2026-04-22
feat: add github create PR action component

re-pixel:
Demo: https://drive.google.com/file/d/1lOY--TLGoMNHG8REseDqnBn6J6o6nwHQ/view?usp=sharing

### Summary

- Adds a new `github.createPullRequest` action component that creates pull requests via the GitHub API, resolving #4273 .
- Introduces a branch resource type to the GitHub integration's ListResources, enabling dynamic branch dropdowns that cascade from the selected repository.
- Registers the component in both the backend (component registry, example output) and frontend (mapper, state registry).


web_src/src/pages/workflowv2/mappers/github/types.ts
```diff
@@ -62,16 +62,21 @@ export interface Release {
 
 export interface PullRequest {
   title?: string;
-  id?: string;
+  id?: number;
```
shiroyasha: Is there maybe a reason why this was previously a string?
Maybe something about how it is rendered in the cnfiguration UI?

cc @forestileao @lucaspin maybe you know

lucaspin: The only thing I remember is that I had to use `string` instead of `number` for `github.getIssue` was because there was no way to use expressions on number-typed fields. But, it seems like this new component is not using the ID in the config, so could be something else

shiroyasha:
@re-pixel looks good! Once you resolve the conflicts, we are ready to merge.

re-pixel:
@shiroyasha resolved conlicts, had to refactor the component since the whole github integration was refactored in the meantime. works fine after the changes.

shiroyasha:
Looks great!

PR #4328, opened 2026-04-22
fix(ws): validate WebSocket Origin to prevent cross-site hijacking

SatyaKilaru:
## Summary

The WebSocket upgrader currently accepts all origins (`CheckOrigin: return true`), exposing cookie-authenticated users to cross-site WebSocket hijacking (CSWSH).

## Why it's exploitable

`/ws/{workflowId}` is protected by `OrganizationAuthMiddleware`, which accepts **either** a Bearer token **or** a session cookie (see `pkg/public/middleware/auth.go:201`). Browsers cannot set custom `Authorization` headers on WebSocket handshakes via the JS WebSocket API, so browser clients fall back to cookies — which are sent automatically on cross-origin requests. A logged-in user visiting an attacker's page can have a WebSocket opened to their workflow stream.

CLI / server-to-server clients that use Bearer tokens are not affected.

## Fix

- Validate the `Origin` header against a list configured via `ALLOWED_WS_ORIGINS` (comma-separated)
- Fall back to `BASE_URL` when `ALLOWED_WS_ORIGINS` is unset, so single-tenant deployments work without extra config
- Permit requests with no `Origin` header (non-browser clients)
- Origin comparison per RFC 6454: scheme + host, case-insensitive, trailing-slash tolerant

## Testing

- New unit test `TestMakeOriginChecker` with 7 cases (exact match, trailing slash, case, mismatched host, mismatched scheme, missing origin, malformed origin)
- Full `make test` passes locally — 6034 tests, 0 failures

## Notes

- Discussed with maintainers on Discord before opening — greenlit as a good first contribution
- Closes the existing `TODO: implement origin checking` at `pkg/public/server.go:180`

### DIFF
--- a/pkg/public/origin_check_test.go
+++ b/pkg/public/origin_check_test.go
@@ -1,0 +1,36 @@
+package public
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestMakeOriginChecker(t *testing.T) {
+	check := makeOriginChecker([]string{"https://app.superplane.com", "http://localhost:8000/"})
+
+	cases := []struct {
+		name   string
+		origin string
+		want   bool
+	}{
+		{"exact match", "https://app.superplane.com", true},
+		{"trailing slash tolerance", "http://localhost:8000", true},
+		{"case-insensitive host", "HTTPS://APP.SUPERPLANE.COM", true},
+		{"different host rejected", "https://evil.example.com", false},
+		{"different scheme rejected", "http://app.superplane.com", false},
+		{"missing origin allowed (CLI)", "", true},
+		{"malformed origin rejected", "not-a-url", false},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


SatyaKilaru (commented):
Heads up on the failing E2E tests: the 4 failures (`TestServiceAccounts`, `TestInvitations` and two subtests) all occur on plain HTTP endpoints (`/api/v1/me`, `/api/v1/service-accounts`, `/api/v1/organizations/...`). None of the failure traces show a WebSocket upgrade, so `CheckOrigin` is never invoked on these requests.

The failure patterns also look like timing issues — `Timeout 10000ms exceeded`, sporadic 404s on `/api/v1/me`, and a `SLOW SQL >= 200ms` warning preceding the `TestInvitations` failure.

Happy to dig deeper if maintainers think these are related to the change; otherwise a re-run should clear them. Unit tests all passed (including the new `TestMakeOriginChecker`).

Issue #1419, opened 2025-12-22
Components naming confusion


ropsii:
We call nodes on the main canvas `components`.

We also call custom components - `components` in the UI.

Custom components are `component groups`, that's the most basic concept to understand. I don't know if that's the best solution, but we should make a clear distinction.

PR #4306, opened 2026-04-21
Canvas Homepage: Sorting & Time Display Enhancement

edbienes:
Closes #3599 (partial implementation)

📋 **Summary**
Implemented sorting capabilities and improved time display for canvas cards on the homepage, addressing part of the canvas list navigation improvements.

✨ **Changes**
🔄 **Sort Functionality**
- Added dropdown with sorting options: **Name (A-Z/Z-A)** and **Created (Newest/Oldest)**
- Works across both grid and list view modes
- Uses accurate date parsing for chronological sorting

⏰ **Enhanced Time Display** 
- Replaced absolute timestamps with **relative time** (e.g., "2d ago")
- Added **hover tooltips** showing full date/time
- Visual indicators (dotted underline) for interactive elements

🚀 **Navigation Improvement**
- Added **"View Runs"** action to canvas context menu
- Provides quick access to `/{orgId}/canvases/{id}/runs`

🔧 **Technical Details**
- **Files:** `web_src/src/lib/date.ts`, `web_src/src/pages/home/index.tsx`
- **TypeScript:** Full type safety maintained
- **Performance:** Efficient sorting with native Date parsing
- **Accessibility:** Proper ARIA labels and keyboard support

🧪 **Testing**
- ✅ TypeScript compilation passes
- ✅ Existing functionality preserved  
- ✅ Responsive design maintained

PR #4330, opened 2026-04-22
feat: support both Docker and Podman as container runtimes

edbienes:
## Summary

- Introduce `CONTAINER_ENGINE ?= docker` in Makefiles, allowing developers to opt into Podman via `CONTAINER_ENGINE=podman`
- Replace all hardcoded `docker` / `docker compose` references with `$(CONTAINER_ENGINE)` / `$(CONTAINER_ENGINE) compose`
- Parameterize `cache_from` image references so both engines can parse them (untagged for Podman, tagged for Docker)

## Motivation

Docker Desktop's memory consumption is a pain point for some developers. Podman is a lightweight, daemonless alternative that supports the same CLI interface. Rather than choosing one over the other, this change lets each developer use their preferred engine while keeping Docker as the default.

## Changes

| File                          | What changed                                                                                                            |
| ----------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| `Makefile`                    | `CONTAINER_ENGINE` selector, computed env var exports, `CONTAINER_RUN_AS_CURRENT_USER` rename, image build/push targets |
| `agent/Makefile`              | Same `CONTAINER_ENGINE` pattern                                                                                         |
| `docker-compose.dev.yml`      | Parameterized `cache_from` with engine-aware defaults                                                                   |
| `scripts/vscode_run_tests.sh` | Respects `CONTAINER_ENGINE`, renamed `docker_exec` → `compose_exec`                                                     |

## Usage

```bash
# Docker (default, no change needed)
make dev.setup
make dev.start

# Podman (explicit opt-in)
make dev.setup CONTAINER_ENGINE=podman
make dev.start CONTAINER_ENGINE=podman

# Or export once in your shell profile
export CONTAINER_ENGINE=podman
make dev.setup
make dev.start
```

## Podman compatibility notes

- **Minimum `podman-compose` version**: Nested variable interpolation (`${VAR:-${VAR2:-default}}`) in compose files requires the March 2026 release or later (podman-compose PR #1419). Earlier versions will fail to parse the compose file.
- **`--cache-from` tags**: Podman/Buildah rejects tagged image references in `--cache-from` by design. The Makefile conditionally uses untagged refs for Podman and tagged refs for Docker.
- **`depends_on: condition: service_healthy`**: Has partial support in some `podman-compose` versions. If health-check gating doesn't work, services may start before dependencies are ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,15 @@ PUBLIC_API_PORT ?= 8000
 PORT ?= $(PUBLIC_API_PORT)
 API_PORT ?= $(PUBLIC_API_PORT)
 AGENT_HTTP_PORT ?= 8090
+# Docker BuildKit requires tagged images in cache_from; Podman's --cache-from
+# rejects tags ("repository must contain neither a tag nor digest").
+ifeq ($(CONTAINER_ENGINE),podman)
 APP_DEV_BASE_CACHE_IMAGE ?= ghcr.io/superplanehq/superplane-dev-base
 AGENT_DEV_BASE_CACHE_IMAGE ?= ghcr.io/superplanehq/superplane-dev-base
+else
+APP_DEV_BASE_CACHE_IMAGE ?= ghcr.io/superplanehq/superplane-dev-base:app-latest
+AGENT_DEV_BASE_CACHE_IMAGE ?= ghcr.io/superplanehq/superplane-dev-base:agent-latest
+endif
 export GRPC_SERVER_ADDR PORT API_PORT AGENT_HTTP_PORT APP_DEV_BASE_CACHE_IMAGE AGENT_DEV_BASE_CACHE_IMAGE
 GENERATED_ARTIFACT_PATHS := pkg/protos pkg/openapi_client web_src/src/api-client agent/src/superplaneapi api/swagger/superplane.swagger.json agent/src/usage_pb2.py agent/src/private/agents_pb2.py agent/src/private/agents_pb2_grpc.py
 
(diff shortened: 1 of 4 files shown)


edbienes:
## Bugbot findings triage

All four findings have been reviewed and none require code changes:

- **Cache image tags (Medium)**: Intentional — Podman/Buildah rejects tags in `--cache-from` by design (`repository must contain neither a tag nor digest`). The conditional in the Makefile correctly uses untagged refs for Podman and tagged refs for Docker.

- **Agent Makefile missing port exports (Medium)**: Not a regression — the compose file's self-contained nested defaults (`${PORT:-${PUBLIC_API_PORT:-8000}}`) handle port resolution. The agent Makefile already exports the cache image variables it needs.

- **Port exports override .env (High)**: Already fixed in 8559c03. The Makefile no longer exports `PORT`, `API_PORT`, `GRPC_SERVER_ADDR`, or `AGENT_HTTP_PORT`.

- **Nested interpolations block Podman (High)**: No longer applicable — `podman-compose` merged full nested interpolation support in March 2026 (podman-compose PR #1419), matching the Docker Compose spec. The `${VAR:-${VAR2:-default}}` syntax works in both engines on current versions.

shiroyasha:
@edbienes thank you for opening this PR! The code does look good, but we won't proceed with merging it.

The maintainers are all using docker only. I don't want to increase the surface area of what needs to maintained at this point in time.

edbienes:
> @edbienes thank you for opening this PR! The code does look good, but we won't proceed with merging it.
> 
> The maintainers are all using docker only. I don't want to increase the surface area of what needs to maintained at this point in time.

@shiroyasha no worries, all good. Any updates on PR #4306?

PR #4333, opened 2026-04-22
fix: Show feedback on API token copy

### LINKED ISSUE #3194: "Copy" button for API tokens doesn't indicate success on click
Normally clicking such buttons results in a visual effect and microcopy 'Copied to clipboard' somewhere.

matheuskafuri:
Implements #3194.

  ## What
  The copy buttons on newly generated API tokens on Profile, Service Account
  Detail (Regenerate), and Service Account Create relied solely on a toast
  to confirm the copy.

  ## Change
  - Extended the shared `<CopyButton>` to support a labeled variant so it can
    replace the existing `<Button variant="outline">Copy</Button>` pattern
    without changing the visual footprint. On click the button briefly
    toggles to "Copied!" with a green check for 2s.
  - Swapped all three API-token copy sites to use it.
  - Added `aria-label` / `aria-live` so screen readers announce the state change.
  - Added a small test suite for the component (labeled toggle, icon-variant
    aria-label, clipboard-failure path).
  - Dropped the success toast. The button's in-state feedback is now the
    primary confirmation; kept the error toast via an `onCopyError` callback
    so callers can still surface failures in their own voice.

  ## Why this shape
  The inline `clipboard.writeText + toast` pattern was duplicated across
  Profile.tsx, ServiceAccountDetail.tsx, and ServiceAccounts.tsx. Rather
  than patch each instance, extending the existing shared component fixes
  #3194 consistently and leaves other callsites (Members invite link,
  IntegrationDetails / Dash0 webhook URLs) a clean upgrade path. I kept
  those out of scope since they copy non-token values and aren't what the
  issue describes.

  ## Demo

https://github.com/user-attachments/assets/2371d3fb-01a1-4715-977b-1aa36a79763a

https://github.com/user-attachments/assets/5ad31128-9d4a-4030-8f4c-8e562118afe2

https://github.com/user-attachments/assets/ef9c9ae5-c931-44d0-a8cd-3ab8516c3170

  ## Notes
  - Fixed a small latent issue while in the file: `CopyButton`'s `setTimeout`
    had no unmount cleanup.
  - The icon-only variant (used in the canvas CLI panel) is untouched.


matheuskafuri:
Addressed the test-regex looseness and restored copiedLabel for idle/copied label symmetry. Left the onCopy callback off intentionally (the button's state change is the success signal) and kept aria-live on the button for now. Happy to extract it into a separate role="status" region if you'd prefer that shape.

matheuskafuri:
The E2E Tests 3/10 failure on this run is a Docker Hub anonymous-pull rate limit (toomanyrequests during
test.setup.build) ([logs](https://superplanehq.semaphoreci.com/jobs/a276a640-5e60-4a4d-a03c-7ca239587928))... not related to the change. Happy to push an empty commit to retry if another run would help, or defer to you on the Semaphore side.

shiroyasha:
I love this! I'll attempt to resolve conflicts.

shiroyasha (approved):
LGTM — duplicate import fix verified locally; CI green (9252 tests).

PR #4334, opened 2026-04-22
chore: add canvas spec/metadata updates under version versioning

### LINKED ISSUE #4303: Treat metadata changes as versioned updates
### Issue

The CLI `update` command currently treats metadata updates (name, description, change management) as separate from other changes.

This creates inconsistency:
- Metadata changes are handled differently from flow/spec changes
- Changes are not uniformly tracked
- Users and agents must reason about multiple update paths

---

### Goal

Unify all publishes under a single model where:

- Metadata changes are treated the same as flow/spec changes
- Every update results in a **new version**
- All changes are consistently tracked in run history

---

### Principles

- There is no distinction between “metadata” and “spec” updates
- Any change (name, description, change management, flow) is a **versioned change**
- Updates should produce a clear, traceable history
- Avoid multiple update mechanisms with overlapping responsibilities

---

### Outcome

- Single mental model: **all changes = version updates**
- Metadata changes are fully traceable
- Consistent and auditable run history
- Reduced ambiguity and safer updates

forestileao:
## Summary

Unifies canvas updates under versioning only.

- Treats metadata changes the same as flow/spec changes
- Stores canvas name, description, and change management settings on `workflow_versions`
- Makes every canvas update produce a versioned change with traceable history

## Changes
- Moved canvas metadata fields from `workflows` to `workflow_versions`
- Updated CLI and UI flows to always update through canvas versions
- Removed the leftover `update_canvas.go` file and moved shared validation to `common.go`

Closes: https://github.com/superplanehq/superplane/issues/4303


forestileao:
Closing by now, there is some thi gs I didnt catch before

web_src/src/pages/canvas/settings/CanvasSettingsForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+    async (values: SettingsSavePayload) => {
+      await updateCanvasVersionMutation.mutateAsync({
+        versionId,
+        name: values.name,
+        description: values.description,
+        nodes: canvas.spec?.nodes || [],
+        edges: canvas.spec?.edges || [],
+        changeManagement: values.changeManagement,
+        preserveLocalCanvasState: true,
+      });
```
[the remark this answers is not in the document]
forestileao: not true

pkg/grpc/actions/canvases/update_canvas_version.go
```diff
@@ -153,7 +173,28 @@ func UpdateCanvasVersionWithUsage(
...
+			return status.Error(codes.InvalidArgument, "canvas name is required")
+		}
+
+		findErr := ensureCanvasNameAvailableInTransaction(tx, organizationUUID, canvasUUID, nextName)
+		if errors.Is(findErr, models.ErrCanvasNameAlreadyExists) {
+			return status.Errorf(codes.AlreadyExists, "Canvas with the same name already exists")
+		}
+		if findErr != nil {
+			return findErr
+		}
```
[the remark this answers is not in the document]
forestileao: it's not common to have many people editing all the time the same canvas. All the time

pkg/grpc/actions/canvases/publish_canvas_change_request.go
```diff
@@ -142,6 +149,7 @@ func PublishCanvasChangeRequest(
 
 		mergedVersion, createVersionErr := models.CreateCanvasSnapshotVersionInTransaction(
 			tx,
+			version,
```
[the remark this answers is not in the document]
forestileao: that is no problem, just spec fields

forestileao:
👍

db/structure.sql
```diff
@@ -615,7 +615,11 @@ CREATE TABLE public.workflow_versions (
     edges jsonb DEFAULT '[]'::jsonb NOT NULL,
     created_at timestamp without time zone NOT NULL,
     updated_at timestamp without time zone NOT NULL,
-    state character varying(32) NOT NULL
+    state character varying(32) NOT NULL,
+    name character varying(128) DEFAULT ''::character varying NOT NULL,
+    description text DEFAULT ''::text NOT NULL,
+    change_management_enabled boolean DEFAULT false NOT NULL,
+    change_request_approvers jsonb DEFAULT '[{"type": "anyone"}]'::jsonb NOT NULL
```
lucaspin: Not sure an `anyone` approver is the best one to be the default on the DB level. Maybe just an empty list here?

db/migrations/20260422175935_canvas-version-metadata.up.sql
```diff
@@ -0,0 +1,30 @@
...
+SET
+  name = COALESCE(w.name, ''),
+  description = COALESCE(w.description, ''),
+  change_management_enabled = COALESCE(w.change_management_enabled, false),
+  change_request_approvers = COALESCE(w.change_request_approvers, '[{"type":"anyone"}]'::jsonb)
+FROM public.workflows AS w
+WHERE w.id = v.workflow_id;
+
+ALTER TABLE public.workflows
+  DROP CONSTRAINT IF EXISTS unique_canvas_in_organization;
```
lucaspin: Is it OK to drop the unique name constraint here? Not sure we are relying on this on other parts of the system. API / CLI had endpoints / commands that accepted a `id_or_name`. Could this interfere there?

pkg/grpc/actions/canvases/common.go
```diff
@@ -0,0 +1,137 @@
...
+	name string,
+) error {
+	if err := tx.
+		Model(&models.Organization{}).
+		Clauses(clause.Locking{Strength: "UPDATE"}).
+		Where("id = ?", organizationID).
+		First(&models.Organization{}).
+		Error; err != nil {
+		return err
+	}
```
lucaspin: Why are we making this query?

pkg/grpc/actions/canvases/common.go
```diff
@@ -0,0 +1,137 @@
...
+
+	publisher, err := changesets.NewCanvasPublisher(tx, nextVersion, liveVersion, options)
+	if err != nil {
+		return err
+	}
+
+	return publisher.Publish(ctx)
+}
+
+func buildCanvasPublisherOptions(
```
lucaspin: Do we really need this function? Seems like an extra level of indirection to just return a changesets.CanvasPublisherOptions struct is unnecessary

pkg/grpc/actions/canvases/create_canvas.go
```diff
@@ -67,11 +72,33 @@ func CreateCanvas(
...
+			if validateErr := validateCanvasChangeRequestApprovers(
+				authService,
+				organizationID.String(),
+				approvers,
+			); validateErr != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "invalid change request approval config: %v", validateErr)
+			}
+			changeRequestApprovers = approvers
+		}
+	}
```
lucaspin: Four levels of nesting here, simplify

pkg/grpc/actions/canvases/update_canvas.go
```diff
@@ -34,150 +35,254 @@ func UpdateCanvas(
...
-				return nil, status.Errorf(codes.InvalidArgument, "invalid change request approval config: %v", validateErr)
-			}
-
-			if !slices.EqualFunc(canvas.ChangeRequestApprovers, approvers, func(left, right models.CanvasChangeRequestApprover) bool {
-				return left.Type == right.Type && left.User == right.User && left.Role == right.Role
-			}) {
-				canvas.ChangeRequestApprovers = approvers
-				changed = true
-			}
+func findWritableCanvas(organizationUUID, canvasID uuid.UUID) (*models.Canvas, error) {
```
lucaspin: I think this findWritableCanvas makes this less readable than having this code inlined in the caller. It adds one extra layer of indirection (you have to go into the function) and it introduces a "writable" quality to a canvas that does not exist nowhere else in the system. It's also not reused anywhere else, so I'd just inline it.

pkg/grpc/actions/canvases/update_canvas_version.go
```diff
@@ -153,7 +173,28 @@ func UpdateCanvasVersionWithUsage(
 			return draftErr
 		}
 
+		nextName := strings.TrimSpace(pbCanvas.GetMetadata().GetName())
+		if nextName == "" {
+			return status.Error(codes.InvalidArgument, "canvas name is required")
+		}
+
+		findErr := ensureCanvasNameAvailableInTransaction(tx, organizationUUID, canvasUUID, nextName)
```
lucaspin: We only need the extra query to check if name is available if the name changed

pkg/models/canvas.go
```diff
@@ -11,15 +12,17 @@ import (
...
 type Canvas struct {
 	ID                      uuid.UUID
 	OrganizationID          uuid.UUID
 	LiveVersionID           *uuid.UUID
 	IsTemplate              bool
-	ChangeManagementEnabled bool
-	ChangeRequestApprovers  datatypes.JSONSlice[CanvasChangeRequestApprover]
-	Name                    string
-	Description             string
+	Name                    string                                           `gorm:"column:name;->"`
```
lucaspin: A comment explaining what this `->` notation means would be nice. I had to look that up 😅

db/migrations/20260422175935_canvas-version-metadata.up.sql
forestileao: Dropping the old constraint on workflows(name) makes sense mechanically, because name no longer lives on workflows.
 
 Although, we still verify canvas name uniqueness in the application layer through ensureCanvasNameAvailableInTransaction(...) in pkg/grpc/actions/canvases/
  common.go.

pkg/grpc/actions/canvases/common.go
forestileao: Once the DB constraint was removed, that lock became the thing preventing concurrent duplicate-name races across transactions

forestileao:
@lucaspin comments addressed. Kept two comments unresolved so you can check if my answer makes sense

pkg/grpc/actions/canvases/common.go
lucaspin: Yeah, this one concerns me a bit. What if we keep the `name` column in the canvas table and the unique constraint, and we keep it updated with the name of the live version? We already have to update `live_version_id` there anyway

forestileao: That would work too. It's reasonable

pkg/grpc/actions/canvases/common.go
```diff
@@ -0,0 +1,131 @@
...
+	existingCanvas, err := models.FindCanvasByNameInTransaction(tx, name, organizationID)
+	if err == nil && existingCanvas.ID != canvasID {
+		return models.ErrCanvasNameAlreadyExists
+	}
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
forestileao: The current branch has workflows.name kept and workflows_organization_id_name_key UNIQUE (organization_id,
  name) present in db/structure.sql:1039. The migration also no longer drops workflows.name; it only drops
  description, change_management_enabled, and change_request_approvers.

  So the race described cannot produce duplicate live canvas names. Even if two transactions both pass
  ensureCanvasNameAvailableInTransaction, the DB unique constraint is the final authority and one transaction
  will fail when it writes workflows.name.

pkg/grpc/actions/canvases/create_canvas.go
```diff
@@ -67,11 +72,29 @@ func CreateCanvas(
...
+			organizationID.String(),
+			changeManagement,
+		)
+		if approversErr != nil {
+			return nil, approversErr
+		}
+		if approvers != nil {
+			changeRequestApprovers = approvers
+		}
+	}
```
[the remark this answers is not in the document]
forestileao: - CreateCanvas starts with the org default.
  - If the request includes spec.changeManagement.enabled, it overwrites the stored canvas-version value.
  - But effective enforcement uses pkg/grpc/actions/canvases/change_management.go:27: if org change management
    is enabled, it always returns true regardless of the canvas stored value.

forestileao:
@lucaspin I handled the comments and noticed it required frontend changes too.

This required to add more code to the PR, but I was not able to ship the fixes in separate PRs

lucaspin (approved):

PR #4362, opened 2026-04-23
fix: validate canvas edge output channels on API writes

### LINKED ISSUE #4318: CLI: silently broken flows when using wrong output channel (e.g., `default` instead of `success` for HTTP)
## Description

Each component has specific output channels (e.g., HTTP Request uses `success`/`failure`, If uses `true`/`false`). When building canvases via CLI YAML, if you specify the wrong channel in an edge (e.g., `channel: default` for an HTTP Request node), downstream nodes **silently never fire**. There's no error, no warning — the workflow just stops.

## Steps to reproduce

1. Create a canvas via CLI with HTTP Request → If edge using `channel: default`
2. Trigger a run
3. HTTP Request succeeds (RESULT_PASSED)
4. If node never executes — no error, no queue item, nothing

## Expected behavior

Either:
- The CLI should validate edge channels against the component's declared output channels during `canvases create`/`update`
- Or there should be a warning/error in the console when an event is emitted on a channel that no subscriber is listening to

## Workaround

Use `superplane index components --name <component>` to check available output channels before wiring edges.

## Impact

This is especially problematic for AI agents and CLI-first users who can't see the visual output channel hints that the UI provides. Debugging silent failures is time-consuming.

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/4318

## Summary

Reject canvas edges that reference an output channel the source node does not expose.

This prevents silent workflow stalls caused by invalid channel names in CLI/API canvas definitions, and returns a `400 InvalidArgument` instead.

## Changes

- Add shared canvas output-channel validation for source nodes
- Validate edge channels during full canvas parsing on create/update
- Validate edge channels during canvas changeset application
- Support validation for:
  - components
  - triggers (`default`)
  - blueprints
- Add a transaction-safe blueprint lookup helper for validation paths
- Add tests covering invalid channel rejection in:
  - canvas create
  - canvas version update
  - canvas patcher changesets

## Behavior

Before:
- Invalid edge channels could be persisted
- Downstream nodes would silently never execute

After:
- The API rejects invalid edge channels with `400`
- Error message includes the source node and requested channel

## Testing

- Ran `make format.go`
- Ran `make lint`
- Ran `make check.build.app`
- Ran `go test -run '^$' ./pkg/grpc/actions/canvases/... ./pkg/models/...`

Note: DB-backed tests in this environment could not run because the test database connection was not configured (`lookup port=: no such host`).

### DIFF
--- a/pkg/grpc/actions/canvases/changesets/output_channels.go
+++ b/pkg/grpc/actions/canvases/changesets/output_channels.go
@@ -1,6 +1,7 @@
 package changesets
 
 import (
+	"errors"
 	"fmt"
 	"slices"
 
@@ -11,6 +12,8 @@ import (
 	"gorm.io/gorm"
 )
 
+var errUnresolvableSourceNodeOutputChannels = errors.New("source node output channels are not resolvable")
+
 func ValidateSourceNodeOutputChannel(
 	tx *gorm.DB,
 	registry *registry.Registry,
@@ -20,6 +23,10 @@ func ValidateSourceNodeOutputChannel(
 ) error {
 	outputChannels, err := listSourceNodeOutputChannels(tx, registry, organizationID, sourceNode)
 	if err != nil {
+		if errors.Is(err, errUnresolvableSourceNodeOutputChannels) {
+			return nil
+		}
+
 		return fmt.Errorf("failed to resolve output channels for source node %s: %w", sourceNode.ID, err)
 	}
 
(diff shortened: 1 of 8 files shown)


forestileao:
👍

pkg/grpc/actions/canvases/serialization.go
```diff
@@ -283,10 +288,19 @@ func ParseCanvas(registry *registry.Registry, orgID string, canvas *pb.Canvas) (
 		if nodeTypeByID[edge.TargetId] == compb.Node_TYPE_WIDGET {
 			return nil, nil, status.Errorf(codes.InvalidArgument, "edge %d: widget nodes cannot be used as target nodes", i)
 		}
+
+		if err := changesets.ValidateSourceNodeOutputChannel(
+			database.Conn(),
```
forestileao: Not sure if I liked db calls here. Lets change it

pkg/grpc/actions/canvases/changesets/output_channels.go
```diff
@@ -0,0 +1,111 @@
...
+	if sourceNode.Type == models.NodeTypeComponent {
+		return listComponentOutputChannels(registry, sourceNode)
+	}
+
+	if sourceNode.Type == models.NodeTypeTrigger {
+		return []core.OutputChannel{core.DefaultOutputChannel}, nil
+	}
+
+	if sourceNode.Type == models.NodeTypeBlueprint {
+		return listBlueprintOutputChannels(tx, organizationID, sourceNode)
```
forestileao: N+1 problem

PR #4363, opened 2026-04-23
feat: add github.onPush path filter and filePathMatches() helper

### LINKED ISSUE #4275: github.onPush: path filter or expr helper
### Context
Filtering push events by changed file paths is a very common requirement.

### Observed behavior
Current pattern is awkward and fragile, e.g.:

```expr
indexOf(toJSON(root().data.commits ?? []), "pkg/integrations/") >= 0
```

This can false-match commit messages and is hard to read/maintain.

### Suggested improvement
Either:
1. Add `filePathMatches()` expression helper that inspects added/modified/removed arrays explicitly, or
2. Add native `paths` filter on `github.onPush` trigger config (similar to GitHub Actions).

### Why this matters
Cleaner path filters will reduce brittle expressions and improve correctness for CI/CD routing workflows.

0vertake:
## Summary

- Add optional paths on `github.onPush` as a list of glob strings (GitHub Actions–style, via doublestar): trigger runs only if some added / modified / removed path matches after applying ! excludes.
- Supports exclude-only lists by assuming an implicit include of ** when every entry is negated.
- Validates globs (invalid patterns are skipped with warnings); avoids “everything blocked” failure modes when includes are unusable while excludes still apply.
- Legacy paths stored as predicate maps: equals values are treated as literals; matches (regex) is skipped with a migrate warning. Decode strips paths before `mapstructure` so legacy payloads don’t break webhook handling.
- UI: string list field for paths; workflow mapper shows globs in trigger metadata where relevant.
- Docs: GitHub integration docs updated for globs, ! excludes, and migration note.

## Test Plan

- go test for `pkg/pathfilter` and `pkg/integrations/github/components/contents` (path filter + legacy decode coverage).
- Manual webhook: expose the dev API with ngrok (or equivalent), configure GitHub → On Push with refs/heads/main and paths.
- **Include + exclude:** directory glob plus ! pattern limiting a subtree (e.g. markdown-only); verify push touching only excluded files does not trigger, push touching other files does.
- **Exclude-only:** single ! subtree pattern; verify changes inside that subtree do not trigger, changes elsewhere do.

Closes #4275


pkg/configuration/any_predicate_list.go
```diff
@@ -62,3 +62,14 @@ func MatchesAnyPredicate(predicates []Predicate, value string) bool {
 
 	return false
 }
+
+// matchesAnyPredicateInList returns true if any value in values matches any predicate.
+func matchesAnyPredicateInList(predicates []Predicate, values []string) bool {
```
lucaspin: This seems to be unused?

pkg/exprruntime/filepath.go
```diff
@@ -0,0 +1,139 @@
...
+//
+// Returns true if any file (added, modified, or removed) across the given commits matches
+// the glob pattern. Supports:
+//   - **  matches any sequence of path components (including across directory separators)
+//   - *   matches any sequence of characters within a single path segment (no /)
+//
+// Example:
+//
+//	filePathMatches(root().data.commits, "pkg/**")
+func FilePathMatchesFunctionOption() expr.Option {
```
lucaspin: Let's not change the expression language for this. This is very specific to the `github.onPush` trigger, so let's keep it there for now.

pkg/integrations/github/on_push.go
```diff
@@ -6,15 +6,18 @@ import (
...
 	"github.com/superplanehq/superplane/pkg/core"
+	"github.com/superplanehq/superplane/pkg/exprruntime"
 )
 
 type OnPush struct{}
 
 type OnPushConfiguration struct {
 	Repository string                    `json:"repository" mapstructure:"repository"`
 	Refs       []configuration.Predicate `json:"refs" mapstructure:"refs"`
+	Paths      []string                  `json:"paths" mapstructure:"paths"`
```
lucaspin, line 19: Why not use `[]configuration.Predicate` for this too?

```golang
type OnPushConfiguration struct {
	Repository string                    `json:"repository" mapstructure:"repository"`
	Refs       []configuration.Predicate `json:"refs" mapstructure:"refs"`
	Paths      []configuration.Predicate `json:"paths" mapstructure:"paths"`
}
```

That already supports a `PredicateTypeMatches` that can be used here too. With that, we could re-use `configuration.MatchesAnyPredicate()`

pkg/configuration/any_predicate_list.go
```diff
@@ -62,3 +62,45 @@ func MatchesAnyPredicate(predicates []Predicate, value string) bool {
 
 	return false
 }
+
+// MatchesAnyPredicateInList evaluates predicates against a list of values.
+//
+// Semantics:
+//   - equals/matches: existential (any value can satisfy)
+//   - notEquals: universal (all values must differ from excluded value)
```
lucaspin: I'm having a really hard time wrapping my head around this 😅 

---

I can't think of a use case where I'd want to use `notEquals`. I think we should just not allow notEquals to be used in this case, only equals and matches. You can do that with:

```golang
  AnyPredicateList: &configuration.AnyPredicateListTypeOptions{
    Operators: []configuration.FieldOption{
      {
        Label: "Equals",
        Value: configuration.PredicateTypeEquals,
      },
      {
        Label: "Matches",
        Value: configuration.PredicateTypeMatches,
      },
    },
  },
},
```

shiroyasha:
How this maps to the YAML? Will I be able to do it like this:

```
- name: "On Documentation Changed"
  component: github.onPush
  paths:
     - 'docs/**/*.md'
```

```
- name: "On any change in billing service"
  component: github.onPush
  paths:
    - 'billing/**/*'
    - '!billing/**/*.md' # ignore documentation changes
```

(not sure if this last example is logically sound, but my intention is valid)

lucaspin:
@shiroyasha the idea was to reuse the [filtering pattern](https://github.com/superplanehq/superplane/blob/main/pkg/configuration/any_predicate_list.go) we already have for this too. So, the YAML would look like this:

```yaml
component: github.onPush
configuration:
  refs:
    - type: equals
      value: refs/heads/main
  paths:
    - type: matches
      value: docs\/.*\/.*
```

That `matches` operator uses RE2 regex syntax, which is the one supported by Go. Issues there are:
- That configuration field uses the `any-predicate-list` type, which returns true if any of the predicates you pass match.
- The negative case, which looks awful, since it's RE2, and you need to have the "everything under billing/, except documentation changes" logic in a single pattern:

```yaml
configuration:
  paths:
    - type: matches
      # Believe me, this means "everything under billing/, except documentation changes"
      # Codex is sure of it :)
      value: "(?s)^(?:billing/.+/\z|billing/.*(?:[^d]$|[^m]d$|[^.]md$))"
```

---

That being said, `any-predicate-list` type is a bit weird right now, and maybe we need a better way to express filters. But I'd like to have one way that is used consistently everywhere, and can be extended easily, other than having components implementing custom filtering logic.

lucaspin:
Dumping one idea here, just so I remove it from my head, but this shouldn't be done as part of this PR:

---

We could introduce a `filter` configuration field type that is more extensible than the current `any-predicate-list` type. A few examples:

```yaml
component: github.onPush
configuration:

  #
  # You can pick `any-predicate` or `all-predicates` to control the logic.
  # Here, I want to run on main or releases/* branches, so I use any logic.
  # These types of expressions are easier to define in the UI and YAML,
  # since you don't need to know what the event data looks like and write custom logic.
  #
  refs:
    filter: any-predicate
    predicates:
    - type: equals
      value: refs/heads/main
    - type: matches
      value: releases/v.*

  #
  # Same things as above, but using all logic with a `globMatches` operator, which is better than the RE2 regex one.
  #
  paths:
    filter: all-predicates
    predicates:
    - type: globMatch
      value: billing/**/*
    - type: globMatch
      value: "!billing/**/*.md"

  #
  # Filtering using an expr-lang/expr expression.
  # More powerful and flexible, but with great power comes great responsibility.
  # You need to know what the event data looks like.
  # Less readable than the predicates, harder to configure it in the UI and YAML.
  #
  paths:
    filter: expr
    expression: |
      let commits = $.commits ?? [];
      let files = flatten(map(commits, concat(.added ?? [], .modified ?? [], .removed ?? [])));
      let billingMd = `^billing/.*\.md$`;
      any(files, {
        let path = #;
        path startsWith "billing/" and not (path matches billingMd)
      })

  #
  # Since the `filter` field is an enum,
  # adding new ways to apply filters is just adding a new value there.
  # For example, JS-based filters could be done like this:
  #
  paths:
    filter: js-expr
    expression: |
      const changedFiles = $.commits.flatMap((commit) =>
        [].concat(commit.added ?? [], commit.modified ?? [], commit.removed ?? [])
      )

      return changedFiles.some(
        (file) => /^billing\//.test(file) && !/^billing\/.*\.md$/i.test(file),
      )
```

---

Created https://github.com/superplanehq/superplane/issues/4487 for this.

0vertake:
> How this maps to the YAML? Will I be able to do it like this:
> 
> ```
> - name: "On Documentation Changed"
>   component: github.onPush
>   paths:
>      - 'docs/**/*.md'
> ```
> 
> ```
> - name: "On any change in billing service"
>   component: github.onPush
>   paths:
>     - 'billing/**/*'
>     - '!billing/**/*.md' # ignore documentation changes
> ```
> 
> (not sure if this last example is logically sound, but my intention is valid)

Hey @shiroyasha, the paths field follows the same predicate format as refs, so the YAML would look like this:
```
 - name: "On Documentation Changed"
    component: github.onPush
    configuration:
      paths:
        - type: matches
          value: 'docs/.*\.md'

  - name: "On any change in billing service"
    component: github.onPush
    configuration:
      paths:
        - type: matches
          value: 'billing/.*'
```
The ! negation (!billing/**/*.md) isn't supported.

shiroyasha:
Thanks for sharing @lucaspin @0vertake.

I definitely prefer having double-star syntax compared to regex matches.

shiroyasha:
@lucaspin @0vertake I did one more fresh review from the start. 
The code looks good, but I would like us to take one more take on this.

- Would like to see glob patterns instead of regexes
- There is a justified reason to have some kind of negations when dealing with monorepos (e.g. https://github.com/superplanehq/superplane/pull/4363#issuecomment-4339669496)

0vertake:
> @lucaspin @0vertake I did one more fresh review from the start. The code looks good, but I would like us to take one more take on this.
> 
> * Would like to see glob patterns instead of regexes
> * There is a justified reason to have some kind of negations when dealing with monorepos (e.g. [feat: add github.onPush path filter and filePathMatches() helper #4363 (comment)](https://github.com/superplanehq/superplane/pull/4363#issuecomment-4339669496))

Sure, I can extend the PR to include this too

PR #4373, opened 2026-04-24
fix: record RESULT_FAILED on failure-channel routing

### LINKED ISSUE #4284: HTTP: RESULT_FAILED on failure channel
### Context
HTTP node attempted to fetch `<nil>.diff` (expression resolved to nil URL).

### Observed behavior
- Node correctly routed output to `failure` channel.
- Execution status still displayed `RESULT_PASSED`.

### Why this is a problem
Operational tooling and humans cannot reliably distinguish success from routed failures.

### Suggested improvement
When HTTP emits `failure`, execution should report failed status (`RESULT_FAILED`) or explicit alternate status (e.g. `RESULT_ROUTED_FAILURE`) rather than `RESULT_PASSED`.

Vukotije:
Closes #4284 

## Summary
- When a component routes a payload to its failure channel (e.g. HTTP's "failure", Merge's "timeout"/"fail", SSH's "failed"), the execution is now recorded as `RESULT_FAILED` instead of `RESULT_PASSED`. The channel event still fires so downstream nodes continue to run.
- **Failure semantics are declared once, centrally**, via a curated `failureLabels` vocabulary in `pkg/core/component.go` ("Failure", "Failed", "Fail", "Timeout"). Every component already using one of these labels gets the correct behavior for free — no per-component declarations changed. This covers HTTP, SSH, Merge, and integrations like DigitalOcean, Render, Cloudflare, GitLab, CircleCI, Harness, GitHub, Daytona, SendGrid, Slack/Telegram wait-for-button, and more.
- **New `CanvasNodeExecution.FailWithOutputsInTransaction`** emits events AND marks the row failed in the same tx. Event-creation and node-ready-transition were extracted to shared private helpers (`createOutputEvents`, `markNodeReady`) to avoid duplicating code with `PassInTransaction`.
- **`ExecutionStateContext` now takes the owning `core.Component`** and resolves channels lazily on first `Emit`, matching the existing context pattern (metadata contexts / integration subscription context hold their data source, not pre-resolved data).
- **Parent execution propagation:** the emit-with-failure path deliberately does **not** bubble failed status up to a parent execution — the emitted event drives parent aggregation via the existing completion path, so bubbling would double-count. Pure `Fail()` (no-emit) retains its bubble-up behavior.

## Design notes
- **Considered a separate `IsFailure bool` field** on `OutputChannel` and a dedicated `EmitFailure` verb on the context; both added surface area to express something already encoded in the Label. The chosen approach (central vocabulary in core) is smaller, self-documenting, and automatically covers every integration using a known failure label.
- **Considered a distinct `RESULT_ROUTED_FAILURE` status.** That's a proto/OpenAPI/UI change touching ~35 files and breaks old SDK clients' enum validation. Rejected as out-of-scope for this fix; can be introduced later if operators need to distinguish "node crashed" from "node routed failure payload".
- **Exact-match on labels is intentional:** case variants or whitespace won't silently classify as failure — the failure-path test for that component fails loudly instead.

## Test plan
- [x] `go test ./pkg/core/...` — new `TestOutputChannel_IsFailure` classifies every label currently in the codebase (25+ rows, incl. case and whitespace variants).
- [x] `go test ./pkg/components/http/...` — new `TestHTTP__Execute__Non2xxRoutesToFailureChannelAndRecordsFailed` asserts failure-channel event AND `RESULT_FAILED`.
- [x] `go test ./pkg/workers/contexts/...` (requires DB) — 5 subtests, incl. "failure channel does not bubble failed status to parent".
- [x] `go test ./pkg/components/merge/...` (requires DB) — existing `AssertExecutionFailed` flipped to expect `RESULT_FAILED`; merge stop-if and timeout tests exercise the new path end-to-end.
- [ ] **Manual smoke:** HTTP node with an unresolvable URL (`<nil>.diff` as in the original report) → execution shows `failed` in UI/API while the failure-channel downstream still runs.


forestileao (commented):
@Vukotije  I think this PR addresses something slightly different from what the task described.

The reported issue was:

```text
Observed behavior
- Node correctly routed output to failure channel.
- Execution status still displayed RESULT_PASSED.
```

So the output channel behavior is already correct. The confusing part is the `result` field exposed by the API here:
[https://github.com/superplanehq/superplane/blob/main/protos/canvases.proto#L834-L835](https://github.com/superplanehq/superplane/blob/main/protos/canvases.proto#L834-L835)

Conceptually, the current implementation makes sense, but it can confuse the AI agent because `result != outputChannel`.

One possible fix would be to make `ctx.Fail` emit a payload with `status.reason = OK`:
[https://github.com/superplanehq/superplane/blob/main/protos/canvases.proto#L861](https://github.com/superplanehq/superplane/blob/main/protos/canvases.proto#L861)

However, since we also use `RESULT_FAILED` for internal errors, the better fix may be to update our SuperPlane Skill repo so the AI agent understands that `result` and `outputChannel` are separate concepts and should not always be interpreted as equivalent.


cc: @lucaspin

PR #4374, opened 2026-04-24
Changed simple if statement into single line

ivanmaurovic7:


markoa:

PR #4378, opened 2026-04-24
chore: upgrade Go to 1.26.2

lucaspin:
I need 1.25.8 in order to use a new package I need for the new integration setup flow, so I'm taking the opportunity to upgrade to the latest version.


lucaspin:
👍

PR #4380, opened 2026-04-24
feat: use checkboxes to render multi-select field

lucaspin:
For integration setup flow, I want an element that displays a bunch of options and the user can on / off them. Since we already have a multi-select element, just added a type option to render it differently.

For example, this is how the the `semaphore.onPipelineDone` trigger result filter looks like with it:

Screenshot 2026-04-24 at 17 07 14

---

This PR is just adding the ability, not updating any components to use it. To me, this looks a lot better than the current default, but I didn't want to update anything else yet. We can make it the default behavior later on.


lucaspin:
👍

PR #4382, opened 2026-04-24
fix: accept X-Hub-Signature-256 for webhook HMAC auth

### LINKED ISSUE #4381: Webhook HMAC auth doesn't support X-Hub-Signature-256 (GitHub compatibility)
## Problem

When configuring a GitHub webhook to send events to a SuperPlane webhook trigger using **Signature (HMAC)** authentication, the request is rejected with a **403**.

GitHub sends the HMAC signature in the header:
```
X-Hub-Signature-256: sha256=<hash>
```

But SuperPlane expects the signature in:
```
X-Signature-256: sha256=<hash>
```

These header names don't match, so HMAC verification always fails for GitHub webhooks.

## Expected behavior

SuperPlane should support `X-Hub-Signature-256` as a recognized HMAC signature header, or provide a configuration option to specify which header to read the signature from.

## Workaround

Currently the only workaround is to disable authentication entirely ("None") or set up a proxy that rewrites the header before forwarding to SuperPlane.

## Steps to reproduce

1. Create a webhook trigger in SuperPlane with **Signature (HMAC)** authentication
2. Add the SuperPlane webhook URL + secret to a GitHub repo/org webhook
3. Trigger a GitHub event (e.g. `issues`, `projects_v2_item`)
4. Observe 403 response — GitHub's `X-Hub-Signature-256` header is not recognized

amriksingh0786:
## Summary

- Webhook HMAC auth previously only read the signature from `X-Signature-256`; GitHub sends it in `X-Hub-Signature-256`, so every GitHub webhook with signature auth failed with a 403.
- Fall back to `X-Hub-Signature-256` when `X-Signature-256` is absent, so GitHub webhooks work out of the box without breaking senders using the original header.
- Documentation updated to mention both accepted headers.

Closes #4381

## Test plan

- [x] `pkg/triggers/webhook` unit tests pass (`docker compose run --rm --no-deps app go test -count=1 ./pkg/triggers/webhook/...`)
- [x] New tests cover: valid signature via `X-Hub-Signature-256`, invalid signature via `X-Hub-Signature-256`, and precedence (`X-Signature-256` wins when both are present)
- [x] Existing "missing signature" / invalid-format / invalid-signature / valid-via-`X-Signature-256` cases still pass
- [ ] Manual: point a real GitHub webhook (signature auth) at a SuperPlane webhook trigger and confirm a test delivery now returns 200

### DIFF
--- a/pkg/triggers/webhook/webhook.go
+++ b/pkg/triggers/webhook/webhook.go
@@ -16,6 +16,8 @@ import (
 const (
 	MaxEventSize           = 64 * 1024
 	DefaultHeaderTokenName = "X-Webhook-Token"
+	SignatureHeader        = "X-Signature-256"
+	GitHubSignatureHeader  = "X-Hub-Signature-256"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


markoa:

PR #4384, opened 2026-04-24
feat: daytona surface bootstrap logs and add configurable timeout 

### LINKED ISSUE #4282: Daytona: bootstrap logs & timeout UX
### Context
`daytona.createRepositorySandbox` bootstrap can remain in `bootstrapping` for long periods.

### Observed behavior
Example run stayed bootstrapping for 13+ minutes despite timeout set to 5 minutes.
Execution view showed only `STATE_STARTED` / `RESULT_UNKNOWN` with no bootstrap logs and no clear progress.
`finishedAt` remained empty, making it unclear whether task was still running, hung, or polling failed.

### Suggested improvement
- Surface bootstrap script logs (streaming or periodic chunks).
- Enforce configured timeout strictly (e.g. fail at 300s if `timeout: 300`).
- Show elapsed vs timeout progress indicator (e.g. "5m23s / 5m00s").

### Why this matters
Improves operability and prevents long blind waits in agent-driven flows.

Vukotije:
## Summary

Closes #4282.

`daytona.createRepositorySandbox` could sit in `bootstrapping` for many minutes with `STATE_STARTED / RESULT_UNKNOWN`, no logs, no progress indicator, and no honored timeout. Agent flows had no way to tell whether the run was making progress, hung, or had silently failed.

This PR delivers the three asks from the issue:

- **Bootstrap logs** stream into the execution sidebar (~2 s lag).
- **Configurable timeout** (minutes, default 5, max 60), nested under the Bootstrap toggle next to the script.
- **Elapsed vs timeout** indicator (`23s / 1m`) in the sidebar.

## What changed

**Backend** (`pkg/integrations/daytona/create_repository_sandbox.go`)
- New `bootstrap.timeout` config field, validated in `Setup`.
- Per-poll `GetSessionCommandLogs` capture into `metadata.Bootstrap.Log`, tail-trimmed to 64 KiB. Re-fetched once after exit so trailing output is not lost.
- `Bootstrap.FinishedAt` set on timeout for a definite terminal timestamp.
- Inline scripts normalized to LF before upload — fixes dash failures from CRLF (`sleep 2\r`, `done\r`).

**Frontend**
- `daytona/create_repository_sandbox.ts` — adds `Elapsed` and `Bootstrap log`  to execution details. Placement matches existing convention (SSH puts stdout/stderr in the Details panel the same way).
- `chainItem/MultilineDetail.tsx` + `ChainItem.tsx` — multiline detail values render as a scrollable monospace block instead of being silently collapsed and truncated. Generic; also fixes SSH stdout/stderr rendering.

## Tests

- Backend: 8 new subtests (configured timeout, in-flight log capture, tail-trim, transient log-fetch failure, FinishedAt-on-timeout, post-exit log re-fetch, CRLF normalization, Setup validation).
- Frontend: 4 new mapper tests; full mapper suite 265/265.

## UI

Config:
image

Working:
image

Sucess:
image

Timeout:
image

## Test plan

- [x] `go test ./pkg/integrations/daytona/...`
- [x] `vitest run src/pages/workflowv2/mappers/`
- [x] `tsc -b`, `gofmt`, `go vet` clean
- [x] End-to-end against real Daytona: multiline script logs cleanly, 1min timeout fails at deadline, post-exit final line appears.


pkg/integrations/daytona/create_repository_sandbox.go
```diff
@@ -235,6 +248,14 @@ func (c *CreateRepositorySandbox) Configuration() []configuration.Field {
...
 							},
 						},
+						{
+							Name:        "timeout",
+							Label:       "Timeout",
+							Type:        configuration.FieldTypeNumber,
+							Required:    false,
+							Description: "Time in minutes the bootstrap script may run before it is failed. Measured from when the script starts; sandbox creation and clone have a separate fixed timeout.",
+							Default:     int(CreateRepositorySandboxDefaultTimeout.Minutes()),
+						},
```
forestileao, line 258: You can also use TypeOptions to tell the UI or the Agent that there is a Min and optionally max timeout Value.

example from graphql.go:
```go
{
            Name:        "timeoutSeconds",
			Type:        configuration.FieldTypeNumber,
			Label:       "Timeout (seconds)",
			Description: "Timeout in seconds for the request",
			Default:     10,
				Number: &configuration.NumberTypeOptions{
					Min: &minTimeout,
					Max: &maxTimeout,
				},
			},
```

lucaspin:
@Vukotije I like how you made it look in the UI. However, I'm not so sure about storing this in the execution metadata. I'd vote to just let this one sit out for a bit until we have a better primitive for this kind of data

pkg/integrations/daytona/create_repository_sandbox.go
```diff
@@ -415,22 +445,36 @@ func (c *CreateRepositorySandbox) poll(ctx core.ActionHookContext) error {
...
+		// bootstrap command was dispatched, so the budget begins when the
+		// script actually starts running — not at sandbox creation.
+		bootstrapStartedAt, err := time.Parse(time.RFC3339, metadata.Bootstrap.StartedAt)
+		if err != nil {
+			return fmt.Errorf("failed to parse bootstrap started at: %v", err)
+		}
+		timeout := time.Duration(metadata.Timeout) * time.Second
+		if time.Since(bootstrapStartedAt) > timeout {
+			c.markBootstrapTimedOut(ctx, &metadata)
+			ctx.Logger.Errorf("bootstrap failed after %v", timeout)
```
forestileao, line 475: add the logic inside case clauses into separate functions just like it was before

Vukotije:
> @Vukotije I like how you made it look in the UI. However, I'm not so sure about storing this in the execution metadata. I'd vote to just let this one sit out for a bit until we have a better primitive for this kind of data

@lucaspin that's why i mentioned a possible new view or similar alternative in dev chat, would you want me to further explore options of this because i agree it semantically isn't the best place for logs, or would you prefer me to hold on until one of the mentors figures further steps?

lucaspin:
@Vukotije since this is not a big issue right now, let's put it on hold, and focus on other tasks. This primitive for storing logs-like data will appear in the runners/sandboxes effort, so we can figure out how this will look like there. Once that is incorporated into the system, we can use it here too. cc @shiroyasha

bender-rodriguez-unit1:

PR #4387, opened 2026-04-25
feat: render canvas memory urls as clickable links

### LINKED ISSUE #4361: Canvas Memory UI: render URL values as clickable links
## Problem

Canvas Memory values that contain URLs (e.g., `https://github.com/org/repo/pull/49` or `http://104.236.219.19.sslip.io`) are rendered as plain text. Users have to manually copy-paste them into a browser.

For dashboards like ephemeral environment tracking, the preview URL and PR link are the most important fields — but they are not clickable.

## Suggestion

Auto-detect values that look like URLs (start with `http://` or `https://`) and render them as clickable hyperlinks in the Canvas Memory table.

This is a frontend rendering change:
- No schema or storage changes
- Simple pattern match on display
- Only affects how values are rendered, not how they are stored

## Context

Ephemeral environments canvas stores `url` (preview link) and `pr_url` (GitHub PR link) in memory. These are the primary fields users interact with — making them clickable turns the memory table into a useful dashboard.

abdusabri:
Closes #4361

image

## Summary

* Canvas memory now renders HTTP/HTTPS string values as clickable links instead of plain text.
* The modal reuses the shared URL helper from the frontend utilities for the URL check.
* External link styling in the canvas memory modal is aligned with the project’s link colors.
* Added frontend tests covering both URL rendering and non-URL plain text rendering.
* Removed an unintended pre-existing red background from the scalar Canvas Memory row renderer. The red styling was introduced with the modal’s initial implementation, was not tied to any error state, and made normal string memory entries look broken (screenshot below, before removing it).

image copy

### DIFF
--- a/web_src/src/pages/workflowv2/CanvasMemoryModal.tsx
+++ b/web_src/src/pages/workflowv2/CanvasMemoryModal.tsx
@@ -1,6 +1,8 @@
 import { Button } from "@/components/ui/button";
 import { Dialog, DialogContent } from "@/components/ui/dialog";
 import type { CanvasMemoryEntry } from "@/hooks/useCanvasData";
+import { isUrl } from "@/lib/utils";
+import type { ReactNode } from "react";
 import { Trash2 } from "lucide-react";
 
 export type CanvasMemoryModalProps = {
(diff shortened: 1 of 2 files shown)


web_src/src/pages/workflowv2/CanvasMemoryModal.tsx
```diff
@@ -132,6 +134,24 @@ function collectColumns(items: Record<string, unknown>[]): string[] {
...
+  if (typeof value !== "string" || !isUrl(value)) {
+    return formattedValue;
+  }
+
+  return (
+    <a
+      href={value}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="text-blue-600 underline hover:text-blue-700 dark:text-blue-400 dark:hover:text-blue-300"
```
abdusabri, line 148: I had done this following some other examples in the code, like the first image below. However, I now realize that there is no actual user-selectable option to have a dark mode/theme. So, let me know if I should keep the dark variants or remove them :)

Image

------------------------------------

I could only trigger dark mode/theme using dev tools (applied a class called `dark-mode-disabled`), but that looks like an incomplete experiment/feature or so, and it does not apply to the canvas even in that case

Image

shiroyasha:

Unfortunately, we haven't reviewed this on time and since than the application diverged significantly.

I'll close the PR for now. Thanks again for taking the time to contribute.

Issue #1837, opened 2026-01-26
Unify component naming across website/docs/product/code


shiroyasha:
Touched this topic with @markoa and @AleksandarCole this morning while writing docs.

**Component** = The boxes that you can use on the canvas
**Trigger** = A type of component that starts runs
**Action** = A type of component that does actions
**Core Components** = Built-in triggers and actions
**Integration** = An integration with an external service.
**Canvas** = Place where you build stuff 

What needs to be checked and updated:

- [x] The repository docs still sometimes use the name application
- [x] The product uses the term application
- [x] The product uses the term "Primitives" instead of "Core"
- [x] The product separates out "Primitives" and "Triggers", should be just "Core"
- [x] The product docs are referring to "triggers and components"
- [x] The code uses the term "application" for integrations
- [x] The code uses the term "workflows" for canvases
- [x] The code uses the term "components" for actions

markoa:
related: https://github.com/superplanehq/docs/issues/17

PR #4393, opened 2026-04-25
chore: rename core.Action to core.Hook

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/1837

---

Rename `core.Action` to `core.Hook`, so we can later on use the action term for what is currently called a component in our core.


lucaspin:
👍

PR #4395, opened 2026-04-25
chore: rename hook API endpoints

lucaspin:
Now that we have core.Hook (see https://github.com/superplanehq/superplane/pull/4393), we should rename the APIs to use /hooks instead.


lucaspin:
👍

PR #4396, opened 2026-04-25
chore: rename core.Component to core.Action

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/1837

---

Now that we renamed core.Action to core.Hook - https://github.com/superplanehq/superplane/pull/4393 - we can finally rename core.Component to core.Action


lucaspin:
👍

PR #4402, opened 2026-04-25
fix: remove positional arg from canvases update

### LINKED ISSUE #4231: Canvases update -f without file id
Currently the CLI requires this:

```
superplane canvases update [id-or-name] -f <path-to-yaml>
```

We want to remove the [id-or-name], and just have this:

```
superplane canvases update -f <path-to-yaml>
```

Reasoning: ID is already part of the YAML and the current format is confusing the agents.

---

### Reasoning from the agent:

**Problem:** For agent and scripted flows, `superplane canvases update -f <file.yaml>` effectively **requires `metadata.id` inside the YAML** (`pkg/cli/commands/canvases/file_loader.go` → error: *canvas metadata.id is required for update*). When the file omits id (portable spec, hand-edited snippets, or exports that do not carry id), users cannot update in place and resort to **delete + recreate**, which **breaks webhooks** and **loses canvas state**.

**Current vs proposed:** Current: With `--file`, the command resolves the canvas only from the file's `metadata.id`; a positional `<name-or-id>` is not combined with `-f` to identify the canvas. Proposed (direction): allow targeting the canvas by **argument or active canvas** while the file supplies spec (and require id match when both present)—exact UX TBD.

**Use case:** Agents and automation that iterate on canvas YAML without embedding stable canvas UUIDs in every file.

**Research (code pointers):**
- `pkg/cli/commands/canvases/update.go` — `Execute`: when `filePath != ""`, loads via `loadCanvasFromFile` only.
- `pkg/cli/commands/canvases/file_loader.go` — `loadCanvasFromFile` enforces non-empty `resource.Metadata.Id`.
- Related: unknown YAML field behavior / agent debugging — https://github.com/superplanehq/superplane/issues/3250

0vertake:
## Summary

- Removes the `[name-or-id]` positional argument from `superplane canvases update` entirely
- Removes the active-canvas fallback logic
- `--file` is now a required flag
- The YAML file's `metadata.id` is the sole source of truth for identifying the target canvas — if it is missing, the command returns a clear error

Before:
```
superplane canvases update [name-or-id] -f <path-to-yaml>
```

After:
```
superplane canvases update -f <path-to-yaml>   # YAML must include metadata.id
```

Closes #4231


lucaspin:
> Updates the canvas using --file. If metadata.id is omitted in the YAML, pass the canvas name or id, or set an active canvas with `superplane canvases active`. When the file includes metadata.id and a name-or-id argument is given, they must refer to the same canvas.

This is a bit confusing. I think we can completely remove the [name-or-id] argument here. I'd even remove the logic for grabbing it from the active canvas too, because I can see cases where that could lead to unintended behavior. I think the reason agents were getting confused with the update command is because the command looks like this:

```
Usage:
  superplane canvases update [name-or-id] [flags]

Flags:
      --auto-layout string             automatically arrange the canvas (supported: horizontal, disable)
      --auto-layout-node stringArray   node id seed for auto layout (repeatable)
      --auto-layout-scope string       scope for auto layout (full-canvas, connected-component)
      --draft                          keep the update as a draft instead of auto-publishing (required when change management is enabled)
  -f, --file string                    filename, directory, or URL to files to use to update the resource
  -h, --help                           help for update
```

Without that `[name-or-id]`, it becomes much clearer that the file you are passing is the source of truth for what you want to update.

shiroyasha:
Looks good. Merging.

PR #4404, opened 2026-04-26
feat: hide columns in canvas memory table

### LINKED ISSUE #4360: Canvas Memory UI: allow hiding columns in the memory table
## Problem

When a Canvas Memory namespace has many fields (e.g., 10 columns for an ephemeral environments tracker), the table is hard to read. There is no way to hide columns that are only needed for internal plumbing (like `droplet_id` or `created_at`) while keeping the user-relevant ones visible (`pr_title`, `url`).

## Suggestion

Add a column visibility toggle to the Canvas Memory table — checkboxes or a dropdown to show/hide individual columns. Similar to how most data table UIs work (Notion, Airtable, GitHub Projects).

This is a UI-only change:
- No storage or schema changes
- Column preferences can be stored in localStorage or per-user settings
- All data remains in the namespace, just hidden from view

## Current Workaround

Split data across multiple namespaces (`environments` for user-facing, `env-internal` for plumbing). Works but adds complexity to the canvas design — extra addMemory, deleteMemory, and readMemory nodes for each namespace.

## Context

Ephemeral environments canvas stores 10 fields per environment. Users only care about ~5 of them. Had to split into two namespaces to keep the table readable.

nikhil-singh-01:
## Summary

Adds a per-namespace **Columns** dropdown to the Canvas Memory modal so users can toggle column visibility — useful when a namespace stores many fields but only a few are user-relevant. Closes #4360.

  - New `useCanvasMemoryColumnVisibility(canvasId, namespace, allColumns)` hook backs the menu and persists hidden columns to `localStorage` under `canvasMemoryHiddenColumns:<canvasId>` (per-canvas,
  per-namespace). Stores *hidden* columns rather than visible, so newly added columns appear by default.
  - `CanvasMemoryModal` refactored: extracts `NamespaceCard`, `ObjectNamespaceTable`, `RawNamespaceTable`, and `ColumnVisibilityMenu` out of the previous monolithic `renderNamespaceTable`.
  - The Columns menu only renders for object-shaped namespaces; raw/scalar namespaces still render the single-column "Value" view, untouched.
  - When all columns are hidden, the table swaps to an "All columns are hidden." inline prompt with a "Show all" action so the table can't dead-end.
  - Drive-by: removes a leftover `bg-red-500` debug class on the raw-values wrapper that was bleeding through to scalar namespaces.

  ## Test plan

  - [x] `vitest` covers the new hook: hydration, toggle, hideAll/showAll, multi-namespace isolation, malformed-storage tolerance — 6/6 passing.
  - [x] `make check.build.ui` (tsc -b + vite build) passes.
  - [x] Manually verified in dev:
    - 10-column `environments` namespace toggles work; counter updates `(visible/total)`.
    - Hidden columns persist across hard reload via `localStorage`.
    - "Hide all" → empty-state with inline "Show all" → restored.
    - Toggles in one namespace don't affect others.
    - Non-object namespace (`raw_notes`) renders the single-column view with no Columns button.
Screenshot 2026-04-26 at 1 00 21 PM
Screenshot 2026-04-26 at 1 00 38 PM

Closes #4360.

### DIFF
--- a/web_src/src/hooks/useCanvasMemoryColumnVisibility.spec.ts
+++ b/web_src/src/hooks/useCanvasMemoryColumnVisibility.spec.ts
@@ -1,0 +1,71 @@
+import { act, renderHook } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import {
+  canvasMemoryHiddenColumnsStorageKey,
+  useCanvasMemoryColumnVisibility,
+} from "@/hooks/useCanvasMemoryColumnVisibility";
+
+const canvasId = "canvas-1";
+
+afterEach(() => {
+  window.localStorage.clear();
+});
+
+describe("useCanvasMemoryColumnVisibility", () => {
+  it("starts with all columns visible when nothing is stored", () => {
+    const { result } = renderHook(() => useCanvasMemoryColumnVisibility(canvasId, "ns", ["a", "b", "c"]));
+    expect(result.current.visibleColumns).toEqual(["a", "b", "c"]);
+    expect(result.current.hidden.size).toBe(0);
+  });
+
+  it("hydrates hidden columns from localStorage", () => {
+    window.localStorage.setItem(canvasMemoryHiddenColumnsStorageKey(canvasId), JSON.stringify({ ns: ["b"] }));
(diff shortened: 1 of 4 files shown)


shiroyasha:

Unfortunately, we haven't reviewed this on time and since than the application diverged significantly.

I'll close the PR for now. Thanks again for taking the time to contribute.

PR #4417, opened 2026-04-26
fix: github.runWorkflow uses run ID from API response

lucaspin:
GitHub API finally returns a run ID when running a new workflow. See [this discussion](https://github.com/orgs/community/discussions/9752). That change was released in [SDK v84](https://github.com/google/go-github/releases/tag/v84.0.0).

---

With that change, we no longer need:
- The `run-name` in the GH workflow YAML to include the execution ID
- To list the runs after dispatching it to find it. Now, we can just use the run ID in the response


lucaspin:
👍

PR #4424, opened 2026-04-27
fix: make organization user removal resilient to partial failures

framirez91:
This PR improves organization user removal by making it resilient to partial failures.

When a user is removed from an organization, the system now keeps authorization and membership state consistent. If role removal fails, the operation stops and any roles removed earlier in the same request are restored. If user soft-deletion fails after roles were removed, those roles are also restored.

The change also includes a typo fix in internal error messages (determing -> determining) and adds regression test coverage for the role-removal failure path.

Additionally, the README now includes a consistency note about this behavior in English, with an extra Spanish translation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,12 @@ We welcome your bug reports, ideas for improvement, and focused PRs.
 - Read the **[Contributing Guide](CONTRIBUTING.md)** to get started.
 - Issues: use GitHub issues for bugs and feature requests.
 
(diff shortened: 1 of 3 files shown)


framirez91:
Thanks! Happy to iterate on this if needed.

I focused on making the user removal process more resilient to partial
failures and ensuring consistency across steps.

Let me know if you'd like me to adjust anything or expand the approach.

markoa:

PR #4444, opened 2026-04-28
feat: catch syntax canvas expression errors at save time

forestileao:
## Problem

  Expression mistakes in node configurations (`{{ $['Wrong Name'].x }}`, `{{ root( }}`, trailing junk like `$['X'].y asdadasd   
  1234`) only surfaced at runtime when a node executed. Agent loops became publish → run → fail → fix instead of save → see
  error → fix.                                                                                                                  
                                                                  
  ## What this does

  Adds a static validation pass during canvas create/update. Runs inside `ParseCanvas` (so it covers both create and update) and
   folds errors into the existing per-node `ErrorMessage` aggregation — same pattern `validateNodeRef` already uses.
                                                                                                                                
  ## What's caught                                                

  | Field | Value | Result |
  |---|---|---|
  | `String` / `Text` | `echo {{ $['Missing'].x }}` | ✗ unknown node reference 'Missing' |
  | `String` / `Text` | `Hi {{ root( }}` | ✗ syntax error: unexpected token EOF |                                               
  | `String` / `Text` | `My value: {{ "x" + invalid() }}` | ✗ compile error: unknown name invalid |
  | Any field | `{{   }}` | ✗ expression body is empty |                                                                        
  | Any field | `{{ root(1) }}` | ✗ root() takes no arguments |   
  | Any field | `{{ previous('a') }}` | ✗ previous() depth must be an integer literal |                                         
  | Any field | `{{ memory.find('ns') }}` | ✗ memory.find() requires a namespace and matches |
  | `Expression` | `$["Build"].day asdadasd 2313312` | ✗ syntax error: unexpected token Identifier("asdadasd") |                
  | `Expression` | `$["Missing"].x` | ✗ unknown node reference 'Missing' |                                                      
                                                                                                                                
  ## What's deliberately not caught                                                                                             
                                                                  
  `FieldTypeExpression` values without `{{ }}` framing flow through `Build()` as literal strings (`ResolveTemplateExpressions`  
  short-circuits when no `{{ }}` is present, see `pkg/workers/contexts/node_configuration_builder.go`). The dominant runtime
  pattern — memory components storing literals like `value: "default"` or `value: "ok"` — keeps these as plain strings; they    
  never reach `expr.Compile`. So bare-value identifiers are not strictly compile-checked:

  | Field | Value | Result |                                                                                                    
  |---|---|---|
  | `Expression` | `default`, `ok`, `down`, `1 + 2`, `"plain"` | ✓ pass (literal at runtime) |                                  
  | `Expression` | `$["Build"].x + 1` | ✓ pass (valid bare expression) |
                                                                                                                                
  Trade-off: `if expression: "undefinedThing"` (which the `if` component *does* compile at runtime) won't be caught statically. 
  Syntax errors and unknown node references in the same field still are.                                                        
                                                                                                                                
  Property paths inside payloads (`$['X'].data.foo`) and runtime-only graph reachability are out of scope — those need real     
  payloads/edges.
                                                                                                                                
                                                                                                       
                                                                  
  Wired in:                                                                                                                     
  - `pkg/grpc/actions/canvases/serialization.go` — hook in `ParseCanvas` after `nodesByID` is built; errors fold into
  `nodeValidationErrors`.                                                                                                       
  - `pkg/configuration/validation.go` — `ExpressionPlaceholderRegex` exported so the new package and `validateObject` share one
  regex.                                                                                                                        
  - `pkg/workers/contexts/node_configuration_builder.go` — uses the moved `ParseReferencedNodes`.
                                                                                                                               
  
<img width="1118" height="634" alt="image" src="https://github.com/user-attachments/assets/c64af749-fe96-4193-994a-c0a9ce027d82" />

  
<img width="1332" height="700" alt="image" src="https://github.com/user-attachments/assets/f349ff03-ce15-4200-bdc3-50b1aa85fcf6" />
<img width="1322" height="660" alt="image" src="https://github.com/user-attachments/assets/f12594fd-406f-48dd-9375-f26250cb9b61" />

### DIFF
--- a/.go-coverage-baseline.json
+++ b/.go-coverage-baseline.json
@@ -20,7 +20,7 @@
     "pkg/cli/core": 46.1,
     "pkg/config": 0,
     "pkg/configuration": 26.2,
-    "pkg/configuration/expressionvalidation": 70,
+    "pkg/configuration/expressionvalidation": 71.5,
     "pkg/crypto": 53.7,
     "pkg/database": 0,
     "pkg/exprruntime": 36.4,
@@ -29,7 +29,7 @@
     "pkg/grpc/actions/agents": 44,
     "pkg/grpc/actions/auth": 80.8,
     "pkg/grpc/actions/blueprints": 0,
-    "pkg/grpc/actions/canvases": 58,
+    "pkg/grpc/actions/canvases": 61.5,
     "pkg/grpc/actions/canvases/changesets": 84.4,
     "pkg/grpc/actions/canvases/layout": 63.4,
     "pkg/grpc/actions/integrations": 0,
@@ -62,7 +62,7 @@
     "pkg/web": 82.1,
     "pkg/widgets/annotation": 0,
     "pkg/workers": 39.5,
-    "pkg/workers/contexts": 42.5,
+    "pkg/workers/contexts": 41.5,
     "pkg/workers/eventdistributer": 0
   },
   "ignoredPackagePrefixes": [
--- a/.go-coverage-baseline.json
+++ b/.go-coverage-baseline.json
@@ -20,6 +20,7 @@
     "pkg/cli/core": 46.1,
     "pkg/config": 0,
     "pkg/configuration": 26.2,
+    "pkg/configuration/expressionvalidation": 70,
     "pkg/crypto": 53.7,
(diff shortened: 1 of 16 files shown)


forestileao:
👍

PR #4447, opened 2026-04-28
feat: preview template payload on hover in Manual Run trigger

### LINKED ISSUE #2810: Manual Run: hover to preview payload
**Problem:** On the Manual Run (Start) node, each configured template/payload is listed with a "Run" button. There is no way to preview the payload content before running. It would be helpful to see a preview on hover, similar to how the Send Message (Slack) component lets you preview the message.

**Current vs proposed:**
- **Current:** No preview for configured payloads; user must click Run (and possibly open the modal) to see what will be sent.
- **Proposed:** Hovering over a template/payload row (or a dedicated preview affordance) shows a tooltip or popover with a preview of the payload (e.g. formatted JSON or key fields), consistent with Slack message preview elsewhere.

**Use case:** Users can quickly verify which payload they're about to trigger without opening the modal or running the event. Especially useful when multiple templates have similar names or when debugging.

___
**Screenshot:**

Image

coffeenjoyer17:
Closes #2810

## Summary

Manual Run templates currently surface only their name and a Run button. When several templates share similar names (or when a user is just verifying before triggering), there is no way to confirm the payload short of opening the configuration. This PR adds a hover preview on the template row that shows the configured payload as formatted JSON.

## Implementation

- Wraps the icon + name part of each row in `PayloadTooltip` (the same component already used to surface component spec previews in `ui/componentBase/index.tsx`).
- The Run button stays outside the tooltip, so hovering Run does not open the preview.
- Adds a `cursor-help` class on the wrapped area so the affordance is discoverable.
- `data-testid="start-template-row"` added on the wrapped row for future tests.

## Notes

- No new dependency: `PayloadTooltip` is the existing pattern (Tippy + Monaco).
- 200 ms tooltip delay (Tippy default in `PayloadTooltip`) prevents accidental triggers from neighbouring hovers.
- `tsc --noEmit` clean. `npm run lint:budget` totals are unchanged from `main` (the pre-existing `react-refresh/only-export-components` and `max-lines` budgets are over before this change too).
- DCO sign-off included.

### DIFF
--- a/web_src/src/pages/workflowv2/mappers/start.tsx
+++ b/web_src/src/pages/workflowv2/mappers/start.tsx
@@ -12,6 +12,7 @@ import { renderTimeAgo } from "@/components/TimeAgo";
 import React from "react";
 import { Button } from "@/components/ui/button";
 import { Play } from "lucide-react";
+import { PayloadTooltip } from "@/ui/componentBase/PayloadTooltip";
 
 interface StartTemplate {
   name: string;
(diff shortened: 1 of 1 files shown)


shiroyasha:

After taking another look, this isn’t something we plan to move forward with right now, so I’ll close the PR for now.

PR #4448, opened 2026-04-28
fix: persist Building Blocks sidebar filter toggles

### LINKED ISSUE #4370: `Connected Integrations on Top` filter does not persist
When a component is selected from the sidebar, the filter resets.

Image

coffeenjoyer17:
Closes #4370

## Summary

The `Connected Integrations on Top` and `Show integration setup status` toggles in the Building Blocks sidebar reset whenever the sidebar remounts — selecting a component, opening the sidebar again, or navigating away and back. The issue references the connected-on-top filter; the setup-status toggle sits in the same dropdown and exhibits the same problem, so this PR persists both for consistency.

## Implementation

Both toggles are now hydrated from `localStorage` on mount and written back on every change, matching the pattern already used for `sidebarWidth` in the same component:

- `componentSidebarShowSetupStatus` — default `true`
- `componentSidebarConnectedOnTop` — default `false`

Key constants are exported from `ui/CanvasPage/index.tsx` next to the existing `COMPONENT_SIDEBAR_WIDTH_STORAGE_KEY`. SSR-safe (`typeof window === "undefined"` guard) like the existing width init.

## Verification

- `tsc --noEmit -p .` clean.
- Existing `filter.spec.ts` and `CategorySection.spec.tsx` still pass (15 tests).
- `npm run lint:budget` totals are unchanged from `main`.
- DCO sign-off included.

### DIFF
--- a/web_src/src/ui/BuildingBlocksSidebar/index.tsx
+++ b/web_src/src/ui/BuildingBlocksSidebar/index.tsx
@@ -7,7 +7,11 @@ import { getBackgroundColorClass } from "@/lib/colors";
 import { DropdownMenu, DropdownMenuCheckboxItem, DropdownMenuContent, DropdownMenuTrigger } from "@/ui/dropdownMenu";
 import { Search, Settings2, X } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


markoa:
Thank you for the submission, but we will not be proceeding. Best of luck!

coffeenjoyer17:
All good. If you've got a minute, any pointer on what'd have worked better would help. No worries if not.

PR #4451, opened 2026-04-28
fix: fail errored workflow nodes during event routing

Superstar-IT:
**Summary**

- Fixes a silent-stop behavior where EventRouter skipped target nodes in state=error and dropped execution flow without trace entries.
- When routing reaches an errored node, the router now creates a finished execution with result=failed, result_reason=error, and result_message from node StateReason (fallback if missing).
- Emits failure events on both failed and failure channels for errored nodes, so downstream error handlers can be triggered consistently.
- Applies this behavior across root, standard, and child (blueprint) execution routing paths.
- Adds coverage in event_router_test to verify failed execution creation, failure event emission, and downstream routing via failed edge.

**Why**
Previously, invalid nodes could swallow incoming executions: upstream looked successful, downstream never ran, and no failed execution was visible in trace. This change makes failures explicit, observable, and routable.

**Test Plan**

- [ ] Added unit/integration test: Test__EventRouter_ProcessExecutionEvent_TargetNodeErrorCreatesFailedExecution
- [ ] Verified assertions for:
- failed execution created for errored node
- resultMessage matches node state reason
- both failed and failure events created
- downstream node receives queue item from failed channel
- [ ]  Run go test ./pkg/workers -run EventRouter in CI/local Go environment
- [ ]  Run repository Go formatting/lint/build checks in standard dev container setup

### DIFF
--- a/pkg/workers/event_router.go
+++ b/pkg/workers/event_router.go
@@ -2,11 +2,13 @@ package workers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/protobuf/proto"
+	"gorm.io/datatypes"
 	"gorm.io/gorm"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


markoa:
Thank you for the submission, but we will not be proceeding. Best of luck!

Issue #4442, opened 2026-04-28
CPU spikes when nodes are moved on the canvas


superleomori:
## Description

CPU usage spikes significantly when nodes are moved on the canvas.

## Steps to Reproduce

1. Open a canvas with multiple nodes
2. Drag and move one or more nodes on the canvas
3. Observe CPU usage during and after the move

## Expected Behavior

Node movement should be smooth with minimal CPU impact.

## Actual Behavior

CPU spikes noticeably when nodes are being moved on the canvas.

## Environment

- Surface: Canvas UI

## Notes

This may be related to layout recalculation, re-rendering, or event listeners triggering on every drag event. Worth investigating whether debouncing or virtualization could help.

shiroyasha:
Assigning you @forestileao. Looked like a serious issue that we should address soon-ish. Feel free to re-assign to someone else.

forestileao:
I can handle it

forestileao:
Closing it, the major issue was fixed on https://github.com/superplanehq/superplane/pull/4496

PR #4453, opened 2026-04-28
fix(canvas): reduce CPU usage when dragging workflow nodes

superdev72:
## The problem

Dragging nodes on the workflow canvas caused high CPU usage. Each drag frame rebuilt overlay data on every node and re-ran the edge sync effect on every `nodes` update, forcing broad React updates and edge remapping.

Related: https://github.com/superplanehq/superplane/issues/4442

## The solution

- Added `CanvasNodeOverlayProvider` / `useCanvasNodeOverlay()` so hover, in-flight connection, edge list for handle logic, execution-chain highlights, and the existing callbacks ref are supplied without cloning `node.data` for every node on every move.
- Pass `state.nodes` directly into React Flow so unchanged nodes keep stable `data` where possible, allowing `memo` on node UI to skip work during drag.
- In `useCanvasState`, derive a `localEdgesSyncKey` from template/pending nodes only and run the edge re-merge effect on `[initialEdges, localEdgesSyncKey]`, using a `nodesRef` snapshot inside the effect so drag-only position updates do not trigger `setEdges` every frame.
- `Block` and handles read overlay from context, with the previous `data._…` fields as fallbacks (e.g. unit tests without a provider).

### Additional implementation notes

- `CanvasNodeRendererCallbacks` lives in `canvasNodeOverlayContext.tsx` to keep types and context together.
- `BlockConnectionState.handleType` allows `null` to match runtime connect state.
- No API, proto, or OpenAPI changes—UI-only performance work.

### Testing notes

- [ ] Drag nodes on a canvas with many nodes; confirm smoother interaction and lower CPU (browser Performance tab / system monitor).
- [ ] Regression: edge hover, new connections, template/pending nodes, execution-chain highlighting, multi-select, read-only canvas.
- [ ] `make format.js` and `make check.build.ui` (or your CI equivalents).

---

Have you done the following:
- [ ] Have you proofread your commits?
- [ ] Did you update the docs, API spec, and README as necessary?

### DIFF
--- a/web_src/src/ui/CanvasPage/Block/handles.tsx
+++ b/web_src/src/ui/CanvasPage/Block/handles.tsx
@@ -1,5 +1,7 @@
 import React from "react";
 import { Handle, Position } from "@xyflow/react";
+
+import { useCanvasNodeOverlay } from "../canvasNodeOverlayContext";
 import { getOutputChannels } from "./data";
 import type { BlockConnectionState, BlockEdgeState, BlockProps, CanvasBlockData } from "./types";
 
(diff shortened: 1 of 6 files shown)


forestileao:
Closing it, the issue was fixed https://github.com/superplanehq/superplane/pull/4496

PR #4457, opened 2026-04-28
feat: enforce Postgres statement and idle‑in‑transaction timeouts on pooled connections

### LINKED ISSUE #4376: Add timeout for database transactions with logging
## Summary

Currently, database transactions in the SuperPlane backend do not have enforced timeouts. Long-running or stuck transactions can degrade performance and reliability of the control plane. We need a two-layer timeout strategy: a catch-all at the database level and more granular, context-driven timeouts at the application level, with appropriate logging.

## Requested work

1. **Database-level catch-all timeout**
   - Configure a global transaction timeout on the database connection:
     ```sql
     SET transaction_timeout = '5min';
     ```
   - This should be applied during database initialization/connection setup so every session inherits the default.

2. **Application-level context-based timeouts**
   - Wrap GORM transactions with a context that carries a shorter, operation-specific timeout:
     ```go
     db.WithContext(ctx).Transaction(func(tx *gorm.DB) error { ... })
     ```
   - Audit existing transaction call sites in `cmd/`, `pkg/`, and `test/` and ensure they pass a context with an appropriate deadline.
   - Define sensible default timeout values per operation category (e.g., canvas mutations, event processing, read-heavy queries).

3. **Logging**
   - Log a warning when a transaction is cancelled or times out, including the operation name and elapsed duration.
   - Ensure errors propagated from timed-out transactions are clearly distinguishable in logs from other DB errors.

## SuperPlane context

SuperPlane workflows (Canvases) involve multiple database writes when creating or updating triggers, components, and edges. A stuck transaction during canvas evaluation or event processing could block the entire pipeline. Adding timeouts with clear logging improves the reliability of the control plane and makes debugging production issues significantly easier.

## Notes

- The 5-minute database-level timeout is a safety net; application-level timeouts should generally be much shorter (seconds to low minutes) depending on the operation.
- Verify that the target database engine (PostgreSQL) supports `transaction_timeout` (available in PostgreSQL 17+). If the deployment target uses an older version, consider using `statement_timeout` or `idle_in_transaction_session_timeout` as alternatives and document the trade-offs.
- Existing tests in `test/` should be reviewed to ensure they still pass with the new timeout constraints.

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/4376

We were seeing the control plane stall with no helpful logs when DB work held connections too long—small `DB_POOL_SIZE` per pod made that 
  painful.
  This PR sets Postgres session limits on every app connection:
  - **statement_timeout** aborts queries that run too long (default **60s**).
  - **idle_in_transaction_session_timeout** ends sessions stuck idle inside a transaction (default **30s**), so abandoned transactions release 
  locks and pooled connections sooner.
  Limits come from `DB_STATEMENT_TIMEOUT_MS` and `DB_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_MS` (defaults + Helm + docker-compose.dev.yml). 
  `pkg/database` wires them into the URI; `pkg/config` only exposes `GetenvOrDefault` for reading those vars.
  On failures that look like these timeouts we log explicitly via the GORM logger wrapper. `connection_test.go` includes an integration check 
  (under Docker / `make test`) that proves `statement_timeout` actually fires.
  **Note:** This does not fix every “waiting for a free connection from the pool” stall; we can add context deadlines later if we want that 
  too.


pkg/database/config.go
```diff
@@ -0,0 +1,43 @@
+package database
+
+import (
+	"os"
+)
+
+const (
+	DefaultDBStatementTimeoutMS                = "60000"
+	DefaultDBIdleInTransactionSessionTimeoutMS = "30000"
+)
```
forestileao: @shiroyasha @lucaspin What do you think about this timeout defaults values?

**DefaultDBStatementTimeoutMS:** Time spent executing each single statement—including SELECT, INSERT, UPDATE, DELETE, and many DDL operations

**DefaultDBIdleInTransactionSessionTimeoutMS:** Time the session spends inside an open transaction (BEGIN …) without doing anything: no SQL, no COMMIT/ROLLBACK.

shiroyasha: They look ok-ish, should work for 99%.

Do we get an exception in sentry when it takes longer? 
If yes, we can adjust based if it has edge cases.

forestileao: @shiroyasha 
Good point. Let me confirm that for sentry

forestileao: added hub.CaptureException to this feature, before it was just using hub.Recover on panics and Generic HTTP 5xx via captureHTTPError → hub.CaptureMessage

shiroyasha (dismissed):

shiroyasha:
From my side 👍 
But it is a drastic enough change that I would like @lucaspin to review it as well.

pkg/database/config.go
```diff
@@ -0,0 +1,43 @@
...
+
+type Config struct {
+	Database Database
+}
+
+type Database struct {
+	StatementTimeoutMS                string
```
lucaspin: We should use time.Duration for these, especially because these come from app configuration. A lot cleaner to configure it with:

```
DB_STATEMENT_TIMEOUT=60s
```

Than with:

```
DB_STATEMENT_TIMEOUT_MS=60000
```

pkg/database/config.go
```diff
@@ -0,0 +1,43 @@
...
+	IdleInTransactionSessionTimeoutMS string
+}
+
+func Load() Config {
```
lucaspin: LoadConfig()? Load() is weird because this is publicly exposed method, so other packages can call `database.Load()` and that doesn't seem like config-related stuff. Also, why do you need `Config` and `Database` objects? Why not a single Config one?

lucaspin:
Just one additional note: if setting this on the connections like this is not reliable enough, we could also look into setting on the database level. See: https://docs.cloud.google.com/sql/docs/postgres/flags

forestileao:
@lucaspin Comments addressed. Related to the flags, we can consider adding them as well. Just wanted to add in the code primarily so we can see it explicitly

lucaspin (approved):

PR #4459, opened 2026-04-28
fix: CLI surface node configuration errors after canvas update

### LINKED ISSUE #4439: CLI should validate component field types on canvas update
## Problem

When updating a canvas via CLI, field type mismatches are silently accepted. The canvas update succeeds, but the affected node gets an `errorMessage` that prevents it from executing. The only way to discover this is by manually inspecting the node via API.

## Examples found during ephemeral env canvas work

| Component | Field | Expected | Accepted (broken) |
|---|---|---|---|
| `wait` | `waitFor` | string (`"30"`) | number (`30`) |
| `http` | `timeoutSeconds` | number (`10`) | string (`"10"`) |
| `http` | `retry.enabled` | boolean (`true`) | string (`"true"`) |

## Impact

- Canvas update succeeds with exit 0 — no indication of a problem
- Node silently stops accepting incoming events
- Execution chain breaks with no trace in the event/execution API
- Only discoverable by checking `errorMessage` on the node via raw API
- Caused 30+ minutes of debugging

## Expected behavior

**Option A (preferred):** CLI validates field types against the component schema during `canvases update` and rejects with a clear error.

**Option B:** API rejects the update if any node has invalid field types, returning the validation errors in the response.

**Option C (minimum):** Executions that hit a node with `errorMessage` should show a FAILED result with the error message, not silently drop.

forestileao:
Closes: https://github.com/superplanehq/superplane/issues/4439

## Description
  This fixes the case where `superplane canvases update` could exit successfully while nodes had invalid configuration (wrong JSON types, 
  etc.). The API already stores validation output on each node as `error_message` (and softer issues as `warning_message`), but the CLI did not
   surface those fields, so broken canvases were easy to miss.
  After a successful update response, the CLI walks `version.spec.nodes` and summarizes `error_message` / `warning_message`. No new fields were
   added to the REST or protobuf responses.
  ### Behavior
  - **Errors (`error_message`)**: command fails with a bullet list (node id, display name, API message). The draft may already be saved; 
  **auto-publish is skipped** when this fires.
  - **Warnings (`warning_message`) only**: command succeeds; in **text** output, an extra “Node warnings” section is printed after the normal 
  summary.
  ### Examples (text UI)
  **1) Configuration errors — command fails (exit ≠ 0)**
  ```text
  canvas was saved but the following nodes have configuration errors (error_message on each node):
    - node wait-1 (Wait): field 'waitFor': must be a string
    - node http-1 (HTTP Request): field 'timeoutSeconds': must be a number
  ```
  **2) No errors, only warnings — command succeeds**
  ```text
  Canvas version updated: 8f3b2c1d-1111-2222-3333-444455556666
  Canvas ID: a1b2c3d4-aaaa-bbbb-cccc-ddddeeeeffff
  Nodes: 4
  Edges: 3
  Integrations: 2
  Node warnings (warning_message):
    - node if-1 (If): <warning text from API>
  ```
  **3) Clean update — command succeeds**
  ```text
  Canvas version updated: 8f3b2c1d-1111-2222-3333-444455556666
  Canvas ID: a1b2c3d4-aaaa-bbbb-cccc-ddddeeeeffff
  Nodes: 4
  Edges: 3
  Integrations: 2
  ```

### DIFF
--- a/pkg/cli/commands/canvases/update.go
+++ b/pkg/cli/commands/canvases/update.go
@@ -95,6 +95,9 @@ func (c *updateCommand) Execute(ctx core.CommandContext) error {
 	}
 
 	version := response.GetVersion()
+	if errText := formatNodeSpecErrorsForCLI(version); errText != "" {
+		return fmt.Errorf("%s", errText)
+	}
 
 	// When not in draft mode, auto-publish the updated draft version.
 	if !draftMode {
@@ -129,6 +132,12 @@ func (c *updateCommand) Execute(ctx core.CommandContext) error {
 			}
 		}
 		_, err := fmt.Fprintf(stdout, "Integrations: %d\n", len(integrations))
+		if err != nil {
+			return err
+		}
+		if warnText := formatNodeSpecWarningsForCLI(version); warnText != "" {
+			_, err = fmt.Fprint(stdout, warnText)
+		}
 		return err
 	})
 }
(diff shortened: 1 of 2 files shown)


forestileao:
👍

PR #4460, opened 2026-04-28
fix: Don't close sidebar on component drop

### LINKED ISSUE #4372: Adding multiple components requires closing and reopening the component menu
When creating a canvas with multiple components I can only add one component at a time from the menu. Once placed and configured I have to close the configuration menu and go back to `+` to add another component. This does not feel intuitive having to close and reopen when handling multiple components. A back button is a proposed solution for this.

https://github.com/user-attachments/assets/86b98e0b-7e63-4b8b-b972-83b0154f155b

mchalapuk:
### What changed

This PR removes the UI logic that automatically closes the "Add Component" sidebar and opens component details after a new component is dropped on the canvas.
 
### Why

Better UX. This change eliminates the need to click "+" multiple times when manually adding multiple components to the canvas. The downside is that the user must click the dropped component to open its details in the sidebar. It's better since the mouse pointer is already there after the drop.

### Related issues

Fixes #4372


web_src/src/ui/CanvasPage/index.tsx
```diff
@@ -993,23 +993,16 @@ function CanvasPage(props: CanvasPageProps) {
 
       // Save immediately with defaults
       if (props.onNodeAdd) {
-        const newNodeId = await props.onNodeAdd({
+        void await props.onNodeAdd({
```
shiroyasha: Why do we need this `void` here?

mchalapuk: It means "throw away the returned value". It's not required here but it documents that there's a returned value and it is discarded.

andrecalil (approved):

PR #4469, opened 2026-04-28
feat: allow secrets create/update to read from stdin with -f

### LINKED ISSUE #4232: CLI secrets create: agent-friendly input
**Problem:** `superplane secrets create` requires `-f` with a **filesystem path** only (`os.ReadFile` in `pkg/cli/commands/secrets/common.go`). Coding agents and scripts that cannot rely on a writable temp file (or that want to pipe generated YAML) have no supported way to provision secrets non-interactively.

**Current vs proposed:** Current: file path only. Proposed: at minimum support **`-f -` / stdin** (and optionally `--from-literal` or similar) so agents can `echo … | superplane secrets create -f -` without touching disk.

**Use case:** Automated onboarding and agent runs that create org secrets from in-memory YAML without extra file lifecycle.

0vertake:
## Summary

  - Extended `secrets create` and `secrets update` to support reading YAML from stdin when `-f -` is passed
  - Updated flag description to document the `-` sentinel
  - Added success-path and stdin-specific test cases to `common_test.go`

  This unblocks non-interactive secret provisioning without any temporary file lifecycle management.

  Closes #4232

### DIFF
--- a/pkg/cli/commands/secrets/common.go
+++ b/pkg/cli/commands/secrets/common.go
@@ -23,11 +23,20 @@ type secretResource struct {
 	Spec       *openapi_client.SecretsSecretSpec     `json:"spec,omitempty"`
 }
 
-func parseSecretFile(path string) (*secretResource, error) {
-	// #nosec
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read resource file: %w", err)
+func parseSecretInput(path string, stdin io.Reader) (*secretResource, error) {
+	var data []byte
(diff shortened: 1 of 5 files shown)


shiroyasha:
Looks good. 👍 @0vertake please log an issue if there are other commands that are reading in files and should be able to use this option.

0vertake:
> Looks good. 👍 @0vertake please log an issue if there are other commands that are reading in files and should be able to use this option.

@shiroyasha Created #4526 based on this. Feel free to give feedback there if something is incorrect

PR #4479, opened 2026-04-29
fix: pass resolved execution config to workflow-level component hooks

### LINKED ISSUE #4441: HTTP component: expressions in URL field not resolved at execution time
## Problem

The HTTP component's `url` field accepts expressions like `{{ $['Node Name'].data.field }}` and the `configuration` object in the execution shows the resolved value, but the actual HTTP request uses the raw template string, causing a URL parse error.

## Steps to reproduce

1. Create an HTTP component with URL containing a node reference expression
2. Trigger the canvas so the upstream node provides the data
3. Check the execution metadata

## Observed

The `configuration.url` field shows the resolved IP (e.g., `http://45.55.209.135.sslip.io/health`), but the retry `lastError` shows the raw template string was used for the actual request:

```
"lastError": "failed to create request: parse \"http://{{ $['Create Droplet'].data.networks.v4[0].ip_address }}.sslip.io/health\": invalid port \".ip_address }}.sslip.io\" after host"
```

## Environment

Self-hosted SuperPlane instance (proto branch on DO droplet, pre-v0.18). May be version-specific — needs verification on latest release.

## Workaround

Used SSH component instead, which resolves expressions in the `host` field correctly.

forestileao:
## Problem
  Internal component hooks triggered via `NodeRequestWorker` for **workflow-level** nodes (no blueprint parent) received `node.Configuration` (raw canvas templates with `{{ … }}`) instead of `execution.Configuration` (snapshot resolved when the execution was created). That broke retries (e.g. HTTP `retryRequest`): metadata could show a resolved URL while the hook still saw the template string and failed URL parsing.
  ## Change
  `invokeParentNodeComponentAction` now passes `execution.Configuration.Data()`, aligned with `invokeChildNodeComponentAction` and normal `Execute`.
  ## Test
  **Before fix:** `Test_NodeRequestWorker_InternalHookUsesExecutionSnapshotConfiguration` fails — hook observes the raw URL string.
  **After fix:** Same test passes — hook observes the resolved URL from the execution snapshot.
  Probe: minimal registered action `hook_probe_cfg_action` / hook `probeHook`; canvas stores raw URL on the node and resolved URL on the execution; asserts the hook receives the resolved value.
  
  Closes https://github.com/superplanehq/superplane/issues/4441

### DIFF
--- a/pkg/workers/node_request_worker.go
+++ b/pkg/workers/node_request_worker.go
@@ -301,7 +301,7 @@ func (w *NodeRequestWorker) invokeParentNodeComponentAction(
 	logger := logging.ForExecution(execution, nil)
 	hookCtx := core.ActionHookContext{
 		Name:           spec.InvokeAction.ActionName,
-		Configuration:  node.Configuration.Data(),
+		Configuration:  execution.Configuration.Data(),
 		Parameters:     spec.InvokeAction.Parameters,
 		HTTP:           w.registry.HTTPContext(),
 		Metadata:       contexts.NewExecutionMetadataContext(tx, execution),
--- a/pkg/workers/node_request_worker_hook_probe_test.go
+++ b/pkg/workers/node_request_worker_hook_probe_test.go
@@ -1,0 +1,182 @@
+package workers
+
+import (
+	"net/http"
+	"sync"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


forestileao:
👍

PR #4485, opened 2026-04-29
fix: normalize GraphQL component icon color in sidebar

### LINKED ISSUE #4483: GraphQL component color is darker than the rest of the core components
Image

nyagamunene:
Fixes [issue](https://github.com/superplanehq/superplane/issues/4483)

- The GraphQL icon SVG used hardcoded #111827 (gray-900) for all strokes
and fills, making it visibly darker than other core components (Filter,
If, Merge, HTTP Request) which render Lucide icons at text-gray-500
(#6B7280). Updated the SVG to use #6B7280 to match.

- Added tests to CategorySection to cover the two rendering paths:
graphql blocks render via  (SVG icon), non-graphql blocks render
via a Lucide icon component.

### DIFF
--- a/web_src/src/assets/icons/graphql.svg
+++ b/web_src/src/assets/icons/graphql.svg
@@ -1,10 +1,10 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32" fill="none">
-  <path d="M16 3.5 27 10v12L16 28.5 5 22V10L16 3.5Z" stroke="#111827" stroke-width="2.25" stroke-linejoin="round"/>
-  <path d="M16 3.5v25M5 10l22 12M27 10 5 22" stroke="#111827" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
-  <circle cx="16" cy="3.5" r="2.5" fill="#111827"/>
-  <circle cx="27" cy="10" r="2.5" fill="#111827"/>
-  <circle cx="27" cy="22" r="2.5" fill="#111827"/>
-  <circle cx="16" cy="28.5" r="2.5" fill="#111827"/>
-  <circle cx="5" cy="22" r="2.5" fill="#111827"/>
-  <circle cx="5" cy="10" r="2.5" fill="#111827"/>
+  <path d="M16 3.5 27 10v12L16 28.5 5 22V10L16 3.5Z" stroke="#6B7280" stroke-width="2.25" stroke-linejoin="round"/>
+  <path d="M16 3.5v25M5 10l22 12M27 10 5 22" stroke="#6B7280" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+  <circle cx="16" cy="3.5" r="2.5" fill="#6B7280"/>
+  <circle cx="27" cy="10" r="2.5" fill="#6B7280"/>
+  <circle cx="27" cy="22" r="2.5" fill="#6B7280"/>
+  <circle cx="16" cy="28.5" r="2.5" fill="#6B7280"/>
+  <circle cx="5" cy="22" r="2.5" fill="#6B7280"/>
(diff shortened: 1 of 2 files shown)


abdusabri:
(note that I'm not one of the maintainers, just a passer-by 😁)

@nyagamunene  Just a note that the GraphQL icon is used in other places in the app as well - not just in the Add Component Sidebar. So, it is worth confirming with the core team/maintainers that it is fine for the lighter color to be applied to all other places as well (as this change is done to the core asset/icon that's used elsewhere as well).

Following are screenshots of such cases (without the current change).

image

--------------------------------------------

image copy

---------------------------------------------

image copy 2

nyagamunene:
> (note that I'm not one of the maintainers, just a passer-by 😁)
> 
> @nyagamunene Just a note that the GraphQL icon is used in other places in the app as well - not just in the Add Component Sidebar. So, it is worth confirming with the core team/maintainers that it is fine for the lighter color to be applied to all other places as well (as this change is done to the core asset/icon that's used elsewhere as well).
> 
> Following are screenshots of such cases (without the current change).
> 
> image
> image copy
> image copy 2

Thank you @abdusabri for pointing that out.

I've traced all the places the SVG is consumed:

- Add Component sidebar: the list item icon (the original motivation for the fix)
- Canvas node header: when a GraphQL node sits on the workflow canvas, via graphqlMapper
- Category group header in the sidebar: the small icon next to the "GraphQL" integration category label
- Template/canvas creation page: where integration icons are displayed during workflow setup

In all of these cases the icon renders on a white or light-grey background, so #6B7280 (gray-500) remains clearly visible and distinctive, it hasn't become invisible or washed out. The change makes it consistent with the other core components (Filter, If, Merge, HTTP Request), which all render Lucide icons at that same shade.

You're right that this was a global change to the core asset rather than a targeted sidebar-only fix. If the maintainers prefer the darker icon on the canvas node or elsewhere, an alternative would be to keep the SVG as-is and instead apply a CSS filter only in the sidebar  element. Happy to adjust whichever way the team prefers.

abdusabri:
> > (note that I'm not one of the maintainers, just a passer-by 😁)
> > 
> > @nyagamunene Just a note that the GraphQL icon is used in other places in the app as well - not just in the Add Component Sidebar. So, it is worth confirming with the core team/maintainers that it is fine for the lighter color to be applied to all other places as well (as this change is done to the core asset/icon that's used elsewhere as well).
> > 
> > Following are screenshots of such cases (without the current change).
> > 
> > image
> > image copy
> > image copy 2
> 
> Thank you @abdusabri for pointing that out.
> 
> I've traced all the places the SVG is consumed:
> 
> - Add Component sidebar: the list item icon (the original motivation for the fix)
> - Canvas node header: when a GraphQL node sits on the workflow canvas, via graphqlMapper
> - Category group header in the sidebar: the small icon next to the "GraphQL" integration category label
> - Template/canvas creation page: where integration icons are displayed during workflow setup
> 
> In all of these cases the icon renders on a white or light-grey background, so #6B7280 (gray-500) remains clearly visible and distinctive, it hasn't become invisible or washed out. The change makes it consistent with the other core components (Filter, If, Merge, HTTP Request), which all render Lucide icons at that same shade.
> 
> You're right that this was a global change to the core asset rather than a targeted sidebar-only fix. If the maintainers prefer the darker icon on the canvas node or elsewhere, an alternative would be to keep the SVG as-is and instead apply a CSS filter only in the sidebar  element. Happy to adjust whichever way the team prefers.

You are welcome! 🙂 

Cool! Sounds good! 

Hope you have the PR merged soon!

markoa:
Thank you for applying, but we will not be proceeding with your application.

nyagamunene:
> Thank you for applying, but we will not be proceeding with your application.

Thanks for the reply. If its not much to ask, could a clarification on the issue be provided?

PR #4486, opened 2026-04-29
chore: use impl.DummyAction for hook configuration test

forestileao:
## What
  Replace the bespoke `testHookProbeAction` type in 
  `node_request_worker_hook_probe_test.go` with 
  `impl.NewDummyAction` from `test/support/impl`.
  ## Why
  Keeps the regression test for workflow-level internal hooks 
  (`#4441`) the same behavior-wise while leaning on shared test 
  doubles instead of reimplementing the full `core.Action` surface.
  ## Notes
  Hook wiring is unchanged: `probeHook` + `HandleHookFunc` capture 
  `ctx.Configuration`; `ProcessQueueFunc` / `ExecuteFunc` match 
  what the dummy previously did.

### DIFF
--- a/pkg/workers/node_request_worker_hook_probe_test.go
+++ b/pkg/workers/node_request_worker_hook_probe_test.go
@@ -1,7 +1,6 @@
 package workers
 
 import (
-	"net/http"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


forestileao:
👍

PR #4488, opened 2026-04-29
fix: handle stop expression evaluation errors in Merge component

### LINKED ISSUE #4482: Merge: conditional stop expression crashes on nil field access, silently blocks queue
## Problem

When a Merge node receives inputs from different component types (e.g., Dash0 synthetic checks + Dash0 Prometheus query), the conditional stop expression can crash on nil field access, silently blocking all queued items.

## Steps to Reproduce

1. Create a Merge with inputs from `dash0.getHttpSyntheticCheck` (outputs `{data: {metrics: {lastOutcome: "Healthy"}, ...}}`) and `dash0.queryPrometheus` (outputs `{data: {data: {result: [...], ...}, status: "success"}}`)
2. Set conditional stop expression:
   ```
   previous().data.metrics.lastOutcome == "Critical" || previous().data.data.result[0].value[1] != "0"
   ```
3. When a synthetic check input arrives, `previous().data.data.result[0].value[1]` accesses nil fields and crashes
4. All queued items in the merge stop processing — no error message, no timeout, just stuck

## Expected Behavior

- Expression evaluation should handle nil/missing fields gracefully (return false, not crash)
- OR: short-circuit evaluation — if the first `||` condition is decidable, skip evaluating the rest
- At minimum: if expression evaluation fails, log an error on the execution and emit on the `fail` channel instead of silently blocking the queue

## Observed Behavior

- Merge queue accumulates items but never processes them
- No `errorMessage` on the node
- No execution appears in the trace
- Only way to recover is manually canceling queued items from the UI

## Workaround

Only use conditional stop expressions that reference fields guaranteed to exist on ALL input types. Cannot mix inputs from different component types in a single merge with conditional stop.

## Impact

This was hit while building a blue/green deployment canvas with parallel health checks (synthetic + PromQL) feeding into a single merge. Had to remove the PromQL check to work around it.

gaortiz1:
## What                                                   

  When a Merge node's conditional stop expression fails to evaluate against an incoming event (e.g. nil field access when two upstream sources emit different shapes), the merge now finishes the execution on the `fail` channel with a clear error message instead of silently blocking the queue.                        
   
  Fixes #4482.                                                                                                                                                                                                                                                                                                              
                                                            
  ## Why                                                                                                                                                                                                                                                                                                                    
                                                            
  Per the issue: when the stop expression hits a missing field, expression evaluation returns an error. Previously, `Merge.ProcessQueueItem` propagated that error up; the queue worker wraps `processNode` in a database transaction, so the propagated error rolled back the dequeue along with everything else. The same 
  item was retried indefinitely, no execution was recorded, and no error surfaced to the user. The only recovery was manually canceling queue items from the UI.
                                                                                                                                                                                                                                                                                                                            
  ## How                                                                                                                                                                                                                                                                                                                    
   
  In `pkg/components/merge/merge.go`, when `ctx.Expressions.Run` returns an error we now mirror the existing branch that handles non-boolean expression output: mark the merge metadata as `StopEarly`, call `ExecutionState.Fail("error", ...)` with the underlying evaluation error, and return `nil` as the error so the 
  surrounding transaction commits.                          
                                                                                                                                                                                                                                                                                                                            
  This treats expression evaluation failure as a valid `fail` channel outcome (the expression cannot reasonably be evaluated against this input), not as a transient error to retry.                                                                                                                                        
   
  While mirroring the new error branch, the existing non-boolean branch shared the exact same "mark stopEarly + persist metadata + fail execution" sequence. Extracted both into a local `failWithError` closure inside the stop-expression block to keep the three stop causes legible at a glance.                        
                                                            
  ### Alternatives considered                                                                                                                                                                                                                                                                                               
                                                            
  - **Make field access nil-safe globally** (option 1 in the issue): would require changing `ResolveExpression`, which is shared by every component. Out of scope and risks masking legitimate expression bugs elsewhere.                                                                                                   
  - **Rely on short-circuit evaluation** (option 2 in the issue): `expr` already short-circuits `||`. In the issue's example the LHS evaluates to `false`, so the RHS is still evaluated and crashes — short-circuit does not help here.
                                                                                                                                                                                                                                                                                                                            
  ## Test plan                                              
                                                                                                                                                                                                                                                                                                                            
  - [x] Added `Test_Merge_StopIfExpression_EvaluationError` in `pkg/components/merge/merge_test.go` reproducing the issue scenario (two sources with different shapes, expression accessing a missing path). Test fails on `main` and passes with this fix.                                                                 
  - [x] `make format.go`, `make lint`, `make check.build.app` clean.
  - [x] `make test PKG_TEST_PACKAGES=./pkg/components/merge` — all 6 tests green.

### DIFF
--- a/pkg/components/merge/merge.go
+++ b/pkg/components/merge/merge.go
@@ -292,29 +292,25 @@ func (m *Merge) ProcessQueueItem(ctx core.ProcessQueueContext) (*uuid.UUID, erro
 	// configs created before the toggle will still work if they have an expression.
 	//
 	if spec.StopIfExpression != "" {
-		out, err := ctx.Expressions.Run(spec.StopIfExpression)
-		if err != nil {
-			return nil, err
+		failWithError := func(message string) (*uuid.UUID, error) {
+			md.StopEarly = true
+			if err := executionCtx.Metadata.Set(md); err != nil {
+				return nil, err
+			}
+			if err := executionCtx.ExecutionState.Fail("error", message); err != nil {
+				return nil, err
+			}
+			return &executionCtx.ID, nil
 		}
 
-		//
-		// If the stop expression does not evaluate to a boolean value,
-		// we fail the execution and mark the merge as stopped early.
-		//
+		out, err := ctx.Expressions.Run(spec.StopIfExpression)
+		if err != nil {
+			return failWithError(fmt.Sprintf("stop expression evaluation failed: %v", err))
+		}
+
 		stopEarly, ok := out.(bool)
 		if !ok {
-			md.StopEarly = true
-			err := executionCtx.Metadata.Set(md)
-			if err != nil {
-				return nil, err
-			}
-
-			err = executionCtx.ExecutionState.Fail("error", fmt.Sprintf("stop expression must evaluate to boolean, got %T: %v", out, out))
-			if err != nil {
-				return nil, err
-			}
-
-			return &executionCtx.ID, nil
+			return failWithError(fmt.Sprintf("stop expression must evaluate to boolean, got %T: %v", out, out))
 		}
 
 		//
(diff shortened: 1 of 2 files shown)


forestileao:
Sorry, but the fix was deployed

PR #4489, opened 2026-04-29
docs: add security note to how it works section

mayerpeter502-cloud:


markoa:
Thank you for applying, but we will not be proceeding with your application.

PR #4490, opened 2026-04-29
feat: add merge discoverability pattern and fix output channel name

### LINKED ISSUE #4294: Discoverability: merge/fork-join patterns
### Context
A workflow needed two independent LLM calls in parallel, then convergence.

### Observed behavior
Agent defaulted to a sequential chain and initially assumed parallel fan-out/fan-in was unsupported, until a human explicitly pointed to `merge`.

Root cause: `superplane index components` lists `merge` among many components without pattern-level guidance, so it’s easy to miss as the answer to “run two branches in parallel and wait for both.”

### Suggested improvement
Add pattern-oriented guidance in docs/CLI, for example:
- "Fan-out / fan-in" recipe using edge fork + `merge`
- `superplane patterns` / `superplane recipes` command showing common multi-node flow patterns

### Why this matters
Improves solution discovery for agents and humans, reducing trial-and-error and sequential-bias designs.

collectivenectar:
## What

- Added `agent/patterns/await-parallel-upstream-nodes.md`: a pattern that teaches the agent to use the `merge` component when a user wants two or more nodes to run in parallel and converge before downstream work proceeds.
- Added `agent/evals/evaluators/canvas_edge_uses_channel.py` and a corresponding eval case (`parallel_http_ssh_merge`) to verify the agent produces the correct topology.
- Fixed the `merge` output channel name in `agent/skills/data-flow.md`: the table listed `stopped` but the real channel is `fail` (confirmed in `pkg/components/merge/merge.go:25`). Any workflow subscribing downstream of merge on `stopped` would silently never fire.

## Why

Issue #4294: the agent had no pattern to match against when asked to build parallel fan-in workflows, so it either skipped `merge` entirely or produced an incorrect topology. The channel name bug compounded this — even when the agent did reach for `merge`, any failure-path subscription would be broken.

## How

`agent/patterns/await-parallel-upstream-nodes.md` follows the format of the two existing patterns (`ephemeral-preview-machines.md`, `issue-auto-resolve-agent-label.md`): no frontmatter, a `Keywords:` line for search scoring, a decision checklist, and a canonical workflow section. The `->` diagram notation is extended to show fan-out from a single trigger and fan-in to `merge` — no new syntax invented. Notes cover the timeout and `stopIfExpression` edge cases without prescribing them.

The eval case asserts three things: `merge` appears exactly once, a path exists leading into it, and the downstream edge uses the `success` channel.

## Test plan

- `make test.agent.unit` — 178 passing
- `make test.agent.evals CASES=parallel_http_ssh_merge` — 4/4 assertions passing (agent reached for `merge`, produced correct fan-in topology, used `success` channel downstream)

Closes #4294

### DIFF
--- a/agent/evals/cases.py
+++ b/agent/evals/cases.py
@@ -108,6 +108,18 @@ dataset = Dataset(
             ),
         ),
         Case(
+            name="parallel_http_ssh_merge",
+            inputs=(
+                "I need a workflow that makes an API call and runs a server health check "
+                "at the same time, and only proceeds after both respond."
+            ),
+            evaluators=(
+                evals.CanvasHasNode("merge", count=1),
+                evals.CanvasHasWorkflow("...", "merge"),
+                evals.CanvasEdgeUsesChannel("success"),
+            ),
+        ),
+        Case(
             name="github_pr_close_long_open_filter",
             inputs=(
                 "When a GitHub pull request is closed, add a filter so the workflow only continues "
--- a/agent/evals/evaluators/__init__.py
+++ b/agent/evals/evaluators/__init__.py
@@ -1,4 +1,5 @@
 from .bracket_selectors_match_canvas_name import BracketSelectorsMatchCanvasNames
+from .canvas_edge_uses_channel import CanvasEdgeUsesChannel
 from .canvas_has_node import CanvasHasNode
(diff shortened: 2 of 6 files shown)


markoa:

PR #4491, opened 2026-04-29
Remove commands from AGENTS.md

fernandortec:
Agent harnesses are usually good at discovering and running scripts on their own. Listing them in agents.md often just wastes tokens and can increase hallucinations.

agents.md should be used to steer behavior when the agent makes recurring mistakes, not to map out the entire application structure or all scripts, unless those are consistently being misinterpreted.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,11 +17,6 @@
 
 ## Build, Test, and Development Commands
 
-- Setup dev environment: `make dev.setup`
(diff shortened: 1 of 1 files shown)


shiroyasha:
@fernandortec thank you for opening this PR! 

We noticed degraded agent understanding when we removed this, so I won't proceed with merging it.

PR #4494, opened 2026-04-29
feat: enforce OAuth allowed_providers for invites and expose in org spec

### LINKED ISSUE #1077: User can log in with provider not allowed by organization
The `allowed_providers` column for all organizations is ["github"] but user can log in with Google

0vertake:
## Summary

- `acceptInvitation` never enforced `allowed_providers` before accepting a pending email invitation after OAuth, so the wrong OAuth provider could still join the org. The fix threads the OAuth provider through `acceptPendingInvitations` → `acceptInvitation`, loads the org, and calls `IsProviderAllowed(provider)` before accepting.
- `IsProviderAllowed`: an empty `allowed_providers` list now means no restriction (allow any configured OAuth provider for that path). Previously an empty list behaved like “deny everyone.”
- Password / magic-code flows pass `provider == ""` and intentionally skip this OAuth-only gate (same as before for those methods).
- Org API + product: `Organization.spec` exposes `allowed_oauth_providers` (describe + patch). Org settings (General) lets admins restrict GitHub/Google for pending email invitations (switch + checkboxes, save, unsaved hint). CLI `organizations update --allowed-oauth-providers` and `organizations get output` document the setting.
- Defaults: new organizations get [github, google] as the initial allow-list (mentor alignment).
- Docs / scope: comments clarify this does not apply to shareable invite links today (separate product decision). 

https://github.com/user-attachments/assets/acf20804-bcc1-4281-8f89-8d55be3e4757

Closes #1077


shiroyasha:
Pinging you @forestileao as this is mostly your area of expertise.

forestileao:
@0vertake Maybe we should set the allowed providers to "github" and "google" instead of just "github".  Until we have a nice way to change that config on an orgs settings.

Also, a sanity check:
- What about people signing in using email directly:
image

- In self hosted we also have direct email/password login controlled by an ENV variable


My conclusion: We should have this control for allowed providers, but permitting all providers and login methods until we have a way to configure that at org level.

What do you think?

lucaspin:
I think we can include in the scope of this pull the ability to configure this in the [Organization.Spec](https://github.com/superplanehq/superplane/blob/main/protos/organizations.proto#L289) + UI changes as well

0vertake:
@forestileao @lucaspin Extended the PR based on the feedback, also changed the title and description to better reflect the new state

forestileao:
@0vertake Shouldn't we add control over direct email login on UI?

0vertake:
> @0vertake Shouldn't we add control over direct email login on UI?

I misread the comments, my bad. I'll add it now

shiroyasha:
@0vertake @forestileao what is the state of this PR?
The code was looking ok-ish.

0vertake:
> @0vertake @forestileao what is the state of this PR? The code was looking ok-ish.

I believe it's awaiting review after further features were added based on the feedback

forestileao:
@0vertake Sorry for letting the PR stale, going to review it tomorrow again

bender-rodriguez-unit1:

PR #4497, opened 2026-04-30
fix: preserve edges when canvas node IDs are regenerated

### LINKED ISSUE #4347: Canvas update silently renames trigger node IDs, breaking edges
## Problem

When updating a canvas via CLI, trigger node IDs are silently renamed by the server. For example:

- YAML specifies: `id: pr-opened`
- Server renames to: `id: github-onpullrequest-pr-opened-bf2320`

Edges referencing the original `pr-opened` ID become disconnected. Events fire on the trigger but no downstream nodes execute because the edge source ID no longer matches.

## Impact

- Silent breakage — no error message, just dead workflows
- Hard to debug — the canvas looks correct in YAML but edges are broken
- Requires exporting the canvas, finding the renamed ID, and updating edges manually

## Expected

Either:
1. Preserve user-specified node IDs, or
2. Return an error/warning when a node ID is renamed, or
3. Automatically update edge references when renaming

forestileao:
## Summary

Fixes canvas publishing so edges are updated when a node ID is regenerated due to an existing deleted node ID conflict.

Closes: https://github.com/superplanehq/superplane/issues/4347

### DIFF
--- a/pkg/grpc/actions/canvases/changesets/canvas_publisher.go
+++ b/pkg/grpc/actions/canvases/changesets/canvas_publisher.go
@@ -39,6 +39,7 @@ type CanvasPublisher struct {
 	draft      *models.CanvasVersion
 	changeset  *pb.CanvasChangeset
 	finalNodes map[string]models.Node
+	renamedIDs map[string]string
 
 	//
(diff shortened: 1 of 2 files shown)


forestileao:
👍

PR #4503, opened 2026-04-30
feat: add run hook to manual trigger

re-pixel:
### Summary

- Adds a `run` hook to the manual "start" trigger so it follows the same `InvokeNodeTriggerHook` pattern used by other triggers, instead of relying on the one-off `EmitNodeEvent` endpoint.
- Wires the `Events` context into the `InvokeNodeTriggerHook` gRPC handler so trigger hooks can emit events.
- Updates the frontend to call `InvokeNodeTriggerHook` when running a template from the manual trigger, while preserving the existing `EmitNodeEvent` path for re-emit and generic emit flows.

### Related issues

https://github.com/superplanehq/superplane/issues/4298


web_src/src/pages/workflowv2/mappers/types.ts
```diff
@@ -177,7 +177,15 @@ export interface EventStateRegistry {
...
-  onRun?: (initialData?: string) => void;
+  /**
+   * Open the run/emit-event modal for this node.
+   *
+   * @param initialData Optional JSON string to prefill the payload editor.
+   * @param templateName Optional name of the trigger template that originated the run.
+   *  When set, the modal will run that specific template via the trigger's `run` hook
+   *  instead of presenting a generic channel selector.
+   */
+  onRun?: (initialData?: string, templateName?: string) => void;
```
lucaspin: I don't like how this onRun() thing is working right now. I think the pattern we are using in the [approval mapper](https://github.com/superplanehq/superplane/blob/main/web_src/src/pages/workflowv2/mappers/approval.ts#L326) for calling the node execution hook is a lot cleaner. Is there any way to use something like that for this as well?

shiroyasha: @re-pixel ping on this one.

re-pixel: @lucaspin I made some changes and now the `onRun` works similarly to approval mapper you mentioned. Now the manual run works fine but I'm struggling to incorporate the `EmitEventModal` back into the flow. That's the modal that shows up when the user clicks **Run** and then can edit the template before actually executing.

I wanted to hear your opinion before proceeding. As I see it, one option would be to have a sort of a side-channel to stash `nodeId`, `templateName` and `initialData`, similar to what we had with `window` but I would make it scoped, not global. The other option is to update `ComponentActionsProps.onRun` to carry arguments needed for the modal, but that would require a lot of changes to the shared UI contracts (`ComponentBase`, safe mappers, tests...). What do you think we should do?

lucaspin (commented):
One small question

lucaspin:
@re-pixel it seems like the modal is not opening at all right now, and the hook endpoint is being called directly.

---

One suggestion on how to approach this: since this modal for emitting events is only used for start triggers, what if we had a generic modal host owned by CanvasPage, which the mappers can use to render custom content?

Something like this:

```typescript
type TriggerActionContext = {
  invokeNodeTriggerHook: (hookName: string, parameters: unknown) => Promise<void>;
  openModal: (modal: {
    title?: React.ReactNode;
    description?: React.ReactNode;
    content: (ctx: { close: () => void }) => React.ReactNode;
  }) => void;
};
```

Then the start mapper can do:

```typescript
actions.openModal({
  title: "Run trigger",
  content: ({ close }) => (
    <RunModal
      nodeName={node.name}
      template={template}
      onClose={close}
      onRun={(payload) =>
        actions.invokeNodeTriggerHook("run", {
          template: template.name,
          payload,
        })
      }
    />
  ),
});
```

---

CanvasPage owns the modal shell and state; mappers can request modal content. With that, plan would be something like:

1. Add a generic modal host owned by CanvasPage.
2. Add actions.openModal(...) to mapper action context.
3. Let the start/manual trigger mapper render its own Run button and open a custom run modal through actions.openModal(...).
4. On modal submit, call actions.invokeNodeTriggerHook("run", { template, payload }).
5. Remove the generic node header Run button/affordance entirely.
6. Remove onRun from generic component/canvas node action props and block plumbing.
7. Split handleRun into explicit functions: handleInvokeStartTriggerRunHook or mapper-local action for manual run, and handleReEmit / emitNodeEvent for re-emitting historical trigger events.
8. Rename or replace EmitEventModal, since it is currently tied to the old “run any node / emit arbitrary event” behavior. The new modal should be start-trigger-specific, and can probably live inside start.tsx

pkg/triggers/start/start.go
```diff
@@ -106,11 +114,82 @@ func (s *Start) Setup(ctx core.TriggerContext) error {
...
+					Label:    "Template",
+					Type:     configuration.FieldTypeString,
+					Required: true,
+				},
+				{
+					Name:  "payload",
+					Label: "Payload Override",
+					Type:  configuration.FieldTypeObject,
+				},
+			},
```
forestileao, line 133: I think we don't need this `Payload Override`. since we can configure the payload template directly in the manual run.

pkg/triggers/start/start.go
```diff
@@ -106,11 +114,82 @@ func (s *Start) Setup(ctx core.TriggerContext) error {
...
+		}
+	}
+
+	if !found {
+		return nil, fmt.Errorf("template %q not found; available: %s", templateName, strings.Join(names, ", "))
+	}
+
+	if override, ok := ctx.Parameters["payload"].(map[string]any); ok {
+		payload = override
+	}
```
forestileao, line 182: then we can remove this part

web_src/src/ui/EmitEventModal/index.spec.tsx
forestileao, line 1: the removal of EmitNodeEvent from backend, cli and proto will be in another PR?

re-pixel: yes i try to follow this 
image

pkg/triggers/start/start.go
re-pixel: Just to clarify, that override is used for the modal that prompts user to change the template's message. Do you imply that we should get rid of this modal?

image

web_src/src/pages/workflowv2/lib/canvas-node-preparation.ts
```diff
@@ -285,3 +310,46 @@ function buildActionContext(queryClient: QueryClient, canvasId: string, nodeId:
...
+    void queryClient
+      .invalidateQueries({ queryKey: [...canvasKeys.events(), canvasId] })
+      .catch((error) => console.error("[CanvasPage] Failed to invalidate events query:", error));
+    void queryClient
+      .invalidateQueries({ queryKey: [...canvasKeys.nodeEvents(), canvasId, nodeId] })
+      .catch((error) => console.error("[CanvasPage] Failed to invalidate node events query:", error));
+    void queryClient
+      .invalidateQueries({ queryKey: canvasKeys.nodeEventHistory(canvasId, nodeId) })
+      .catch((error) => console.error("[CanvasPage] Failed to invalidate node event history query:", error));
+  };
```
forestileao: please justr confirm if we need these invalidates. since the events already come from the websocket

forestileao (approved):

PR #4506, opened 2026-04-30
fix: add worker backpressure and cancellation handling

AlPairo:
## Summary

Adds bounded worker polling and cancellation-aware processing for the main worker loops.

## Changes

- Added shared worker concurrency/batch limit of `25`.
- Added per-tick polling limits to node, event, request, queue, integration, and webhook workers.
- Updated pending list queries to accept limits and use stable oldest-first ordering.
- Made semaphore acquisition use worker context cancellation.
- Made RabbitMQ reconnect sleeps respect cancellation.
- Prevented follow-up messages from being published when stale or already-processed records are skipped.
- Tightened stale request locking for node and integration requests.


shiroyasha:
Thank you for opening this PR! The work here looks really impressive. 

Unfortunately, we didn't review it on time, and now it is outdated. I’ll close this PR for now. Thanks again for taking the time to contribute.

PR #4509, opened 2026-04-30
fix: propagate DeleteIntegration transaction failures

AlexDio123:
## What changed
- Updated `DeleteIntegration` to return the transaction error when the delete transaction fails.
- This includes failures while deleting integration webhooks or soft-deleting the integration record itself.

## Why
- Previously, `DeleteIntegration` could return success even when the transaction failed internally.
- That creates false-success responses and can mislead callers/UI into thinking deletion completed.

## How
- Added explicit error handling after `database.Conn().Transaction(...)`.
- If the transaction returns an error, the handler now returns that error instead of `DeleteIntegrationResponse{}`.

## Impact
- Behavior is now stricter and more correct: failed delete operations surface to clients.
- Clients that previously saw success in edge failure paths will now receive an error (expected and correct behavior).

## Notes
- This PR is intentionally scoped to error propagation only (no broader refactor of the delete flow).


shiroyasha:
Looks correct! Assigning @lucaspin to review.

PR #4512, opened 2026-04-30
fix: clarify nested schema discovery in CLI index

### LINKED ISSUE #4276: CLI: nested schema for Daytona bootstrap
### Context
`daytona.createRepositorySandbox` exposes `bootstrap: object (optional)`.

### Observed behavior
`superplane index` does not describe the nested structure, so users infer shape from examples only, e.g.:
- `{ from: "file", path: "..." }`
- `{ from: "inline", script: "..." }`

### Why this is a problem
- Hard to configure correctly without trial-and-error.
- Poor DX for both humans and agents.

### Suggested improvement
In `superplane index` output, include nested schema details for object fields (bootstrap, retry, etc.), including:
- keys and types
- required/optional
- allowed enum values
- examples

Vukotije:
Closes #4276 

## Summary

This fixes the Daytona bootstrap discoverability issue in the shared `superplane index` CLI path.

`daytona.createRepositorySandbox` already exposes `bootstrap` as an optional object with nested schema metadata. The nested fields, enum options, defaults, placeholders, and visibility conditions are present in the structured index response. This was not an integration schema problem.

The problem was that the default text renderer for `superplane index` only displayed top-level configuration fields. For nested fields like `bootstrap`, users only saw that the field was an `object`, but not what keys the object accepts.

Current text behavior looked like:

```text
Configuration:
  NAME              TYPE                  REQUIRED  DESCRIPTION
  snapshot          integration-resource  no        Base environment snapshot for the sandbox
  target            string                no        Target region for the sandbox
  autoStopInterval  number                no        Time in minutes before the sandbox auto-stops
  repository        string                yes       Repository URL to clone
  env               list                  no        Environment variables to set in the sandbox
  secrets           list                  no        Inject organization secret keys as files or environment variables after the sandbox starts
  bootstrap         object                no        Execute script after the sandbox is running, and the repository is cloned
```

That made users and agents infer valid shapes from examples only, such as:

```json
{ "from": "file", "path": "scripts/bootstrap.sh" }
```

```json
{ "from": "inline", "script": "npm ci && npm test" }
```

This PR keeps the default text output compact (top-level fields only). Instead of printing an extra line after the configuration table, **guidance lives in command help**: users are pointed to `-o json` / `-o yaml` with `--name` when they need nested schema fields, enum options, defaults, and conditions—matching structured output already returned by the API.

For the Daytona case, the intended detailed command is:

```bash
superplane index actions --name daytona.createRepositorySandbox -o yaml
```

That structured output includes the nested schema under `configuration[].typeOptions.object.schema`, for example:

```yaml
configuration:
  - name: bootstrap
    type: object
    required: false
    typeOptions:
      object:
        schema:
          - name: from
            type: select
            required: true
            default: inline
            typeOptions:
              select:
                options:
                  - label: Inline Script
                    value: inline
                  - label: Repository File
                    value: file
          - name: script
            type: text
            required: false
            placeholder: npm ci && npm test
            visibilityConditions:
              - field: from
                values:
                  - inline
          - name: path
            type: string
            required: false
            placeholder: scripts/bootstrap.sh
            visibilityConditions:
              - field: from
                values:
                  - file
```

This improves DX for both humans and agents without duplicating the full structured schema renderer inside the compact text table. The same pattern applies beyond Daytona to other nested configuration (object/list schemas, select options, defaults, placeholders, visibility and required conditions).

## What changed

- **`superplane index`**: extended `Long` help and kept examples showing `-o json` / `-o yaml` with `--name` for full nested schema inspection.
- **`superplane index actions`** and **`superplane index triggers`**: added `Long` help with the same guidance (subcommand `--help` does not show the parent command’s long description).
- **Tests**: assert that text configuration output stays a single top-level row for nested object fields (e.g. `bootstrap` / `object`) and does not inline nested enum labels—detail remains in JSON/YAML output.


pkg/cli/commands/index/actions.go
```diff
@@ -197,7 +197,89 @@ func renderConfigurationText(stdout io.Writer, configuration []openapi_client.Co
 		_, _ = fmt.Fprintf(writer, "  %s\t%s\t%s\t%s\n", field.GetName(), field.GetType(), required, field.GetDescription())
 	}
 
-	return writer.Flush()
+	if err := writer.Flush(); err != nil {
+		return err
+	}
+
+	if hasDetailedConfigurationMetadata(configuration) {
```
lucaspin: I don't think we need all this logic here. I think we can put this message:

> use -o json or -o yaml with --name to inspect nested schema fields, enum options, defaults, and conditions.

In the command help message instead of in the output.

shiroyasha:
LGTM

PR #4515, opened 2026-04-30
feat: /api/v1/integrations exposes capabilities

lucaspin:
Update /api/v1/integrations API to return capabilities and capability groups + flag for indicating if an integration supports the new setup flow. Registry was updated to expose methods for registering setup providers during an integration registration.


protos/integrations.proto
```diff
@@ -52,8 +51,33 @@ message IntegrationDefinition {
...
+
+  Type type = 1;
+  string name = 2;
+  string label = 3;
+  string description = 4;
   repeated Configuration.Field configuration = 5;
-  repeated Actions.Action actions = 6;
-  repeated Triggers.Trigger triggers = 7;
-  string instructions = 8;
+  repeated Actions.OutputChannel output_channels = 6;
```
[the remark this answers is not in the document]
lucaspin: UI is not using that information anymore

lucaspin:
👍

protos/integrations.proto
```diff
@@ -52,8 +51,33 @@ message IntegrationDefinition {
...
+  //
+  // NOTE: should be removed once we migrate every integration to the new flow
+  //
+  repeated Configuration.Field configuration = 8;
+  string instructions = 9;
+}
+
+message CapabilityDefinition {
+  enum Type {
+    TYPE_UNKNOWN = 0;
```
shiroyasha, line 67: Why do we need type_unknown? Backward compatibility?

lucaspin: I think it's a common practice to have the default value (0) to be something that means nothing: https://protobuf.dev/best-practices/dos-donts/#unspecified-enum

PR #4516, opened 2026-04-30
fix: update GitHub component mock payloads with comprehensive data

### LINKED ISSUE #3617: Component docs and configuration autocomplete are incomplete
Noticed with github.onIssue, the [example data](https://github.com/superplanehq/superplane/blob/main/pkg/integrations/github/example_data_on_issue.json) is contrived and doesn't demonstrate full payload, which is propagated to [docs](https://docs.superplane.com/components/github/#example-data-1).

This payload is also used to drive `{{` expressions autocomplete until the first event arrives — which leaves users with false impression that the component is poor.

This should be investigated more broadly, any other example payloads that do not reflect reality should be tracked with issues.

MahmoudFawzy1992:
Fixes #3617 

**What changed:**
The original mock data for GitHub webhooks (specifically `onIssue`) was contrived and missing standard nested objects (labels, sender schemas, repository data, etc.), which broke the expressions autocomplete engine for users.

I updated `example_data_on_issue.json` with a comprehensive, real-world webhook payload. 
*Scope Expansion:* I also audited and updated the other GitHub payload examples (`issue_comment`, `pull_request`, `release`, `branch_created`, `tag_created`) to ensure parity across all GitHub triggers.


shiroyasha:
@MahmoudFawzy1992 nice! Thanks for opening this PR.

I see that the example data is referencing an existing organization https://github.com/acme-labs, which is a bit weird.

@forestileao what is our usual practice here, what do we use for example data?

forestileao:
@MahmoudFawzy1992 Let's prefer using fake data, but with same datatype. Just checkled and in fact some example jsons were missing some fields

MahmoudFawzy1992:
@shiroyasha @forestileao
Thanks for the review and the heads-up!

I've gone through and scrubbed all the JSON payloads across the entire GitHub integration. All 28 example files now strictly use generic example-org, example-repo, and example.com identifiers so we aren't referencing any real accounts.

I also did another pass over the schemas to ensure any lingering missing fields (like the root organization object) were added back in with standard mock data.

Let me know if everything looks good to go!

markoa:

PR #4528, opened 2026-04-30
fix: expand append connector handle hitbox

forestileao:
This updates canvas append handles so the visible connector affordance behaves like the actual connection handle.

The append connector line and plus button are now rendered inside the React Flow source handle, making the whole visible append control part of the connection hitbox. Dragging from the append line or plus icon now starts a connection, and dragging an input handle onto those visible append controls can complete the connection more intuitively.

This applies to both single-output end nodes and multi-output channel append handles.

**Changes**
- Added a reusable `AppendSourceHandle` wrapper for append connector handles.
- Moved append line and plus button visuals inside the React Flow source handle.
- Extended the append source handle pseudo-hitbox to cover the visible line/button area.
- Kept existing click-to-append behavior.
- Updated the Block test mock to render React Flow handle children and support combined class names.

### DIFF
--- a/web_src/src/ui/CanvasPage/Block.spec.tsx
+++ b/web_src/src/ui/CanvasPage/Block.spec.tsx
@@ -1,4 +1,5 @@
 import { fireEvent, render, screen } from "@testing-library/react";
+import type React from "react";
 import { describe, expect, it, vi } from "vitest";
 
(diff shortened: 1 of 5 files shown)


forestileao:
👍

PR #4532, opened 2026-05-01
fix: show full error messages instead of truncating at 300px

### LINKED ISSUE #4163: Error messages are truncated hiding critical debugging information
Error information is truncated on the details tab when an erroring request is sent via a component.

Image

shaheerhassan1:
### What

Removes truncation from error message spans in the Errors Console and Runs Console, so the full error message is always visible to the user.

### Why

Closes #4163

Error messages in the Details tab were being cut off with CSS `truncate` and `max-w-[300px]`, hiding critical debugging information. When a component fails, the full error message is essential for diagnosing the problem a hard 300px cap makes that impossible.

### Changes

- `ErrorsConsoleContent.tsx` replaced `truncate max-w-[300px]` with `break-words min-w-0` on the resultMessage span
- `ExecutionRow.tsx` same fix applied to the errorMessage span

### Before / After

Before: Error message cuts off mid-sentence with an ellipsis  
After: Full error message wraps and remains fully readable


markoa:

PR #4537, opened 2026-05-02
fix(daytona): show running status while jobs are in progress

### LINKED ISSUE #3261: Daytona: Running jobs should not say "No data returned"
This is misleading and discouraging, the process is ongoing:

Image

cedaesca:
## What changed

The Daytona component "Running Jobs" view showed "No data returned" while jobs were actively in progress, which was misleading and discouraging — users assumed the integration was broken when it was actually working.

This PR updates the empty-state copy so that when a job is in the `running` state, the view shows `Pending…` instead of "No data returned".

## Why

The previous message implied an error or absent data, but the process was ongoing. See the issue for the screenshot of the misleading state.

## How

Swapped the two "No data returned" literals in `getExecutionDetails()` of `web_src/src/pages/workflowv2/mappers/daytona/base.ts` for "Pending…" — the same branch is hit while an execution is running and when it finishes without a payload, so a neutral copy fixes both cases without state-aware logic (per the issue author's "copy change for now" note).

## Related issues

Fixes #3261

### DIFF
--- a/web_src/src/pages/workflowv2/mappers/daytona/base.spec.ts
+++ b/web_src/src/pages/workflowv2/mappers/daytona/base.spec.ts
@@ -1,0 +1,150 @@
+import { isValidElement, type ReactElement } from "react";
+import { describe, expect, it } from "vitest";
+
+import { baseMapper } from "./base";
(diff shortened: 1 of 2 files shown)


shiroyasha:
Thank you for opening this pr! Unfortunately, we didn't review it on time, and it seems that the linked issue was already fixed.

I'll close this PR for now.

PR #4538, opened 2026-05-02
feat: support stdin (-f -) for roles and groups create/update

### LINKED ISSUE #4526: CLI: stdin -f for roles and groups create/update
## Summary

Extend the same **`-f -` / stdin** input path that exists for `superplane secrets create|update` to **`roles`** and **`groups`** create/update, so resource YAML can be piped or read from memory without a filesystem path.

## Background

- Closed by the same pattern: secrets CLI now treats `-` as stdin (see #4232 and the `parseSecretInput` approach in `pkg/cli/commands/secrets/common.go`).
- `roles` and `groups` still read `-f` only via `os.ReadFile` in `pkg/cli/commands/roles/common.go` (`parseRoleFile`) and `pkg/cli/commands/groups/common.go` (`parseGroupFile`).

## Problem

**Current:** `superplane roles create|update` and `superplane groups create|update` require a real file path for `-f`. Scripts, CI, and agents that emit YAML on stdout (or avoid temp files) cannot use the same `… | superplane … -f -` workflow as secrets.

**Proposed:** When `-f` is `-`, read the resource body from stdin; otherwise keep current `os.ReadFile` behavior. Update flag descriptions to document “path … or `-` to read from stdin”, consistent with secrets.

## Use case

Same as #4232 for this resource class: non-interactive provisioning (pipelines, heredocs, `yq`/template output) without an extra file lifecycle, and **consistent CLI behavior** across org resource commands.

## Scope

- **In scope:** `pkg/cli/commands/roles`, `pkg/cli/commands/groups` — mirror secrets (pass stdin, e.g. cobra `InOrStdin()`, and tests analogous to `pkg/cli/commands/secrets/common_test.go`).
- **Out of scope for this issue:** `canvases` `-f` (file, directory, or URL — different semantics; track separately if needed).

AleksanderDudek:
## Summary

Extends the `-f -` / stdin input path that already exists for `superplane secrets create|update` to `roles` and `groups` create/update, so resource YAML can be piped or read from memory without a filesystem path.

## Why

`roles` and `groups` still read `-f` only via `os.ReadFile`. Scripts, CI pipelines, and agents that emit YAML on stdout cannot use the same `... | superplane ... -f -` workflow as secrets — creating an inconsistency in CLI behavior across org resource commands.

## How

- `pkg/cli/commands/roles/common.go`: renamed `parseRoleFile(path)` → `parseRoleInput(path, stdin io.Reader)`, adding `-` path check that reads from `stdin` instead of the filesystem
- `pkg/cli/commands/groups/common.go`: same rename and stdin support for `parseGroupInput`
- Updated `create.go` and `update.go` in both packages to pass `ctx.Cmd.InOrStdin()`
- Updated `-f` flag descriptions in `root.go` for both packages to document `- to read from stdin`
- Added `common_test.go` for both packages with 5 tests each, mirroring `pkg/cli/commands/secrets/common_test.go`:
  - file path read
  - file path rejects unknown fields
  - stdin read
  - stdin rejects unknown fields
  - empty stdin returns error

## Related issues

Closes #4526

### DIFF
--- a/pkg/cli/commands/groups/common.go
+++ b/pkg/cli/commands/groups/common.go
@@ -23,11 +23,20 @@ type groupResource struct {
 	Spec       *openapi_client.GroupsGroupSpec     `json:"spec,omitempty"`
 }
 
-func parseGroupFile(path string) (*groupResource, error) {
-	// #nosec
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read resource file: %w", err)
+func parseGroupInput(path string, stdin io.Reader) (*groupResource, error) {
+	var data []byte
+	var err error
+	if path == "-" {
+		data, err = io.ReadAll(stdin)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read from stdin: %w", err)
+		}
+	} else {
+		// #nosec
+		data, err = os.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read resource file: %w", err)
+		}
 	}
 
 	apiVersion, kind, err := core.ParseYamlResourceHeaders(data)
--- a/pkg/cli/commands/groups/common_test.go
+++ b/pkg/cli/commands/groups/common_test.go
@@ -1,0 +1,56 @@
+package groups
+
+import (
+	"os"
+	"strings"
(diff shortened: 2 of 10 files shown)


markoa:

PR #4541, opened 2026-05-02
feat: new UI for integration setup + semaphore setup provider implementation

lucaspin:
Main changes:
- Implementation of the IntegrationSetupProvider interface for the Semaphore integration
- New UI components and the new integration setup flow

### UI

Three main components:
- CapabilityBasedIntegrationDetails - details page for integrations created through the new flow. The page for integrations created using the old flow was moved to to LegacyIntegrationDetails component.
- IntegrationSetup - main and child components for the new setup flow - submitting and reverting steps, selecting capabilities, ...
- IntegrationDetailsRoute - helper component to route to the appropriate integration details page, depending on how the integration was setup.

### SetupProvider Implementation

Important to note here that both implementations need to work at the same time, so the old Sync()-based flow was still kept.


lucaspin:
👍

PR #4547, opened 2026-05-03
Add Docker prerequisite to Quick Start section

ceyanesb:
The Quick Start section assumes Docker is installed but doesn't explicitly state this as a requirement. This can be confusing for users unfamiliar with Docker or those setting up SuperPlane for the first time.
Added a "Prerequisites" note at the beginning of the Quick Start section to make the Docker requirement clear before users attempt to run the demo container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ A few concrete things teams build with SuperPlane:
 
 ## Quick start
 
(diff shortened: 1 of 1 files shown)


shiroyasha:
Thank you for opening this PR! I'll close this PR as we will not proceed with this change.

PR #4603, opened 2026-05-03
fix: preserve native type when JSON config field is a single expression

### LINKED ISSUE #4499: HTTP action: expressions in JSON body fields resolve to strings, breaking numeric APIs
## Problem

```yaml
json:
  random_steering:
    pool_weights:
```

## Workaround

## Suggestion

Either:

## Context

juanmorellana2021:
## Problem

Closes #4499

When an HTTP action node has a JSON body field whose value is a single template expression — e.g. `"amount": "{{ payload.price }}"` — the resolved value is always coerced to a string via `fmt.Sprintf("%v", value)`. This means numeric and boolean values are serialised as `"42"` or `"true"` instead of `42` or `true`, breaking downstream APIs that expect native JSON types.

## Root Cause

`ResolveTemplateExpressions` uses `expressionRegex.ReplaceAllStringFunc`, which operates exclusively on strings. Even when the expression evaluates to a `float64` or `bool`, the callback must return a `string`, so the native type is lost.

## Fix

Added a fast path before the `ReplaceAllStringFunc` loop: if the entire field value (after trimming whitespace) is exactly one expression, `ResolveExpression` is called directly and its raw `any` result is returned — preserving the native Go type for correct JSON serialisation downstream.

Interpolated strings such as `"count={{ expr }}"` continue to work as before because they contain surrounding text and therefore take the existing string-replacement path.

## Changes

- `pkg/workers/contexts/node_configuration_builder.go` — single-expression fast path in `ResolveTemplateExpressions`
- `pkg/workers/contexts/node_configuration_builder_test.go` — updated three assertions that expected legacy string coercion; added `Test_NodeConfigurationBuilder_SingleExpression_PreservesNativeType` covering numeric, float, boolean, and mixed-interpolation cases

### DIFF
--- a/pkg/workers/contexts/node_configuration_builder.go
+++ b/pkg/workers/contexts/node_configuration_builder.go
@@ -220,6 +220,13 @@ func (b *NodeConfigurationBuilder) ResolveTemplateExpressions(expression string)
 		return expression, nil
 	}
 
+	// If the entire field value is a single expression, return the raw resolved value
+	// to preserve native types (number, bool, object, array) for JSON serialization.
+	// Without this, fmt.Sprintf coerces everything to a string, breaking numeric/bool APIs.
+	if matches := expressionRegex.FindStringSubmatch(expression); len(matches) == 2 && strings.TrimSpace(expression) == matches[0] {
+		return b.ResolveExpression(matches[1])
+	}
+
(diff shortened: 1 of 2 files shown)


markoa:

PR #4608, opened 2026-05-03
feat: capability selection as a step in new setup flow

lucaspin:
Right now, the setup flow initiates after the user chooses the capabilities they want. This assumes all capabilities are selectable for all possible flows, which is not the case. For example, for the GitHub integration, the `github.getWorkflowUsage` component can only be used if you are connected to an organization, and not a user account. Therefore, it should not be offered at all as a possible capability if you are connected to a user account.

---

This pull request moves capability offering and selection into the setup flow as well. Capability selection becomes a new kind of step, so it can be placed as the first step, or later down the flow. This allows integration implementations to properly control which capabilities to offer for selection depending on flow chosen by the user.

### DIFF
--- a/.go-coverage-baseline.json
+++ b/.go-coverage-baseline.json
@@ -1,5 +1,5 @@
 {
-  "minTotalCoverage": 44.5,
+  "minTotalCoverage": 44.6,
   "minCoverageByPackage": {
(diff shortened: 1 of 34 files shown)


lucaspin:
👍

Issue #4286, opened 2026-04-21
Docs: max event payload size (~64KiB)


AleksandarCole:
### Context
Using `http` component to fetch GitHub compare `.diff` can produce large bodies.

### Observed behavior
Workflow failed with:

```
event payload too large: 103293 bytes (max 65536)
```

HTTP call itself succeeded, but payload could not traverse event graph due to ~64KiB limit.

### Implication
Large diff/content should not be passed through normal event payload path. Use commit metadata already present or delegate large fetch/review to external agent systems.

### Suggested improvement
- Document payload size limits clearly in Data Flow and component docs.
- Consider chunking/external blob references for larger payload patterns.

PR #4612, opened 2026-05-04
docs: add payload size limits to HTTP Request component (#4286)

tjhub1983:
## Summary

Adds documentation about the 64KB payload size limit for the HTTP Request component.

**Before:** No mention of size limits.

**After:** New "Payload Size Limits" section documents the 64KB limit and alternatives for large data.

## Changes

- `docs/components/Core.mdx`: Added Payload Size Limits section with alternatives

## Test

Build the docs and verify the new section appears.


markoa:

PR #4614, opened 2026-05-04
fix(errors-console): show full error message text in the list

tjhub1983:
## Summary

Shows the full error message in the errors console instead of truncating it.

**Before:** Error messages truncated to `max-w-[300px]` with `truncate` class.

**After:** Full error message displayed with `break-words` to show all text.

## Changes

- `web_src/src/pages/workflowv2/ErrorsConsoleContent.tsx`: Changed className from `truncate max-w-[300px]` to `break-words`


shiroyasha:
We'll do major UI refactoring in this area, so unfortunately this PR won't apply anymore.
I'll close it for now.

PR #4617, opened 2026-05-04
fix: core graphql icon

### LINKED ISSUE #4483: GraphQL component color is darker than the rest of the core components

barthc:
Closes #4483. The simple fix here is to use CSS `opacity` to adjust the GraphQL SVG.


markoa:

Issue #1742, opened 2026-01-21
Don't ask for SMTP setup when you are setting up a local demo installation of superplane


shiroyasha:
<img width="1322" height="746" alt="Image" src="https://github.com/user-attachments/assets/f5f2a563-791b-456e-bef7-136baa7405a4" />

AgustinRamiroDiaz:
Hello! I have some questions: 

1. Wouldn't there be some cases where we actually want this set up? I imagine that someone may want to test superplane locally with smtp for mail notifications
2. What exactly are you referring as `local demo installation`? I see that there are many ways to install superplane

PR #4618, opened 2026-05-04
feat(setup): allow skipping SMTP setup

AgustinRamiroDiaz:
# What changed

Now you can skip the SMTP set up from the "SMTP setup screen"

image

# Why

I've miss-clicked on "Set up SMTP" and was not able to skip it, so I had no way of amend my mistake. I've found that a workaround was to refresh the webpage and start all over, but it's not a nice UX 

# How

Since it's a small change and I didn't ask for permission first, I've tried to leave the smallest footprint possible by reusing the preexisting function `handleSkipSMTP`

I've also noticed that there are better ways to do the wizard with form libraries like `react-hook-form`, and doing validation with something like `zod`. We can discuss this as future enhancements

# Related issues

Kind of related to https://github.com/superplanehq/superplane/issues/1742 , but not exactly the same issue

# No breaking changes


shiroyasha:

PR #4624, opened 2026-05-04
fix: corregir etiquetas y agregar campos a ListBlueprintsRequest y Li…

Fabricioz01:
…stBlueprintsResponse


forestileao:
Sorry, but I'll close the PR, the API proto definition changed but there is no  implementation change in the PR.

PR #4628, opened 2026-05-04
fix: handle undefined org on superplane logo button

jakeactually:
Currently when no organization is selected the superplane logo redirects to 404, this pr prevents that

I separated the Logo to another component and kept both scenarios with consistent behaviour so it easier to understand for a user


shiroyasha:

It looks like the issue this PR was addressing has already been fixed, so this change is no longer needed.

PR #4632, opened 2026-05-04
fix: allow replacing default integration instance name

forestileao:
## Summary

- Prevent the integration setup default instance name from being reinserted after the user clears the field.
- Track user edits separately from automatic default initialization.
- Add a Vitest regression covering clearing the default name and typing a replacement.


forestileao:
👍

PR #4633, opened 2026-05-04
feat: migrate Claude to new setup flow

0vertake:
## Summary

Registers Claude with the same integration setup pattern as Semaphore: capability selection, API key collected in a setup step, stored as an editable secret, validated against Anthropic before completion. Legacy integrations keep using config-based apiKey until fully migrated.

## Behavior
- New setup: `SetupProvider` drives the wizard; key lives in secret storage; Sync no longer requires `apiKey` in decoded configuration.
- Legacy setup: unchanged — `apiKey` still required in config; clients still read via `GetConfig("apiKey")`.

https://github.com/user-attachments/assets/b0df236b-58de-4e08-8fd4-92ade1e4800c

### DIFF
--- a/pkg/integrations/claude/claude.go
+++ b/pkg/integrations/claude/claude.go
@@ -11,7 +11,9 @@ import (
 )
 
 func init() {
-	registry.RegisterIntegration("claude", &Claude{})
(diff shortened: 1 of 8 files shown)


bender-rodriguez-unit1:

PR #4637, opened 2026-05-05
fix: update gcp wif service account impersonation and integration setup

WashingtonKK:
### Summary
  
The GCP Workload Identity Federation integration was exchanging a federated STS token but never using it to impersonate a service account. Because federated identities have very limited direct access, syncs using WIF would fail at the first GCP API call. This PR adds the missing impersonation step and the full guided setup flow needed to collect the service account email.

### What changed

  - Service account impersonation (sts.go) — After a successful STS token exchange, SuperPlane now calls the IAM Credentials API (generateAccessToken) to impersonate the configured service account and obtain a short-lived OAuth access token scoped to the Cloud Platform.
  - New workloadIdentityServiceAccountEmail configuration field (gcp.go) — Users must now provide the email of the service account SuperPlane should impersonate when using WIF. The field is only shown for the WIF connection method.
  - Provider resource name normalization — The WIF provider field now accepts both the //iam.googleapis.com/… resource name and the full https://iam.googleapis.com/v1/… IAM URL (copied from the Google Cloud Console), and normalises it internally.
  - Guided setup flow (setup_provider.go) — Adds a SetupProvider implementation with a multi-step wizard: capability selection → connection method → credential entry (service account key or WIF provider + service account). Each step renders contextual instructions from embedded templates.
  - WIF initial sync delay (next_integration_setup_step.go) — IAM bindings can take up to a minute or more to propagate after WIF setup. When setup completes for a WIF integration, the first sync is now scheduled 90 seconds in the future instead of immediately, to avoid transient permission errors.
  - Updated instructions — The in-app instructions now list the required APIs to enable, the correct attribute mapping, the exact federated principal to grant roles/iam.workloadIdentityUser on, and a note about IAM propagation delay.

### Demo

https://youtu.be/IuKwbOKwnco

> Note: GCP IAM bindings can take up to a minute or longer to propagate after setup. To avoid an immediate sync failure, the first sync for WIF integrations is automatically delayed by 90 seconds.


bender-rodriguez-unit1:

PR #4641, opened 2026-05-05
chore: update openai to new setup flow

forestileao:
**Refactors OpenAI integration setup to use a clearer multi-step flow.**

The setup now asks for the API key first, then the optional Base URL, with simple instructions for each step. It also mirrors Semaphore’s setup provider structure with separate handlers, embedded instruction templates, and less repeated setup metadata.

image
image
image


lucaspin:
A couple of questions:
- Shouldn't the "base URL" step come before the "API key" one?
- What does a openAI-compatible provider mean?

pkg/integrations/openai/templates/base-url-instructions.tpl
```diff
@@ -0,0 +1,10 @@
+The Base URL tells SuperPlane which API endpoint to call.
+
+Leave this empty when using OpenAI's hosted API. SuperPlane will use:
```
lucaspin: "Leaving it empty" is a bit weird instruction is a bit weird if the default is already present in the input field

pkg/integrations/openai/openai.go
```diff
@@ -76,16 +80,18 @@ func (o *OpenAI) Cleanup(ctx core.IntegrationCleanupContext) error {
 }
 
 func (o *OpenAI) Sync(ctx core.SyncContext) error {
-	config := Configuration{}
-	if err := mapstructure.Decode(ctx.Configuration, &config); err != nil {
-		return fmt.Errorf("failed to decode configuration: %v", err)
-	}
+	if ctx.Integration.LegacySetup() {
```
lucaspin: Why was this needed? Sync() is never called for integrations using the new setup flow, no?

pkg/integrations/openai/setup_provider.go
```diff
@@ -0,0 +1,402 @@
...
+	case SetupStepEnterBaseURL:
+		return s.onEnterBaseURLSubmit(ctx.Step.Inputs, ctx)
+	}
+
+	return nil, errors.New("unknown step")
+}
+
+func (s *SetupProvider) onCapabilitySelectionSubmit(ctx core.SetupStepContext) (*core.SetupStep, error) {
+	ctx.Capabilities.Request(ctx.Step.Capabilities...)
+	ctx.Capabilities.Available(s.capabilityDiff(ctx.Step.Capabilities)...)
```
lucaspin: Just realized a simpler way to do this - without needing capabilityDiff() - is:

```
ctx.Capabilities.Available(g.allCapabilityNames()...)
ctx.Capabilities.Request(ctx.Step.Capabilities...)
```

shiroyasha:
IMHO, I would kill the "openai compatible" setup and drop support for alternatives.
If we get any real request that this is needed, we can re-add.

forestileao:
image

forestileao:
@lucaspin I will be closing this, maybe we can repopen if this become our focus again

PR #4646, opened 2026-05-05
feat: new GitHub setup flow

lucaspin:
### Changes in GitHub integration

- Implements core.IntegrationSetupProvider interface for GitHub.
- GitHub permissions can be tricky, so a [CapabilityMapper](https://github.com/superplanehq/superplane/pull/4646/changes#diff-fb67fb58a2745402c3883c1ec5305fd5e724a6b0ae337babe916b212436d1ce2) helps us navigate the permissions needed based on the capabilities chosen.
- Requesting more capabilities after having a setup integration is also possible, and OnCapabilityUpdate() may send you you to a new setup flow, if the permissions needed for the new capabilities requested are not yet granted.
- common.NewClient() now returns a custom client, not *github.Client. That allows all components to not have to deal with finding the owner.
- GitHub clients now uses HTTPContext as transport, ensuring timeouts and protections given by HTTPContext are used for GitHub too. Also, makes it easier to write unit tests.

### Core changes

- CapabilityContext included in HTTPRequestContext, so integration implementations can manage capability states in Integration.HandleRequest(). This is needed for the GitHub App setup flow.
- New IntegrationSetupContext. Same reason as the CapabilityContext change above.


lucaspin:
👍

PR #4599, opened 2026-05-03
chore: render large expression numbers without scientific notation

Fieldmatic:
## Summary

Fixes expression template rendering for large numeric IDs so values like `566712522` no longer become scientific notation such as `5.66712522e+08`.

## Changes

- Format `float64`, `float32`, and `json.Number` expression results as decimal strings during template interpolation
- Keep existing string-template behavior unchanged for other value types
- Add no-DB regression tests for root payloads, previous payloads, embedded URLs, fractional floats, and `json.Number`

## Testing

```sh
go test -count=1 ./pkg/workers/contexts -run 'TestNodeConfigurationBuilder_ResolveTemplateExpressions|TestNodeConfigurationBuilder_ResolveExpression_DateWithTimezoneOption'
docker compose -f docker-compose.dev.yml exec app go test -count=1 ./pkg/workers/contexts
docker compose -f docker-compose.dev.yml exec app go test -count=1 ./pkg/grpc/actions/canvases -run 'Test__EmitNodeEvent'
go test -count=1 ./pkg/components/addmemory ./pkg/components/readmemory ./pkg/components/updatememory ./pkg/components/upsertmemory
go test -count=1 ./pkg/integrations/digitalocean
go test -count=1 ./pkg/integrations/grafana -run 'Test.*Time|Test.*Query'
```

Closes https://github.com/superplanehq/superplane/issues/4345

PR #4649, opened 2026-05-06
fix(expressions): preserve type for single-expression template fields

### LINKED ISSUE #4499: HTTP action: expressions in JSON body fields resolve to strings, breaking numeric APIs

WiFoDev:
## Summary

Fixes #4499. When a configuration field's value is exactly one `{{ ... }}` expression, the expression's result type is now preserved (numbers stay numbers, booleans stay booleans, objects stay objects).

Previously every expression result was coerced to a string via `fmt.Sprintf("%v", ...)`, which broke any downstream code expecting typed values — most visibly the HTTP component's `json` body, where
expressions like `'{{ count > 5 ? 0.1 : 0.9 }}'` would render as `"0.9"` and trigger `JSON type error: cannot decode string into field, expected float64` from APIs like Cloudflare Load Balancer.

Mixed text (`prefix-{{ x }}`) and multi-expression strings (`{{ a }}{{ b }}`) keep the existing string-concatenation behavior.

## Before / after

```yaml
# HTTP component config
json:
  pool_weights:
```

| | Resolved JSON body |
| --- | --- |
| Before | `{"pool_weights":{"pool-id-1":"0.9"}}` ❌ |
| After  | `{"pool_weights":{"pool-id-1":0.9}}`   ✅ |

## Scope

- `pkg/workers/contexts/node_configuration_builder.go` — refactor  `ResolveTemplateExpressions` to detect the single-expression case  and return the raw result.
- `pkg/workers/contexts/node_configuration_builder_expr_test.go` — new  no-DB regression tests for type preservation across float, int, bool,  map, string; plus the boundary cases (mixed text, multiple expressions, surrounding whitespace).
- `pkg/workers/contexts/node_configuration_builder_test.go` — update  pre-existing `Build()` integration tests that asserted the old  string-coercion behavior. Bool/int fields now correctly assert their  native types; values pulled via `root()` from JSON-stored canvas  events are `float64` after JSON round-trip, which is also asserted.

## Impact

- The two existing callers of `ResolveTemplateExpressions`  (`pkg/workers/contexts/event_context.go`, `pkg/grpc/actions/canvases/emit_node_event.go`) immediately stringify  the result via `fmt.Sprintf("%v", resolved)` for run-title rendering,  so they are unaffected.
- For end users, expressions that were previously stringified now  preserve their type. Anyone relying on that coercion (e.g. a single  `{{ numericId }}` in a string-typed field) can wrap the expression  with `string(...)` or use the mixed-text form `'-{{ x }}'`.

## Related

Adjacent work: #4599 improves number formatting in the multi-expression / string-concat path. Compatible with this PR — they touch different branches of the same function.

## Validation

```sh
make test PKG_TEST_PACKAGES=./pkg/workers/contexts   # 69/69 pass
make format.go
make lint
make check.build.app
```

### DIFF
--- a/pkg/workers/contexts/node_configuration_builder.go
+++ b/pkg/workers/contexts/node_configuration_builder.go
@@ -216,32 +216,33 @@ func asAnyMap(value any) (map[string]any, bool) {
 }
 
 func (b *NodeConfigurationBuilder) ResolveTemplateExpressions(expression string) (any, error) {
-	if !expressionRegex.MatchString(expression) {
+	matches := expressionRegex.FindAllStringSubmatchIndex(expression, -1)
+	if len(matches) == 0 {
 		return expression, nil
 	}
 
-	var err error
-
-	result := expressionRegex.ReplaceAllStringFunc(expression, func(match string) string {
-		matches := expressionRegex.FindStringSubmatch(match)
-		if len(matches) != 2 {
-			return match
-		}
-
-		value, e := b.ResolveExpression(matches[1])
-		if e != nil {
-			err = e
-			return ""
-		}
-
-		return fmt.Sprintf("%v", value)
-	})
-
-	if err != nil {
-		return nil, err
+	// When the entire field value is a single {{ ... }} expression, return the
+	// raw result so that numbers, booleans, and objects survive into typed
+	// contexts like JSON request bodies. See issue #4499.
+	if len(matches) == 1 && matches[0][0] == 0 && matches[0][1] == len(expression) {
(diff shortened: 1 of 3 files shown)


shiroyasha:
Thank you for opening this PR! Very nice!

It seems that we have two other pull-requests that are attempting to solve this problem but in a more generic way. I'll close this PR for now, but lets take inspiration from this approach when implementing the general fix.

- https://github.com/superplanehq/superplane/pull/4652
- https://github.com/superplanehq/superplane/pull/5628

PR #4651, opened 2026-05-06
feat: validate capability state on canvas update

lucaspin:
When using a component from an integration set up using the new flow, we should check if the component is enabled for use. Integrations already set up using the old flow will keep working as usual.


lucaspin:
👍

PR #4652, opened 2026-05-06
fix: preserve native types for pure expressions in JSON body fields

### LINKED ISSUE #4499: HTTP action: expressions in JSON body fields resolve to strings, breaking numeric APIs

Skiba111:
## Problem

When an HTTP action JSON body field contains a single expression like:

```json
{ "weight": "{{ $[\"Check Env\"].data.result == true ? 0.1 : 0.9 }}" }
```

the expression evaluates correctly to `0.9` (float64), but `ResolveTemplateExpressions` always calls `fmt.Sprintf(\"%v\", value)` via `ReplaceAllStringFunc`, coercing the result to the string `"0.9"`. APIs that expect a numeric field (e.g. Cloudflare Load Balancer `pool_weights`) then reject the request with:

> cannot decode string into field pool_weights, expected float64

Closes #4499

## Root cause

`regexp.ReplaceAllStringFunc` can only return `string`. Every expression result — regardless of its actual type — was formatted as a string before being returned from `ResolveTemplateExpressions`.

## Fix

Added an early-exit path: when the entire field value is a single `{{ … }}` expression with no surrounding text, `ResolveExpression` is called directly and its native return value (`float64`, `bool`, `int`, etc.) is returned unchanged.

Mixed-content fields like `"prefix {{ expr }}"` continue to go through `ReplaceAllStringFunc` and produce strings — no change in behaviour for those cases.

```go
// When the entire value is a single expression with no surrounding text,
// return the native type to preserve numbers, booleans, etc.
if matches := expressionRegex.FindStringSubmatch(expression); matches != nil && matches[0] == expression {
    return b.ResolveExpression(matches[1])
}
```

## Tests

- Updated three existing tests that asserted `"true"` / `"42"` for boolean and integer expressions — those assertions documented the bug; they now assert the correct native types.
- Added `Test_NodeConfigurationBuilder_NativeTypePreservation` which covers the exact scenario from the issue: `float64`, `bool`, `int`, and `string` pure-expression fields, plus one interpolated field to confirm string coercion is still applied there.

### DIFF
--- a/pkg/workers/contexts/node_configuration_builder.go
+++ b/pkg/workers/contexts/node_configuration_builder.go
@@ -220,6 +220,12 @@ func (b *NodeConfigurationBuilder) ResolveTemplateExpressions(expression string)
 		return expression, nil
 	}
 
+	// When the entire value is a single expression with no surrounding text,
+	// return the native type so numbers and booleans are not coerced to strings.
+	if matches := expressionRegex.FindStringSubmatch(expression); matches != nil && matches[0] == expression {
+		return b.ResolveExpression(matches[1])
+	}
+
 	var err error
 
 	result := expressionRegex.ReplaceAllStringFunc(expression, func(match string) string {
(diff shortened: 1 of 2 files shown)


shiroyasha:
Interesting! Thank you for opening this PR.
Assigning @re-pixel to review it as he is working on a similar problem.

@Skiba111 before we could legally merge this pr into our open source project, we would need you to sign the DCO. https://github.com/superplanehq/superplane/blob/main/docs/contributing/commit_sign-off.md

Skiba111:
Hi, I've added the DCO sign-off to the commit. Please let me know if anything else is needed before the merge. Thanks!

PR #4654, opened 2026-05-06
fix: use transaction for HTTP network policy resolution

forestileao:
## Summary

Fixes HTTP network policy resolution so it uses the active DB transaction when called from transaction-scoped code.

## Why

`ResolveHTTPPolicy()` reads installation metadata. When integration/action/trigger code runs inside a transaction and makes an HTTP request, the shared HTTP context can refresh its policy cache and call that resolver.

Before this change, the resolver used `database.Conn()`, which could block when the pool was exhausted because the surrounding worker already held a transaction connection.

## What changed

- Added a transaction-aware HTTP policy resolver.
- Added a transaction-bound registry HTTP context.
- Updated worker and setup/hook paths that run inside transactions to pass their `tx` into HTTP policy resolution.
- Added focused tests for the transaction resolver behavior.


forestileao:
👍

Issue #4647, opened 2026-05-06
Adopt golangci-lint v2 for static analysis - config evaluated, real bugs found


rjeczalik:
## Summary

After evaluating and configuring [golangci-lint v2](https://golangci-lint.run/) for this repository, the resulting configuration runs 38 linters — each selected based on **signal quality**: whether it catches real bugs, correctness issues, or security problems. Style-only and high-false-positive linters were explicitly rejected. The initial scan found real bugs including a nil pointer dereference, a broken schedule calculation, and 60 instances of silently swallowed errors.

## What was done

1. Audited the existing `.golangci.yml` (19 linters, v2 format)
2. Researched all available linters in golangci-lint v2.10.1 against the codebase
3. Tested each candidate linter individually, inspected findings, categorized as bug/correctness/noise
4. Added 22 linters that catch real issues, removed 2 obsolete ones
5. Tuned out style-only checks (staticcheck ST1xxx, gocritic captLocal/singleCaseSwitch/ifElseChain)
6. Every decision is documented in `docs/research/golangci-lint-evaluation.md`

## Real bugs found

Confirmed bugs discovered by the new configuration:

| Finding | Linter | File | Impact |
|---------|--------|------|--------|
| **Nil pointer dereference** — `groupSpec.Role` accessed before nil check at line 49, nil check at line 54 | staticcheck SA5011 | `pkg/grpc/actions/auth/update_group.go` | Runtime panic |
| **Broken schedule calculation** — `time.Add()` return value discarded, weekday adjustment never applied | staticcheck SA4017 | `pkg/triggers/schedule/schedule.go:725` | Wrong schedule timing |
| **Append result discarded** — `append()` assigned to different variable, original slice unchanged | staticcheck SA4010 + gocritic | `pkg/integrations/aws/aws.go:899`, `pkg/integrations/gcp/pubsub/client.go:335` | Data silently dropped |
| **Defer won't run** — `log.Fatalf` called after defer, process exits before cleanup | gocritic | `pkg/web/asset_handler.go:117` | Resource leak on fatal |
| **60 silently swallowed errors** — `err != nil` checked but `nil` returned | nilerr | Multiple files | Errors invisible to callers |
| **Printf format injection risk** — dynamic format string with no arguments | staticcheck SA1006 | `pkg/authentication/authentication.go` | Potential format string bug |
[...]

PR #4655, opened 2026-05-06
fix: avoid nil panic in UpdateGroup when group spec is omitted

inacionery:
## What

`auth.UpdateGroup` panicked on a nil pointer when a client called `GroupsService.UpdateGroup` without a `Group.Spec`.

## Why

The gRPC service nil-checks `req.Group` but passes `req.Group.Spec` straight through:

```go
// pkg/grpc/groups_service.go
if req.Group == nil {
    return nil, status.Error(codes.InvalidArgument, "group must be specified")
}
return auth.UpdateGroup(ctx, domainType, domainID, req.GroupName, req.Group.Spec, s.authService)
```

Inside auth.UpdateGroup, groupSpec.Role was dereferenced before the existing if groupSpec != nil guard a few lines below. A request like {group: {metadata: {…}, spec: null}} reaches this path and crashes the server. This is the staticcheck SA5011 finding cataloged in #4647.

How

- Move the role read inside the existing if groupSpec != nil block so we never dereference a nil pointer.
- Default updatingRole to currentRole so the response shape is preserved when no spec is provided.
- Behavior with a non-nil spec is unchanged: an empty Role still falls back to currentRole, and authService.UpdateGroup is still called only when a spec was provided.
- Added a regression test (TestUpdateGroup/nil_spec_returns_current_state_without_panic) that panics on the old code and passes on the new code.

Alternative considered

Rejecting req.Group.Spec == nil at the gRPC layer with InvalidArgument is a stricter, equally valid contract. I preserved current semantics (return current state) because the existing displayName/description branches already structurally handle the nil case — keeping this a pure bug fix rather than a behavior change. Happy to follow up with the stricter contract if you'd prefer it.

Breaking changes

None.

Refs

Refs #4647 — addresses the SA5011 nil-pointer dereference finding from the linter audit.


shiroyasha (approved):

shiroyasha:
Looks good! Triggered CI to verify the change.

PR #4658, opened 2026-05-06
I'm a perfect fit for this role - Luca Longoni

lucabit-dev:
SuperPlane is the kind of company I want to be working at right now. Early-stage, AI-native, building dev tools in public for other engineers. The thesis that the platform engineering layer is about to be rebuilt is the kind of bet I'd want to be part of.
I've been working primarily inside agentic environments (Claude Code, Cursor, Codex) for about a year, on real production codebases. At MADELU I built a real-time trading platform end to end on Node, TypeScript and React with AWS. I founded Bunzi, a multi-tenant SaaS I ran solo from architecture to production.
Go isn't where my backend sits today, but ramping up on a new language inside an agentic workflow is faster than it used to be.
Building in the open, on a product engineers actually depend on, is the part that closes it for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# I think I'm the perfect fit for this role - Luca Longoni
+
 # SuperPlane
 
 SuperPlane is an **open source control plane for platform engineering**.


markoa:
Hey Luca, thanks for your interest. We review candidates who apply via the job pages on superplane.com/careers.

lucabit-dev:
Hey Markoa, thanks for your reply. I've just sent my application via
superplane.com/careers, it would be great if you can take a look at
it, I really think I'm a perfect fit. I hope to be working side by side
soon.

Best regards,
Luca

PR #4659, opened 2026-05-06
feat(ui): dialog when disabling capabilities currently in-use

lucaspin:
Disabling capabilities is a potentially breaking action, if the capabilities being disabled are currently in-use somewhere. So, we need to ask for confirmation before proceeding.


lucaspin:
👍

PR #4660, opened 2026-05-06
feat: migrate Cursor to new setup flow

0vertake:
## Summary

Registers Cursor with `IntegrationSetupProvider` so new installs use capability selection and stored secrets, while existing integrations keep reading keys from legacy configuration until they are migrated off `LegacySetup()`.

## Changes

- Register Cursor via `RegisterIntegrationWithOptions` and wire `SetupProvider` for capability groups (Agents vs Admin & Usage) and conditional API key steps.
- Teach `NewClient` and `Sync` to read credentials from configuration when `LegacySetup()` is true and from integration secrets otherwise; reuse shared verification for both paths.
- Implement setup steps including skipping the key step when stored secrets already satisfy the selected capabilities, and avoid deleting secrets when reverting from the enter-keys step.
- Extend tests for the setup provider and for `Sync` under the new setup flow (launch-only, admin-only, both keys).

https://github.com/user-attachments/assets/13c26e16-0f8e-46bf-a533-485006731f59


lucaspin:
@0vertake I think we should use multiple steps instead of a single one requesting both keys. For example, the "You may need separate keys for Agents and Admin" message we had before was a way of circumventing not being able to separate things into multiple steps, and it's not OK to have that for this new flow

bender-rodriguez-unit1:

PR #4663, opened 2026-05-06
fix: prevent external cleanup failures from blocking workers

forestileao:
## Summary
- Make webhook cleanup best-effort so provider cleanup failures do not block local webhook deletion
- Treat provider 404s as already-cleaned-up resources for GitHub and Semaphore cleanup
- Prevent soft-deleted webhooks from indefinitely blocking integration cleanup
- Add regression tests for missing provider resources and failed external cleanup

### DIFF
--- a/pkg/workers/webhook_cleanup_worker.go
+++ b/pkg/workers/webhook_cleanup_worker.go
@@ -105,7 +105,7 @@ func (w *WebhookCleanupWorker) processAppInstallationWebhook(tx *gorm.DB, webhoo
 	})
 
 	if err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


forestileao:
👍

pkg/workers/webhook_cleanup_worker.go
```diff
@@ -105,7 +105,7 @@ func (w *WebhookCleanupWorker) processAppInstallationWebhook(tx *gorm.DB, webhoo
 	})
 
 	if err != nil {
-		return err
+		w.log("Best-effort cleanup failed for webhook %s: %v", webhook.ID, err)
```
lucaspin, line 108: We should try a couple of times, right? What if there's a network hiccup?

PR #4664, opened 2026-05-06
fix: stop merge queue items from retrying on expression errors

forestileao:
## Summary
Fixes an infinite retry loop in the merge component when `stopIfExpression` fails during evaluation.

## What was happening
1. `NodeQueueWorker` picked up a queue item for a merge node.
2. The merge component dequeued the queue item.
3. The merge component evaluated `stopIfExpression`.
4. The expression failed because it referenced a node that was not in the execution chain.
5. `ProcessQueueItem` returned that error.
6. Because processing runs inside a DB transaction, returning the error rolled back the whole transaction.
7. The queue item deletion was rolled back too.
8. The same queue item stayed in the queue and was picked up again every worker tick.

## Fix
- Treat merge stop-expression evaluation errors as failed merge executions.
- Mark


forestileao:
@shiroyasha I'll need your review on this to know if the issue described in description on this PR makes sense

PR #4670, opened 2026-05-07
chore: cache compiled regex in Predicate.Matches

neeldutta1994:
## What

  `Predicate.Matches` (`pkg/configuration/any_predicate_list.go`) called `regexp.MatchString(p.Value, value)` on every invocation, which compiles the regex each time. Cache compiled regexes (and compile errors) in a `sync.Map` keyed by pattern string so repeated evaluations of the same predicate set reuse the compiled regex.

  ## Why

  `MatchesAnyPredicate` is on the hot path of webhook handlers — gitlab `on_tag` / `on_issue`, elastic `on_alert` / `on_case_status_change`, and others — and is often called inside per-label / per-tag loops, so the same patterns get recompiled N×M times per event.

  Benchmark (`BenchmarkMatchesAnyPredicate`, 4 predicates × 5 values per op):

  | | ns/op | allocs/op | B/op |
  |--|--|--|--|
  | Before | ~16,000 | 503 | 39,316 |
  | After  | ~500    | 0   | 0      |

  ~30× faster, zero allocations.

  ## How

  - Add a package-level `sync.Map` `matchPatternCache` keyed by pattern string, value `matchPatternEntry{re, err}`.
  - New `compileMatchPattern(pattern)` helper: cache-hit returns the stored entry; cache-miss compiles once and stores the result (success or error).
  - `Predicate.Matches` calls the helper for `PredicateTypeMatches`. Public API unchanged.

  Failed compiles are also cached so a malformed pattern doesn't get retried on every call.

  ## Tests

  Added `pkg/configuration/any_predicate_list_test.go`:
  - Correctness tests for all `Predicate` types (equals, notEquals, matches, invalid regex, unknown type) and `MatchesAnyPredicate`.
  - `BenchmarkMatchesAnyPredicate` covering the typical webhook workload.

  ## Breaking changes

  None.

### DIFF
--- a/pkg/configuration/any_predicate_list.go
+++ b/pkg/configuration/any_predicate_list.go
@@ -1,6 +1,9 @@
 package configuration
 
-import "regexp"
+import (
+	"regexp"
+	"sync"
+)
 
 const (
 	PredicateTypeEquals    = "equals"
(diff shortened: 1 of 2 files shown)


markoa:

PR #4672, opened 2026-05-07
feat: add canvas folders backend

forestileao:
## Summary
- add canvas folder models, migrations, and db structure updates
- add gRPC actions/service registration and authorization for canvas folders
- wire folder metadata into canvas serialization and version restore paths

## Stack
- Depends on #4671
- Next: split/canvas-folders-ui

## Validation
- make test PKG_TEST_PACKAGES=./pkg/grpc/actions/canvasfolders

### DIFF
--- a/.go-coverage-baseline.json
+++ b/.go-coverage-baseline.json
@@ -1,9 +1,9 @@
 {
-  "minTotalCoverage": 44.6,
+  "minTotalCoverage": 44.7,
   "minCoverageByPackage": {
     "pkg/authentication": 6.3,
     "pkg/authorization": 52.4,
-    "pkg/cli": 27.8,
+    "pkg/cli": 44.4,
     "pkg/cli/commands/canvases": 35.9,
     "pkg/cli/commands/canvases/models": 100,
     "pkg/cli/commands/events": 39.1,
@@ -29,13 +29,13 @@
     "pkg/grpc/actions/agents": 44,
     "pkg/grpc/actions/auth": 80.8,
     "pkg/grpc/actions/blueprints": 0,
-    "pkg/grpc/actions/canvases": 62.1,
+    "pkg/grpc/actions/canvases": 62.6,
     "pkg/grpc/actions/canvases/changesets": 84.7,
-    "pkg/grpc/actions/canvases/layout": 63.4,
+    "pkg/grpc/actions/canvases/layout": 63.1,
     "pkg/grpc/actions/integrations": 0,
     "pkg/grpc/actions/me": 54.2,
     "pkg/grpc/actions/messages": 0,
-    "pkg/grpc/actions/organizations": 67.2,
+    "pkg/grpc/actions/organizations": 67,
     "pkg/grpc/actions/secrets": 42.6,
     "pkg/grpc/actions/serviceaccounts": 5.5,
     "pkg/grpc/actions/triggers": 84.6,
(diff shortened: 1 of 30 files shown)


db/migrations/20260506152447_add-canvas-folders.up.sql
```diff
@@ -0,0 +1,30 @@
...
+  organization_id uuid NOT NULL,
+  title character varying(128) NOT NULL,
+  background_color character varying(32) NOT NULL DEFAULT 'color_1',
+  sort_order bigint NOT NULL,
+  created_at timestamp without time zone NOT NULL,
+  updated_at timestamp without time zone NOT NULL,
+  CONSTRAINT canvas_folders_pkey PRIMARY KEY (id),
+  CONSTRAINT canvas_folders_organization_id_title_key UNIQUE (organization_id, title),
+  CONSTRAINT canvas_folders_organization_id_fkey FOREIGN KEY (organization_id) REFERENCES public.organizations(id) ON DELETE CASCADE,
+  CONSTRAINT canvas_folders_background_color_check CHECK (background_color IN ('color_1', 'color_2', 'color_3', 'color_4', 'color_5', 'color_6'))
```
shiroyasha: would be better to call them by name, yellow, purple, etc...

pkg/models/canvas_version.go
```diff
@@ -242,7 +243,15 @@ func PromoteToLiveInTransaction(tx *gorm.DB, version *CanvasVersion, nodes []Nod
...
-	return MapCanvasNameUniqueConstraintError(tx.Save(canvas).Error)
+	return MapCanvasNameUniqueConstraintError(tx.
+		Model(&Canvas{}).
+		Where("id = ?", canvas.ID).
+		Updates(map[string]any{
+			"live_version_id": version.ID,
+			"name":            version.Name,
+			"updated_at":      now,
+		}).
+		Error)
```
shiroyasha, line 254: Is this change related to folders? Why?

.go-coverage-baseline.json
```diff
@@ -44,13 +44,13 @@
     "pkg/jwt": 55.3,
     "pkg/lint/examplepayloads": 84.6,
     "pkg/logging": 0,
-    "pkg/models": 9.8,
-    "pkg/networkpolicy": 82.1,
+    "pkg/models": 8.7,
```
shiroyasha, line 47: Is this decreasing the coverage?

pkg/models/canvas_version.go
forestileao: This one is required, we are selecting just some fields to be updated, then the folder_id prop was always being nil, since we were not requesting it from the DB. Causing the folder_id to reset to nil on publish.

There was another fix that was adding folder_id into the select for update, but I was not willing to add unecessary data to it. Updating the specific fields looks more clean

pkg/authorization/interceptor.go
```diff
@@ -261,6 +262,31 @@ func NewAuthorizationInterceptor(authService Authorization) *AuthorizationInterc
...
+		pbCanvasFolders.CanvasFolders_UpdateCanvasFolderPosition_FullMethodName: {
+			Resource:   "canvases",
+			Action:     "update",
+			DomainType: models.DomainTypeOrganization,
+		},
+		pbCanvasFolders.CanvasFolders_DeleteCanvasFolder_FullMethodName: {
+			Resource:   "canvases",
+			Action:     "update",
+			DomainType: models.DomainTypeOrganization,
+		},
```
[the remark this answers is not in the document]
forestileao: makes sense to be update

shiroyasha:
👍

PR #4673, opened 2026-05-07
feat: add canvas folders UI

forestileao:
## Summary
- add home page canvas folder UI, menus, folder sections, and canvas assignment submenu
- add canvas folder data hooks and home page unit coverage
- extend e2e home page coverage for folders

## Stack
- Depends on #4672

## Validation
- docker compose -f docker-compose.dev.yml exec app bash -c "cd web_src && npm run test -- src/pages/home/index.spec.tsx --run"
- make check.lint.ui
- make check.build.ui

### DIFF
--- a/web_src/src/pages/home/CanvasActionsMenu.tsx
+++ b/web_src/src/pages/home/CanvasActionsMenu.tsx
@@ -17,6 +17,7 @@ interface CanvasActionsMenuProps {
   onEdit: (canvas: CanvasCardData) => void;
   canUpdateCanvases: boolean;
   canDeleteCanvases: boolean;
+  permissionsLoading: boolean;
 }
 
 export function CanvasActionsMenu({
@@ -26,6 +27,7 @@ export function CanvasActionsMenu({
   onEdit,
   canUpdateCanvases,
   canDeleteCanvases,
+  permissionsLoading,
 }: CanvasActionsMenuProps) {
   const [isDialogOpen, setIsDialogOpen] = useState(false);
   const deleteCanvasMutation = useDeleteCanvas(organizationId);
@@ -70,7 +72,7 @@ export function CanvasActionsMenu({
         }}
       >
         {!canManage ? (
-          <PermissionTooltip allowed={false} message="You don't have permission to manage this canvas.">
+          <PermissionTooltip allowed={permissionsLoading} message="You don't have permission to manage this canvas.">
             <button
               className="p-1 rounded hover:bg-gray-100 dark:hover:bg-gray-800 text-gray-500 dark:text-gray-400 disabled:opacity-50 disabled:cursor-not-allowed"
               aria-label="Canvas actions"
--- a/web_src/src/pages/home/CanvasActionsMenu.tsx
+++ b/web_src/src/pages/home/CanvasActionsMenu.tsx
@@ -1,0 +1,158 @@
+import { PermissionTooltip } from "@/components/PermissionGate";
+import { Dialog, DialogActions, DialogDescription, DialogTitle } from "@/components/Dialog/dialog";
+import { Button } from "@/components/ui/button";
+import { LoadingButton } from "@/components/ui/loading-button";
+import { DropdownMenu, DropdownMenuContent, DropdownMenuItem, DropdownMenuTrigger } from "@/ui/dropdownMenu";
+import { useDeleteCanvas } from "@/hooks/useCanvasData";
+import { showErrorToast, showSuccessToast } from "@/lib/toast";
+import { MoreVertical, Pencil, Trash2 } from "lucide-react";
+import { useState, type MouseEvent } from "react";
+import { CanvasFolderSubmenu } from "./CanvasFolderSubmenu";
+import type { CanvasCardData, CanvasFolderData } from "./types";
+
+interface CanvasActionsMenuProps {
+  canvas: CanvasCardData;
+  canvasFolders: CanvasFolderData[];
+  organizationId: string;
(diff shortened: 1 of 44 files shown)


web_src/src/pages/home/CanvasActionsMenu.tsx
```diff
@@ -0,0 +1,193 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+    setIsDialogOpen(true);
+  };
+
+  const handleDelete = async () => {
+    if (!canDeleteCanvases) return;
+    const currentPath = location.pathname;
+    const canvasPath = `/${organizationId}/canvases/${canvas.id}`;
+    const isViewingCanvas = currentPath === canvasPath || currentPath.startsWith(`${canvasPath}/`);
```
shiroyasha: Isn't there some more elegant way to figure this out based on props?
Relying on routing info is britle.

web_src/src/pages/home/CanvasActionsMenu.tsx
```diff
@@ -0,0 +1,193 @@
...
+      <div
+        className="flex-shrink-0"
+        onClick={(event: MouseEvent<HTMLDivElement>) => {
+          event.preventDefault();
+          event.stopPropagation();
+        }}
+      >
+        {!canManage ? (
+          <PermissionTooltip
+            allowed={canManage || permissionsLoading}
```
shiroyasha: So while the permissions are loading, this is true?
Also, how can it be that we already have access to canManage on line 90, but the permissions are still loading?

web_src/src/pages/home/CanvasActionsMenu.tsx
```diff
@@ -0,0 +1,193 @@
...
+                allowed={canUpdateCanvases || permissionsLoading}
+                message="You don't have permission to update canvases."
+              >
+                <DropdownMenuItem
+                  onClick={(event: MouseEvent<HTMLElement>) => {
+                    event.preventDefault();
+                    event.stopPropagation();
+                    if (!canUpdateCanvases) return;
+                    onEdit(canvas);
+                  }}
```
shiroyasha: Would be much easier to understand if it would not be inlined:

```
<DropdownMenuItem onClick={onChangeName}>...
```

web_src/src/pages/home/CanvasCardsGrid.tsx
```diff
@@ -0,0 +1,205 @@
...
+            <div className="mb-3">
+              <Text className="line-clamp-2 text-left text-[12px] !leading-normal text-gray-800 dark:text-gray-400">
+                {canvas.description}
+              </Text>
+            </div>
+          ) : null}
+
+          <div className="flex justify-between items-center">
+            <p className="mt-1 text-left text-[11px] leading-none text-gray-500 dark:text-gray-400">
+              {canvas.createdBy?.name ? (
```
shiroyasha: Is there any valid instance where we don't have a createdBy populated?
Having not-necessary fallbacks is a common agent problem.

web_src/src/pages/home/CanvasCardsGrid.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+interface CanvasMiniMapProps {
+  nodes?: SuperplaneComponentsNode[];
+  edges?: SuperplaneComponentsEdge[];
+}
+
+function CanvasMiniMap({ nodes = [], edges = [] }: CanvasMiniMapProps) {
+  const positionedNodes = nodes.filter(
+    (node) => typeof node.position?.x === "number" && typeof node.position?.y === "number",
+  ) as Array<SuperplaneComponentsNode & { position: { x: number; y: number } }>;
```
shiroyasha: This should be a named type.

web_src/src/pages/home/CanvasCardsGrid.tsx
```diff
@@ -0,0 +1,205 @@
...
+  );
+}
+
+interface CanvasMiniMapProps {
+  nodes?: SuperplaneComponentsNode[];
+  edges?: SuperplaneComponentsEdge[];
+}
+
+function CanvasMiniMap({ nodes = [], edges = [] }: CanvasMiniMapProps) {
+  const positionedNodes = nodes.filter(
```
shiroyasha: What does it even mean to have non-positioned nodes?

shiroyasha:
Looks clean. Left a couple of comments.

web_src/src/pages/home/CanvasActionsMenu.tsx
forestileao: I removed the route-aware delete handling from this menu. The menu now only deletes the canvas and shows the toast; route cleanup/navigation should be handled by the canvas page if needed, not inferred here from location.pathname.

web_src/src/pages/home/CanvasActionsMenu.tsx
forestileao: removed it from the canvas actions menu so the tooltip/menu state only depends on the actual permission booleans

web_src/src/pages/home/CanvasCardsGrid.tsx
forestileao: the minimap can only draw nodes with coordinates. I made that explicit with a hasMiniMapPosition type guard, so “non-positioned” here just means nodes that are not drawable in the preview.

forestileao:
@shiroyasha Comments are resolved. Good that it also cleanup some other things we had to update in the home page

forestileao:
👍

PR #4679, opened 2026-05-07
fix: add optimistic feedback for canvas folder updates

forestileao:
## Summary
- Add optimistic cache updates when moving canvases into or out of folders
- Add localized pending states for creating a folder and assigning a canvas
- Add tests for optimistic folder membership updates and rollback


forestileao:
👍

PR #4681, opened 2026-05-07
feat: add backend workflow runs

forestileao:
## Summary
- add workflow_runs and nullable run_id references for runtime records
- create and finalize runs from the workflow engine
- propagate run IDs through queue items, executions, and output events

## Tests
- make test PKG_TEST_PACKAGES="./pkg/models ./pkg/workers"
- git diff --check


pkg/models/canvas_run.go
```diff
@@ -0,0 +1,284 @@
...
+
+	err := query.Find(&runs).Error
+	if err != nil {
+		return nil, err
+	}
+
+	return runs, nil
+}
+
+func CountCanvasRuns(workflowID uuid.UUID) (int64, error) {
```
shiroyasha: Where is this used?

Counting can get very expensive on bigger tables.

forestileao: This should be in the other PR where we have ListRuns.

I will move this to the other.

BTW I understand the concern, but it is indexed and we already have total_counts for ListNodeEventsResponse

forestileao:
👍

PR #4685, opened 2026-05-08
fix: prevent monaco suggest widget clipping in configuration sidebar

### LINKED ISSUE #1804: Monarch editor: autocomplete cropped

Prodigy343:
## What

Fixes #1804 — Monaco's autocomplete (and other overflow) widgets get clipped by the configuration sidebar, making suggestions unreadable in JSON, Text, and XML body fields used by HTTP and other components.

## Why

By default, Monaco appends overflow widgets (suggest, hover, parameter hints) inside the editor's own DOM container. The configuration sidebar (`pkg/web_src/src/ui/componentSidebar`) has both `overflow: hidden` and a `transform`-animated ancestor — which combine to:

1. Clip widgets at the sidebar's bounds (the visible bug from the issue's screenshot)
2. Trap `position: fixed` widgets inside the transformed sidebar's containing block (per CSS spec — a `transform` ancestor becomes the containing block for fixed-positioned descendants)

So `fixedOverflowWidgets: true` alone wasn't enough — the widgets stopped being clipped but rendered with a position offset that grew when the window was resized.

## How

Two-part fix using Monaco's documented overflow-widget API — the same pattern VS Code uses internally:

1. **`fixedOverflowWidgets: true`** — render overflow widgets with `position: fixed`.
2. **`overflowWidgetsDomNode: <body-level node>`** — mount widgets as a child of `<body>` so they escape the sidebar's `transform`-induced containing block.

The body-level node is created lazily by a small new helper at `web_src/src/lib/monacoOverflowWidgets.ts`, which gives it the `monaco-editor` class so Monaco's widget CSS rules still scope correctly when rendered outside the editor.

The two options are applied to the three Monaco-backed renderers used in component configuration:

- `web_src/src/ui/configurationFieldRenderer/ObjectFieldRenderer.tsx` (JSON Payload)
- `web_src/src/ui/configurationFieldRenderer/TextFieldRenderer.tsx`
- `web_src/src/ui/configurationFieldRenderer/XMLFieldRenderer.tsx`

## Verification

Tested locally with `make dev.start`:

1. Created a canvas, added an HTTP component, set Method=POST, Body=JSON.
2. Clicked into the JSON Payload field, typed `${{` to trigger expression autocomplete.
3. **Before:** suggestions clipped at the sidebar's right edge — text like `.chema`, `evious`, `tnand` (matches the original issue screenshot).
4. **After:** dropdown renders cleanly anchored to the cursor, suggestions fully visible (`schema`, `previous`, `bitnand`, etc.), stays anchored when the window is resized.
5. Verified `make check.build.ui` passes.
6. Verified `npm run format` reports no changes.

## Files changed

- `web_src/src/lib/monacoOverflowWidgets.ts` (new — 25 lines, body-level overflow-widget host)
- `web_src/src/ui/configurationFieldRenderer/ObjectFieldRenderer.tsx` (+3)
- `web_src/src/ui/configurationFieldRenderer/TextFieldRenderer.tsx` (+3)
- `web_src/src/ui/configurationFieldRenderer/XMLFieldRenderer.tsx` (+3)

No breaking changes. No public API surface changes.

Closes #1804.


shiroyasha:
Unfortunately, we have already removed all instances of the monaco editor which automatically fix this issue.

Issue #4350, opened 2026-04-23
Canvas Memory: readMemory needs "read all" + addMemory needs atomic create-if-not-exists


bender-rodriguez-unit1:
## Problem 1: readMemory requires matchList

`readMemory` requires at least one match item. There is no way to enumerate all records in a namespace without matching on a specific field.

**Workaround:** Add a common field (e.g., `repo: storejs`) to all records and match on it. Works but feels hacky.

**Suggestion:** Allow empty `matchList` to return all records in the namespace.

## Problem 2: No atomic create-if-not-exists

When two events fire simultaneously (e.g., two `/deploy` comments within seconds), both pass the `readMemory` check (not found) and both proceed to `addMemory`, creating duplicate records and duplicate infrastructure.

**Workaround:** None that fully prevents the race. `upsertMemory` helps with dedup on the memory side, but the downstream `createDroplet` still fires twice.

**Suggestion:** Add an `addMemoryIfNotExists` component (or a flag on `addMemory`) that atomically checks + inserts and emits on `created` / `alreadyExists` channels. This would let the workflow branch on whether the record was newly created or already existed.

## Context

Both issues surfaced during an ephemeral environments workflow where Canvas Memory tracks active preview environments by PR number.

PR #4698, opened 2026-05-08
feat: enable empty matchList for read memory component

re-pixel:
This PR makes match list optional for read memory. It was previously required.

image
image

### Related Issue

https://github.com/superplanehq/superplane/issues/4350


bender-rodriguez-unit1:

PR #4699, opened 2026-05-08
fix: stop canvas work for deleted orgs

forestileao:
## Summary

This stops canvases from continuing to run after their organization has been soft-deleted.

Workers now ignore queued work, events, executions, node requests, and webhook deliveries for canvases whose organization is deleted. The canvas cleanup worker also treats canvases in deleted orgs as cleanup candidates after the org grace period, even if the canvas itself was not soft-deleted.


forestileao:
👍

PR #4701, opened 2026-05-08
fix: batch event retention cleanup

forestileao:
## Summary
Batch event retention cleanup so the worker locks and deletes eligible root event chains in groups instead of one root event per transaction.

This keeps the same safety checks for active work, but reduces repeated queries and transaction overhead when there is retention backlog.


forestileao:
👍

PR #4705, opened 2026-05-08
fix(github): enrich onPRComment payload with full pull request

### LINKED ISSUE #4352: GitHub onPRComment: no branch or SHA in payload makes commit status and deployments difficult
## Problem

`github.onPRComment` uses the `issue_comment` webhook. The payload contains `data.issue` (not `data.pull_request`), and `data.issue.pull_request` only has URLs — no `head.ref`, `head.sha`, or branch information.

This makes it impossible to:
- Publish commit status checks (needs SHA)
- Clone the PR branch directly (needs branch name)
- Reference the PR head commit in downstream components

## Current Workarounds

1. **For branch:** Use `git fetch origin pull/{number}/head` instead of cloning by branch name
2. **For SHA:** Store it in Canvas Memory during `onPullRequest` event, then read it back from memory during `onPRComment` handler

Both workarounds add complexity to what should be a simple `/deploy` command handler.

## Suggestion

Either:
1. Enrich the `onPRComment` payload with the PR head SHA and branch (fetch from GitHub API at trigger time), or
2. Add a `github.getPullRequest` component that fetches full PR details by number, or
3. Document the limitation prominently in the GitHub trigger reference with recommended patterns

## Context

Any slash-command pattern (`/deploy`, `/test`, `/release`) triggered via PR comments will hit this. It is the most natural UX for opt-in workflows but the hardest to implement due to the missing data.

burakugar:
Closes #4352.

## Problem

`github.onPRComment` is wired to GitHub's `issue_comment` webhook. That payload only carries URLs under `data.issue.pull_request`, no `head.sha` and no `head.ref`. Slash-command flows like `/deploy` end up with two annoying workarounds: stash the SHA in Canvas Memory during an `onPullRequest` event and read it back, or `git fetch origin pull/<n>/head` to dodge the missing branch name.

## Fix

Right after the signature/event/action/content-filter checks pass and we know we're going to emit, look the PR up via the GitHub API and attach the full object at `data["pull_request"]`. Downstream nodes can now read `data.pull_request.head.sha`, `data.pull_request.head.ref`, etc. directly.

Best-effort by design: if there's no integration on the request, no PR number on the payload, the client can't be built, the API call errors, or the marshal fails — we log a warning and emit the original event. Never drop the webhook silently.

A few small things worth calling out:

- The enriched data goes to a new top-level key (`data.pull_request`) instead of overwriting the existing minimal `data.issue.pull_request`. Keeps existing expressions like `data.issue.pull_request.html_url` working.
- 7s timeout on the API call so we don't get anywhere near GitHub's ~10s webhook delivery deadline. Going past that triggers retries and duplicate events.
- Added `GetPullRequest` to the common client; placed it next to `CreatePullRequest` to match the existing grouping.

## Tests

Two new tests on top of the existing ones:

1. Successful enrichment — mock GitHub API returns a PR, assert `data.pull_request.head.sha` / `head.ref` / `base.sha` / `base.ref` show up on the emitted payload.
2. Degraded mode — API returns 500, assert the event still emits and `data.pull_request` is absent.

Existing tests (which don't pass an `Integration`) keep passing because of the `ctx.Integration == nil` early-out. `go test ./pkg/integrations/github/...` is green.

### DIFF
--- a/pkg/integrations/github/common/client.go
+++ b/pkg/integrations/github/common/client.go
@@ -168,6 +168,10 @@ func (c *Client) CreatePullRequest(ctx context.Context, repository string, pullR
 	return c.underlying.PullRequests.Create(ctx, c.owner, repository, pullRequest)
 }
 
+func (c *Client) GetPullRequest(ctx context.Context, repository string, pullNumber int) (*github.PullRequest, *github.Response, error) {
+	return c.underlying.PullRequests.Get(ctx, c.owner, repository, pullNumber)
+}
+
 func (c *Client) CreateStatus(ctx context.Context, repository string, sha string, status github.RepoStatus) (*github.RepoStatus, *github.Response, error) {
 	return c.underlying.Repositories.CreateStatus(ctx, c.owner, repository, sha, status)
 }
--- a/pkg/integrations/github/components/pulls/common.go
+++ b/pkg/integrations/github/components/pulls/common.go
@@ -1,16 +1,23 @@
 package pulls
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"regexp"
+	"time"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/superplanehq/superplane/pkg/configuration"
 	"github.com/superplanehq/superplane/pkg/core"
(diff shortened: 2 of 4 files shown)


markoa:

PR #4708, opened 2026-05-09
docs(agent): add canvas-memory skill and memory() to expressions

### LINKED ISSUE #4200: Expression memory() is not part of the autocomplete
While writing expressions in the UI, the `memory()` function is not auto-completed.

### LINKED ISSUE #4351: Skills: missing Canvas Memory, DigitalOcean reference, and incorrect guidance
### 1. Canvas Memory components not in skills

`addMemory`, `readMemory`, `upsertMemory`, `deleteMemory` are not mentioned in `superplane-canvas-builder` or `superplane-cli` skills. The `components-and-triggers.md` reference lists HTTP, Filter, If, Approval, Merge, Wait, SSH, Noop — but no memory components.

Canvas Memory is critical for stateful workflows (ephemeral envs, incident tracking, counters). An agent relying only on skills will miss it entirely and build HTTP workarounds.

### 2. No DigitalOcean reference file

The GitHub integration has a dedicated `references/github.md` with triggers, payload examples, output channels, and gotchas. DigitalOcean has nothing. Agent must discover all DO components, field formats (fingerprint vs name vs slug), and payload structures via trial and error.

### 3. `integration-resource` field format undocumented

No consistent guidance on when to use name vs ID vs fingerprint for `integration-resource` fields:
- SSH keys: wants fingerprint (`06:fe:...`)
- Regions: wants slug (`nyc3`)
- Apps/Droplets: wants name or ID depending on context
- `useNameAsValue` flag exists but is not explained in skills

### 4. Skills say "always use `--draft`"

Both `superplane-cli` and `superplane-canvas-builder` skills instruct: "every `superplane canvases update` command should include `--draft`". When change management is disabled (the default for new orgs), `--draft` creates a version that does NOT go live. Agents update the canvas, verify it looks correct, but the live canvas still runs the old version.

**Fix:** Skills should say "use `--draft` when change management is enabled; omit it when disabled."

### 5. `wait` component config mismatch

`components-and-triggers.md` documents: `type: duration/timestamp`. Actual CLI reports: `mode: interval/countdown` with `waitFor` and `unit` fields. Different field names AND different values.

## Context

All discovered during Mission 2 (ephemeral environments). Each gap added 15-30 minutes of wasted trial-and-error time.

CristianMaldonado24:
- Add agent/skills/canvas-memory.md documenting addMemory, readMemory, upsertMemory, and deleteMemory components with YAML examples and channel behaviour (found/not_found).
- Add memory(key) row to the SuperPlane functions table in expressions.md with a cross-reference to the new canvas-memory skill.

Closes #4351 (partial) - documents missing Canvas Memory components Closes #4200 (partial) - memory() now appears in the expression reference

### DIFF
--- a/agent/skills/canvas-memory.md
+++ b/agent/skills/canvas-memory.md
@@ -1,0 +1,56 @@
+# Canvas Memory
+
+Canvas Memory provides persistent key-value storage scoped to the canvas. Use it to share state across nodes, accumulate values across runs, or cache expensive lookups.
+
+## Components
+
+### addMemory
+
+Stores a new value for a key. Fails if the key already exists — use `upsertMemory` if you want create-or-update semantics.
+
+```yaml
+type: addMemory
+key: "{{ $['Trigger'].run_id }}"
+value: "{{ $['Trigger'].payload | toJSON }}"
+```
+
+### readMemory
+
+Reads the current value for a key. Emits on the `found` channel if the key exists, `not_found` otherwise. You can also reference memory values inline using the `memory()` expression function (see [expressions.md](expressions.md)).
+
+```yaml
+type: readMemory
+key: "last_deployed_sha"
+```
+
+Channels: `found`, `not_found`
+
+### upsertMemory
+
+Creates the key if it does not exist; updates it if it does.
+
+```yaml
+type: upsertMemory
+key: "deploy_count"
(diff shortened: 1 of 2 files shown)


markoa:

PR #4716, opened 2026-05-10
fix: add password change flow for email/password accounts

### LINKED ISSUE #4650: No way to change password for email/password accounts
If a user registered with email/password, there's no way to change or update their password.

No settings page, no password reset flow, nothing. You're stuck with whatever you signed up with.

erik-kroon:
## Summary

Fixes #4650 by adding a self-service password change flow for accounts that use email/password authentication.

## Core changes

- Adds `PATCH /account/password` for account-session password updates.
- Requires the current password before writing a new password hash.
- Blocks password changes while impersonating another account.
- Exposes `has_password_auth` on `/account` so the UI only shows the password form for eligible accounts.
- Adds a Profile settings password section that matches the existing settings page style.
- Adds backend regression coverage for unauthenticated, non-password, wrong-password, and successful password-change cases.

## Behavior / invariants

- Password changes are available only to accounts with password auth records.
- The current password must verify before the stored hash is replaced.
- Impersonation sessions cannot change another account password.

## Validation

- `make format.js`
- `npm run build`
- `npx eslint src/pages/organization/settings/Profile.tsx src/contexts/AccountContext.tsx`
- `go test ./pkg/public -run '^$'`
- `go test ./pkg/models -run '^$'`
- `go build cmd/server/main.go`

### DIFF
--- a/pkg/models/account.go
+++ b/pkg/models/account.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"errors"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


markoa:

PR #4722, opened 2026-05-10
fix: validate UUIDs and fix typo in RemoveUser

fegzycole:
## Summary

  - Add `uuid.Parse` validation for `orgID` and `userID` in `RemoveUser`, returning `codes.InvalidArgument` on bad input — consistent with the pattern used
  throughout the `organizations` package (e.g. `update_integration_property.go`, `list_integration_resources.go`)
  - Fix typo in the gRPC error response: `"error determing user roles"` → `"error determining user roles"` (this was user-visible via the status message)
  - Add test cases covering both invalid-UUID inputs

  ## Test plan

  - [ ] `make test PKG_TEST_PACKAGES=./pkg/grpc/actions/organizations/...` passes (110 tests)

### DIFF
--- a/pkg/grpc/actions/organizations/remove_user.go
+++ b/pkg/grpc/actions/organizations/remove_user.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"slices"
 
+	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


markoa:

PR #4729, opened 2026-05-11
fix: include soft-deleted canvases in retention cleanup

forestileao:
## Summary

Updates event retention cleanup to include expired root events from soft-deleted canvases and organizations.

The retention worker still uses the organization retention window to decide eligibility, but no longer skips events only because the canvas or organization is soft-deleted.


forestileao:
👍

PR #4735, opened 2026-05-11
feat: Add Experimental Features Management

### LINKED ISSUE #4028: Feature Flags
Hard to be fast without this.
The solution needs to:

- Work in SaaS
- Work in Self-Hosted
- Can be set in runtime without restarting the app
- Per-Organization

andrecalil:
## Summary
Introduced the concept of **Experimental Features** and the respective management UI.

- Experimental Features should be added to `pkg/features/features.go`, that's the single source of truth
- Management UI lives at `/admin/organizations/:orgId` (installation admin only). Org admins cannot toggle these from regular settings
- Per-organization opt-in is stored as a string list on `organizations.enabled_experimental_features`
- Once an experimental feature is "promoted", it can be tagged as `Released`, which will ignore whatever the organization has in the settings (effectively making it publicly available)

### Frontend
```
import { useExperimentalFeature } from "@/hooks/useExperimentalFeature";
...
const runnersAreAvailable = useExperimentalFeature("runners");
```  
### Backend
```
organization, err := models.FindOrganizationByID(organizationID)
...
runnersAreAvailable := organization.HasExperimentalFeature("runners")
// or, when you only have an org id and want to skip the DB read for released features:
runnersAreAvailable, err := models.HasExperimentalFeature(orgID, "runners")
```

This resolves #4028.


db/structure.sql
```diff
@@ -1520,6 +1755,76 @@ CREATE INDEX idx_workflows_live_version_id ON public.workflows USING btree (live
...
+--
+
+CREATE INDEX uix_stage_executions_stage ON public.stage_executions USING btree (stage_id);
+
+
+--
+-- Name: uix_stages_canvas; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE INDEX uix_stages_canvas ON public.stages USING btree (canvas_id);
```
forestileao: I think your DB might be outdated and structure.sql was commited by mistake?

We don't have stages since a long time

Run `make db.migrate.all` to regenerate

---

This is why https://superplanehq.semaphoreci.com/jobs/41dac602-377b-4366-9e72-583a7d3a3467 is failing, its not e2e

web_src/src/lib/features.ts
```diff
@@ -0,0 +1,8 @@
+import type { OrganizationsOrganization } from "@/api-client";
+
+export function hasExperimentalFeature(
+  organization: OrganizationsOrganization | null | undefined,
+  featureId: string,
+): boolean {
+  return organization?.spec?.enabledExperimentalFeatures?.includes(featureId) ?? false;
```
forestileao: this helper only checks the org’s stored enabledExperimentalFeatures list, but the backend treats features marked as Released as enabled for every org: [pkg/models/organization.go#L551-L563](https://github.com/superplanehq/superplane/blob/b6282bca999b72ea45a116dc38620a446c62019b/pkg/models/organization.go#L551-L563).

And the organization response here is also populated from the raw DB field rather than the effective feature set: [pkg/grpc/actions/organizations/describe_organization.go#L36-L39](https://github.com/superplanehq/superplane/blob/b6282bca999b72ea45a116dc38620a446c62019b/pkg/grpc/actions/organizations/describe_organization.go#L36-L39).

So once a feature is promoted to Released, backend callers will see it as enabled, while useExperimentalFeature() on the frontend will still return false for orgs that never explicitly opted in. That seems to break the advertised "released = publicly available" behavior.

db/structure.sql
andrecalil: I did ran it and pushed, that was my latest commit. If I run it now, I get nothing changed/created.
I'm considering removing all my containers and running setup from the start.

pkg/models/organization.go
```diff
@@ -472,6 +485,84 @@ func ListOrganizationsPendingUsageLimitsRefreshInTransaction(
...
+	now := time.Now()
+	return tx.
+		Model(&Organization{}).
+		Where("id = ?", orgID).
+		Updates(map[string]any{
+			"enabled_experimental_features": updated,
+			"updated_at":                    &now,
+		}).
+		Error
+}
```
forestileao, line 520: This is doing a read-modify-write of the full enabled_experimental_features JSONB array without any locking or atomic DB update. If two requests enable different flags at the same time, both can read the same starting value and the later write will clobber the earlier one.

The same pattern exists in disable as well: [pkg/models/organization.go#L523-L548](https://github.com/superplanehq/superplane/blob/b6282bca999b72ea45a116dc38620a446c62019b/pkg/models/organization.go#L523-L548).

Since the admin UI sends independent toggle requests per feature: [web_src/src/pages/admin/OrgExperimentalFeaturesTable.tsx#L46-L68](https://github.com/superplanehq/superplane/blob/b6282bca999b72ea45a116dc38620a446c62019b/web_src/src/pages/admin/OrgExperimentalFeaturesTable.tsx#L46-L68), this looks like it can lose feature IDs in practice. can we make this atomic or lock the row before updating?

db/migrations/20260511133726_add-experimental-features-to-organizations.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE organizations
+    ADD COLUMN enabled_experimental_features JSONB NOT NULL DEFAULT '[]'::jsonb;
```
forestileao, line 4: Does it make sense to call it `enabled_experimental_features` + having `released` property inside each feature? Maybe  we should call it just `experimental_features`.

What do you think?

db/structure.sql
forestileao: @andrecalil check if test db didn't messed up if an AI agent tried to test something manually.

Sometimes we just need to:

`make db.delete DB_NAME=superplane_test`
`make db.create DB_NAME=superplane_test`
`make db.migrate DB_NAME=superplane_test`

db/migrations/20260511133726_add-experimental-features-to-organizations.up.sql
andrecalil: It would be simpler, but here's the thing: suppose we build a complex experimental feature. It has multiple checks at the FE and at the BE.
If we wanted to _quickly_ release it, we could simply make a release marking it as `released`, then add a follow up ticket to clean up whenever it was used.
This mimics what feature flag engines offer as "default value". When a given feature is ready to be used, you set `default value = true` and follow up removing its usage in the codebase.

It's not a strong requirement and we can definitely remove it if it seems like too much.

forestileao: Makes sense

pkg/models/organization.go
andrecalil: I've added a transaction for each change, this should ensure the row get locked. Also added tests to validate it. Please give it another go.

web_src/src/lib/features.ts
andrecalil: I've reworked the hooks to simplify a bit and ensure this logic is also implemented in the FE. Please give it another go.

db/structure.sql
andrecalil: These commands helped. I think the pipeline is gonna clear now.

forestileao (approved):

PR #4737, opened 2026-05-11
fix: poll integration redirect setup state

### LINKED ISSUE #4666: Integration setup: redirect prompt step should poll integration setup state
Right now, the GitHub integration setup page can get stale during the GitHub App creation / installation flow. Since the flow starts with a redirect on a new tab, the whole will finish in the new tab, and the old tab will remain in the old state until a refresh.

Two ways to consider here:
- Poll integration setup state when on a redirect prompt step
- Do not use another tab for redirects

officialasishkumar:
## Summary
- Closes #4666
- Poll describe integration while setup is on a redirect prompt step.
- Apply polled setup state to the existing setup controller so completed redirects advance without a manual refresh.
- Add coverage for redirect prompt polling through completion navigation.

## Validation
- `npm exec prettier -- --write src/hooks/useIntegrations.ts src/pages/organization/settings/components/IntegrationSetup/useIntegrationSetupController.ts src/pages/organization/settings/components/IntegrationSetup/PreCreateIntegrationSetup.spec.tsx`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/web_src/src/hooks/useIntegrations.ts
+++ b/web_src/src/hooks/useIntegrations.ts
@@ -84,8 +84,15 @@ export const useConnectedIntegrations = (organizationId: string, options?: { ena
   });
 };
 
+type UseIntegrationOptions = {
+  enabled?: boolean;
+  refetchInterval?: number | false;
+  refetchIntervalInBackground?: boolean;
+  staleTime?: number;
+};
+
 // Hook to fetch a single integration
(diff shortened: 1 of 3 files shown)


shiroyasha:
Interesting! 

@lucaspin is this going in the direction that was described in the issue, and is the code still up to date so we can use it in the project?

PR #4742, opened 2026-05-12
feat: add parallel fan-in pattern for merge discoverability

### LINKED ISSUE #4294: Discoverability: merge/fork-join patterns

Vukotije:
## What

- Adds `agent/patterns/parallel-fan-in-merge.md`: a decision pattern teaching the canvas agent to use the `merge` component when a user wants two or more steps to run in parallel and converge before downstream work proceeds.
- Adds an eval case `parallel_branches_converge_with_merge` to `agent/evals/cases.py` to lock the topology in (uses existing evaluators; no new evaluator code).
- Fixes a wrong channel name in `agent/skills/data-flow.md`: the table listed `Merge | success, stopped, timeout`, but the real channels are `success, fail, timeout` (see `pkg/components/merge/merge.go` lines 22-26). Any subscription on `stopped` would silently never fire.

## Why

Closes #4294. The agent had no pattern matching "two branches in parallel that converge", so it defaulted to a sequential chain and missed `merge` until a human pointed to it. The decision-pattern infrastructure was already wired in (`search_decision_patterns` is registered at `agent/src/ai/tools/registry.py:50`), so the only gap was the absence of a `merge` recipe. The skill correction is an incidental but real bug fix that compounds the discoverability problem when the agent does reach for `merge`.

## How

The new pattern follows the format of the two existing patterns (`ephemeral-preview-machines.md`, `issue-auto-resolve-agent-label.md`): no frontmatter, a `Keywords:` line tuned to natural-language phrasings ("parallel", "fan-in", "wait for both", "converge", etc.), a numbered decision checklist, a canonical workflow with a fan-out diagram, and short notes covering the optional `enableTimeout` and `enableStopIf` knobs. Density matches the existing patterns (22 content lines vs 21/28 for the existing two).

The eval uses only existing evaluators:

```python
evals.CanvasHasNode("http", count=2),
evals.CanvasHasNode("merge", count=1),
evals.CanvasHasWorkflow("http", "...", "merge"),
```


forestileao (approved):

forestileao:
@shiroyasha Will need your approval on this one too

PR #4743, opened 2026-05-12
feat: CLI manual run and replay event

### LINKED ISSUE #4298: CLI: programmatic run and event replay for canvases

re-pixel:
This PR defines CLI commands for running manual triggers and replaying events.

```
superplane canvases run <canvas-name-or-id> <start-node-id> --template <template-name> [--payload <json-or-path-to-file>]
superplane canvases run <canvas-name-or-id> <start-node-id> --replay <event-id>
```

Optional --payload flag lets you override the template.

### Related Issues

Closes https://github.com/superplanehq/superplane/issues/4298

### Further improvements (UX)

Users would probably want to use --replay by just using the event id and omitting the --trigger flag, this isn't supported ATM. Also, currently --template is required even when using --payload override. Fixing this would require workarounds or changes in the API which are not in the scope for this PR.


pkg/cli/commands/canvases/root.go
```diff
@@ -244,13 +244,45 @@ AI agents: for canonical canvas YAML shapes and wiring rules, install skills:
...
+		Short: "Run a manual trigger or replay a trigger root event",
+		Long: `Start a Manual Run trigger via its "run" hook, or re-emit an existing root trigger event.
+
+Manual run invokes the trigger hook with the same parameters as the UI (template name and optional JSON payload override).
+
+Examples:
+  superplane canvases run my-canvas --trigger start-node --template "Hello World"
+  superplane canvases run my-canvas --trigger start-node --template "Hello World" --payload '{"message":"one-off"}'
+  superplane canvases run my-canvas --trigger start-node --template "Hello World" --payload ./payload.json
+  superplane canvases run my-canvas --trigger start-node --template "Hello World" --payload @./payload.json
```
shiroyasha: I don't like that this says `--trigger`. If feels like you can use any trigger, while this is only meant for manual runs. I'm pretty sure this will screw up agents as well.

@re-pixel @forestileao what do you think about:

```
superplane canvases run my-canvas start-node --template "Hello World"
```

or more generic:

```
superplane canvases run [canvas] [node-id] --template [template-name]
```

shiroyasha:
The code itself is good 👍 
Left a comment about usability.

pkg/cli/commands/canvases/root.go
re-pixel: i agree, will change it

bender-rodriguez-unit1:

PR #4754, opened 2026-05-13
fix: use component-defined color for GraphQL mapper

### LINKED ISSUE #4483: GraphQL component color is darker than the rest of the core components

Lucky-Rathore:
## What changed
Changed the GraphQL mapper to use the component's defined color instead of hardcoded black.

## Why
The GraphQL component was appearing darker than other core components because the mapper was using `getColorClass("black")` instead of using the component's defined color (`violet`).

## How
Updated `web_src/src/pages/workflowv2/mappers/graphql.ts` to use `getColorClass(context.componentDefinition.color)` which matches the pattern used by other integration components.

Closes #4483


shiroyasha:
Thank you for the PR, and sorry for the late review!

Unfortunately, this change would still make the component violet and not in tone with the rest of the core components. I'll close this PR for now.

Issue #2811, opened 2026-02-03
HTTP component: paste curl to prefill config


AleksandarCole:
**Problem:** The Core – HTTP component lets users configure the request (URL, headers, body, query params, etc.) through a multi-step or longer form. Configuring a request from scratch or from an existing curl command is tedious. It would be helpful to have a helper in the config UI where the user can paste a curl command and we prefill URL, headers, body, and query params where possible.

**Current vs proposed:**
- **Current:** User fills URL, headers, body, query params manually in the config form.
- **Proposed:** A helper UI (e.g. "Paste curl" or "Import from curl") that accepts a pasted curl string, parses it, and prefills the config fields. Validation is required; we won't support every curl variant, so the UI should handle parse failures and partial matches (e.g. "We filled what we could; please review"). This is a non-trivial UI/parsing feature.

**Use case:** Users who have a curl command (e.g. from docs, browser DevTools, or Postman "Copy as cURL") can quickly bootstrap the HTTP component config instead of retyping URL, headers, and body. Speeds up setup and reduces errors.

___
**Screenshot:**

<img width="850" height="748" alt="Image" src="https://github.com/user-attachments/assets/68f40642-4933-4e82-a1f8-b4ade566514f" />

PrashantIndurkar:
Hey @AleksandarCole

I’d like to work on this issue if it’s okay.

Before I start, I had a couple of questions about the UX:

1. For pasting the cURL command, should we:

* Add a **separate input / modal** where users paste the full curl, or
* Reuse the existing **URL field** and detect when a user pastes a `curl ...` command (like Postman)?

2. If we add a separate input, where would you prefer it?

* After the URL field, or
* At the top of the form (before Method/URL)?

Let me know your preference, and I’ll start working on it. Thanks! 🙌

AleksandarCole:
@PrashantIndurkar thanks for taking this one, feel free to give it a go!

We should keep the current configuration form but above the `Method` we can add a new input/textarea. When user pastes/edits curl the form bellow can be updated and some validation can show below the input with details if we failed to parse some parts.

AgustinRamiroDiaz:
@AleksandarCole I think @PrashantIndurkar was not able to work on this
Can I take it?

NanaWhan:
Hey @AleksandarCole , if no one has started on this yet, happy to pick it up. I've already mapped out the implementation.

PR #4761, opened 2026-05-13
feat: prefill HTTP component fields from a pasted curl command

jonatasrancan:
Implements: #2811

## What changed:

A new "Prefill from a curl command" section renders above the HTTP component's configuration form in the sidebar. Pasting a curl command and pressing Enter (or clicking "Parse & fill") populates method, URL, headers, query params, content type, and body in one shot. The authorization scheme (Bearer, Basic, custom header) is detected too; secret references are intentionally left blank for the user to wire up. Re-pasting a different curl wipes any fields the new command doesn't supply, so successive pastes don't leave stale state.

## Why:

The HTTP component is the most common way for users to call an external API from a canvas. API vendors distribute their examples as curl commands, so users currently retype every field by hand. Letting them paste the curl and prefill the form removes that friction and matches what every API doc page is set up to copy.

### Parser utilities

- `web_src/src/lib/parseCurl.ts`: tokenizer-based parser that splits a curl command into URL, method, headers, authorization, body, and query params. URL extraction follows curl's own argument-parsing rule -- anything starting with `-`/`--` is an option (with a known list of value-taking flags), the first remaining positional token is the URL. This supports bare hostnames, IPs, `localhost:8080`, custom schemes, etc., not just http(s)://. The tokenizer respects single quotes (literal), double quotes (backslash-escaped), and adjacent quoted/unquoted segments (`--data="hello"`). A small `splitOnce(s, sep)` helper dedupes the five separator-split sites.
- `web_src/src/lib/curlToHttpConfig.ts`: translates the ParsedCurl into an HTTPConfigurationPatch. Infers contentType from the Content-Type header or the curl body flag (`--json`, `--data-urlencode`), routes the body into the right field (json / formData / text / xml), and maps `-u` / `--oauth2-bearer` / Authorization header onto authorization.type. Every curl-owned field is initialised to `undefined` so re-parsing clears prior prefill via the existing filterVisibleConfiguration merge in SettingsTab.

### CustomField registry

Component-specific sidebar UI plugs in through the existing per- component `customFieldRenderers` registry, with two small contract additions:

- `web_src/src/pages/workflowv2/mappers/types.ts`:
  - `CustomFieldRendererContext.applyConfigurationPatch` lets the rendered UI push a partial config patch back into the form.
  - `CustomFieldRenderer.position?: "before" | "after"` controls placement relative to the configuration fields. Existing renderers default to "after".
  - `SettingsCustomField { position, render }` is the shape returned by `getCustomField` for the sidebar.
- `web_src/src/pages/workflowv2/lib/render-canvas-node-custom-field.ts` and `pages/workflowv2/index.tsx`: thread the apply callback through and return `{ position, render }` from `getCustomField`.
- `web_src/src/pages/workflowv2/mappers/safeMappers.ts`: the wrapper preserves `renderer.position`.
- `web_src/src/pages/workflowv2/mappers/http.ts`: new `httpCustomFieldRenderer` with `position: "before"` returns the CurlImportSection bound to the apply callback.
- `web_src/src/pages/workflowv2/mappers/index.ts`: registers `http` in `customFieldRenderers`.

### UI

- `web_src/src/pages/workflowv2/mappers/http/CurlImportSection.tsx`: always-expanded panel with a textarea, "Parse & fill" button, and inline error/hint slots. Enter parses; Shift+Enter inserts a newline. Surfaces a hint when authorization is detected but the credential field still needs a secret reference.
- `web_src/src/ui/componentSidebar/SettingsTab.tsx`: builds a single `applyConfigurationPatch` callback and renders the customField in either the "before" slot (above the fields) or the existing "after" slot, based on the renderer's `position`.

https://github.com/user-attachments/assets/c6121d1a-e447-4026-862f-5cb1cc996cce


web_src/src/ui/componentSidebar/SettingsTab.tsx
```diff
@@ -694,6 +704,9 @@ export function SettingsTab({
           <div
             className={`border-t border-gray-200 dark:border-gray-700 pt-6 space-y-4 ${isReadOnly ? "pointer-events-none opacity-60" : ""}`}
           >
+            {customField?.position === "before" && !isReadOnly && (
+              <div>{customField.render(nodeConfiguration, applyConfigurationPatch)}</div>
+            )}
```
jonatasrancan, line 709: My first approach was to have ` web_src/src/pages/workflowv2/mappers/http/CurlImportSection.tsx`  rendered direct here, and then check which component was been rendered to decide to show or not the `CulrImportSection`.

It worked fine, and was less changes than this solution, but it was introducing very custom logic that belongs to the ` http` component and kind of made ` SettingsTab` aware of  the component it was rendering

I did some more research, and then found out this ` customField`  approach, and kind of make sense to use it, had to change more places and add more stuff, but in the end makes more sense, this way we can reuse it more easily and also add this `position` property to decide where to add them

But let me know, if the previous version was ok, or if there is another way to do this

shiroyasha:
Unfortunately, we haven't reviewed it on time.

Since we have introduces runners, users can now directly use the curl command if they want to to. I’ll close this PR.

PR #4765, opened 2026-05-13
feat: split login/signup pages and add new-user survey

### LINKED ISSUE #4724: Auth UX improvements: Separate login/signup and new-user survey
## Problem

The current cloud auth flow uses a single `/login` page for both existing and new users. Because the same passwordless flow can either sign a user in or create a new account, there is no clear distinction between logging in and signing up from the user's perspective.

This came up while looking into how to trigger a post-signup survey.
## Potential solution

Split `/login` and `/signup` into two distinct frontend pages, while keeping the same passwordless magic-code flow on both. Use the backend's account creation signal (`wasCreated`) to trigger a post-signup survey only when a new account is actually created — regardless of which page the user came from.

Full spec: [[link](https://gist.github.com/gaga1307/31b3fd74b3c3d0df07f9c36af0f4363e)]

gaga1307:
Closes #4724

## Summary

Today the sign-in page is the only entry point regardless of whether a user already has an account, which makes it unclear where to start as a new user. After this PR:

- The auth flow now uses separate /signup and /login pages, following the standard split-page UX commonly used in SaaS products.
- Genuinely new users (magic-code or OAuth) are routed through a `/welcome` page that renders a PostHog onboarding survey if one is active. Existing users are never sent through this page.

### Backend (`pkg/authentication/authentication.go`)
- `findOrCreateAccountForMagicCode` and `findOrCreateAccountForProvider` now return a `wasCreated bool` so the caller can tell genuinely-new accounts apart from existing ones.
- `wasCreated` is threaded through `handleAuth`, `handleDevAuth`, `handleAuthCallback`, `handleMagicCodeVerify`, `handleSuccessfulAuth`, and `issueSessionAndRedirect`.
- New helper `getPostAuthRedirectURL` returns `/welcome?redirect=<original>` only when `magicCodeEnabled && wasCreated`; otherwise returns the original redirect URL.

### Frontend routing (`web_src/src/App.tsx`)
- Added `<Route path="signup" element={<Login mode="signup" />} />`.
- Added `<Route path="welcome" element={withAuthOnly(WelcomeSurvey)} />`.

### Login page (`web_src/src/pages/auth/Login.tsx`)
- New `mode?: "login" | "signup"` prop driven by the route. A `useEffect` syncs `isSignupMode` from `mode` **only when `magicCodeEnabled` is true** and resets `magicCodeStep`, `magicCode`, and `showPasswordLogin` on transitions.
- New `CardBody` wrapper applies a subtle fade/slide animation between `/login` and `/signup`.
- `<Link>`-based cross-navigation between `/login` and `/signup`.
- `handleRedirectAfterAuth` prioritizes backend `/welcome` redirects.

### Generic survey renderer
- Renamed `OwnerSetupSurvey.tsx` → `PostHogSurveyForm.tsx`. The component is now a generic PostHog survey form.

### New welcome page (`web_src/src/pages/auth/WelcomeSurvey.tsx`)
- Reads `?redirect=...` from the URL with the same safe-path validation as `Login.tsx` (plus an explicit exclusion of `/welcome`).
- Uses `posthog.onSurveysLoaded` + `getActiveMatchingSurveys(..., true)` to find a survey named `New User Onboarding Survey`.
- If PostHog is disabled, no matching survey exists, or PostHog hasn't been configured: renders `<Navigate to={redirectTo} replace />` immediately — no blank screens, no loading text, no flash.
- If a survey is found: renders `PostHogSurveyForm`.

## Test plan

Tested with the following config in docker-compose.dev.yml:

```yaml
ENABLE_PASSWORD_LOGIN: "no"
ENABLE_MAGIC_CODE_LOGIN: "yes"
BLOCK_SIGNUP: "no"
```
With POSTHOG_KEY set in a separate .env file. Tested both with an active PostHog survey named New User Onboarding Survey and without one (no matching survey).

- Demo: [video](https://drive.google.com/drive/folders/1vPG-DIOVTayGdkXjRfWnIqSqFASAzKpp?usp=sharing)


web_src/src/pages/auth/Login.tsx
```diff
@@ -191,12 +210,34 @@ export const Login: React.FC = () => {
...
   const handleToggleMode = (nextMode: "login" | "signup") => {
     setIsSignupMode(nextMode === "signup");
     setFormError(null);
   };
 
   const handleRedirectAfterAuth = async (response: Response) => {
+    const finalURL = response.url || "/";
+    try {
+      const parsedURL = new URL(finalURL, window.location.origin);
+      if (parsedURL.origin === window.location.origin && parsedURL.pathname === "/welcome") {
```
forestileao, line 233: This makes `/welcome` win over `redirectTarget`, but the magic-link verify path posts to `/auth/magic-code/verify` without forwarding the current `redirect` query. So a user who signs up from a magic link that had `redirect=/invite/...` gets `/welcome` with no `redirect` and finishes at `/` instead of the original target. Please include the redirect when verifying magic links, or append `redirectTarget` here when `finalURL` is `/welcome` without one.

web_src/src/pages/auth/WelcomeSurvey.tsx
```diff
@@ -0,0 +1,96 @@
...
+      canceled = true;
+      unsubscribe();
+    };
+  }, []);
+
+  if (shouldRedirect) {
+    return <Navigate to={redirectTo} replace />;
+  }
+
+  if (!survey) {
```
forestileao, line 83: With PostHog enabled, this can render a blank page forever if the surveys callback never fires, for example because PostHog is blocked or the request fails. `shouldRedirect` only starts true when PostHog is disabled, and there is no timeout or error fallback. Please add a fallback that redirects to `redirectTo` when surveys do not load.

forestileao (commented):
Reviewed the auth flow changes. I left two comments for issues that can affect new-user redirects and the welcome page fallback.

PR #4769, opened 2026-05-13
feat: surface unpublished-draft choice on canvas Edit button

forestileao:
## Summary

  When a canvas has an unpublished draft and the user is viewing the live version, clicking **Edit** now opens a small dropdown with two clear choices instead of
  silently dropping them into the existing draft:

  - **Continue editing** — opens the existing draft (previous default behavior).
  - **Discard draft & start fresh** — deletes the draft, then creates a new one from live and enters edit mode.

  The dropdown header shows "You have an unpublished draft" and a live "Last edited X ago" timestamp. When no unpublished draft exists, the Edit button behaves
  exactly as before (single click → edit).

  ## Why

  It's currently invisible that an existing draft is being reused when re-entering edit mode. Users who wanted to start over had to enter edit mode and then Discard
  from the version-edit toolbar, which is two extra steps and creates a moment where the previous draft's changes are loaded onto the canvas. Surfacing the choice at
   the Edit-button click makes the state explicit and gives a one-click "start fresh" path.

### DIFF
--- a/web_src/src/pages/workflowv2/index.tsx
+++ b/web_src/src/pages/workflowv2/index.tsx
@@ -5059,6 +5059,33 @@ export function WorkflowPageV2() {
     handleUseVersion(liveCanvasVersionId);
   }, [handleUseVersion, isRunsMode, isViewingLiveVersion, liveCanvasVersionId]);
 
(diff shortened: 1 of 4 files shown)


forestileao:
👍

PR #4770, opened 2026-05-13
feat: gate Runners behind an experimental flag

andrecalil:
## Summary
Gated Runners (`runner`) behind the newly introduced _experimental feature flag_.
This is handled at `web_src/src/hooks/useComponentData.ts`. The source of truth remains `pkg/features/features.go`.


shiroyasha:
Looks good 🚀

PR #4772, opened 2026-05-13
feat: allow signed-in users to change their password

### LINKED ISSUE #4650: No way to change password for email/password accounts

Vukotije:
Closes #4650

## Summary

Adds an in-app "Change password" flow for accounts that signed up with email + password. OAuth-only accounts see nothing — they manage credentials at their identity provider.

A successful change rotates the password hash, invalidates every other session belonging to the account, and revokes all API tokens for the account. The user stays signed in on the tab they used to perform the change.

## How session invalidation works

Since JWTs are stateless, and we can't delete them:

- A new nullable column `accounts.password_changed_at` records the most recent password rotation.
- Every account-scoped auth path (cookie session, scoped JWT token, impersonation) now compares the token's `iat` to `password_changed_at` via `Account.IsSessionFresh`. Tokens issued before the rotation are rejected with 401.
- API tokens (`users.token_hash`) aren't JWTs, so they're cleared in the same DB transaction as the password update.
- The current session would otherwise lock itself out, so the handler reissues a fresh `account_token` cookie in its response and clears any active impersonation cookie.

## Constraints respected

- Refuses the change when password login is disabled (deployments without local auth).
- Refuses the change while an impersonation session is active.
- All DB writes (password hash, `password_changed_at`, token-hash wipe) happen in a single `gorm` transaction.
- Follows the repo's transaction guidelines: no `database.Conn()` calls inside functions that already receive a `tx`.

## Test plan

- [ ] `make test PKG_TEST_PACKAGES=./pkg/public` — handler tests for `/account/password` (success, wrong current password, mismatch, too short, OAuth-only account, impersonation active, password login disabled).
- [ ] `make test PKG_TEST_PACKAGES=./pkg/public/middleware` — freshness checks for cookie + scoped-token auth.
- [ ] `make test PKG_TEST_PACKAGES=./pkg/models` — `IsSessionFresh`, `MarkPasswordChangedInTransaction`, `ClearTokenHashesForAccountInTransaction`.
- [ ] `make lint && make check.build.app` — backend.
- [ ] `make check.build.ui` — frontend.
- [ ] Manual: sign in as an email/password user → Profile page → "Change password" button visible → submit a valid change → toast appears, current tab stays signed in.
- [ ] Manual: open a second browser/incognito session before changing the password → after change, second session is signed out on next request.
- [ ] Manual: sign in via OAuth (no password set) → Profile page → no "Change password" button.
- [ ] Manual: with an existing API token, change the password → confirm the old API token is rejected (401).
- [ ] Manual: while impersonating another account, attempt the change → 403.


shiroyasha:
Looks good! @forestileao wdyt, ready for merge?

forestileao:
@shiroyasha Yep, tha's good, let me just update the migration timestamp and merge it

forestileao:
👍

PR #4776, opened 2026-05-13
fix: stop logging gRPC metadata in authorization interceptor

OlegGitH:
The unary authorization interceptor logged the full incoming gRPC metadata at ERROR level whenever the required identity headers were missing:

    log.Errorf("User not found in metadata, metadata %v", md)
    log.Errorf("Organization not found in metadata, metadata %v", md)

gRPC metadata.MD carries every header the caller sent, including the Authorization header (Bearer JWT or user API token), x-token-scopes, and any forwarded cookies. Any client that drifts (missing x-user-id / x-organization-id) writes its credentials into the log pipeline, where retention typically outlives the tokens themselves.

Log only the gRPC method name and which required header was missing. No behavior change otherwise; the same codes.NotFound response is returned.


markoa:

PR #4783, opened 2026-05-14
feat: update cloudflare to use account api tokens over user api tokens

### LINKED ISSUE #4759: Update cloudflare authentication method
Cloudflare recommends using account api tokens in favour of user api tokens. We currently support usage of user api tokens and this should be updated to use account api tokens which are also easier to configure with custom templates.

-------

> While user tokens act on behalf of a particular user and inherit a subset of that user's permissions, account API tokens allow you to set up durable integrations that can act as service principals with their own specific set of permissions. This approach is ideal for scenarios like CI/CD, or building integrations with external services like SIEMs where it is important that the integration continues working, even long after the user who configured the integration may have left your organization altogether. User tokens are better for ad hoc tasks like scripting, where acting as the user is ideal and durability is less of a concern.

--------

### References

- https://developers.cloudflare.com/fundamentals/api/get-started/account-owned-tokens/

WashingtonKK:
Closes:  https://github.com/superplanehq/superplane/issues/4759

This updates cloudflare integration docs to guide the user to use an account api token in favour of the user api tokens. Account API tokens are preferred over user api tokens for long running integrations.


AleksandarCole:
@shiroyasha as you did original integration - can you check and let us know if this approach is ok?

PR #4789, opened 2026-05-14
feat: Add Env Vars to Runners

andrecalil:
## WIP


pkg/components/runner/commands.go
```diff
@@ -25,3 +36,79 @@ func validateCommands(commands string) error {
...
+
+			value, err := secrets.GetKey(variable.Secret.Secret, variable.Secret.Key)
+			if err != nil {
+				return nil, fmt.Errorf("failed to resolve environment variable %s secret %s/%s: %w", name, variable.Secret.Secret, variable.Secret.Key, err)
+			}
+
+			resolved = append(resolved, BrokerEnvironmentVariable{
+				Name:  name,
+				Value: string(value),
+			})
```
shiroyasha, line 109: This is ok a temporary solution, but we will need to figure out a way to make this encrypted e2e down to the runner.

Lets create an issue for this to track progress.

Issue #4771, opened 2026-05-13
Navbar logo clickability navigates incorrectly


Vukotije:
https://github.com/user-attachments/assets/f4e7e91d-da6e-46c5-8dc6-de3bc7561cb3

**Description**

When interacting with the **SuperPlane logo**, there are a few navigation issues that result in confusing or broken behavior for the user.

**Observed behavior**

1. When you click the logo while you are already on the Canvases page, it is still clickable and appears to “navigate”, but you stay on the same page and nothing effectively changes for the user. This feels redundant and misleading; it probably should not be clickable in this state.  

2. When you click the logo from the items in the navbar, they relocate you back to the Canvases view instead of the expected destination. This is good.

3. If the user has no canvases to land on (for example, not logged into any organization), clicking the logo will lead to a 404.

**Expected behavior**

- When already on the Canvases page, clicking the logo should be disabled so it is clearly non-interactive.  
- If there is no valid canvas/org to route to, clicking the logo should fall back to a safe, non-404 page (e.g. a neutral home/dashboard or an empty-state screen).

shiroyasha:
No longer an issue.

PR #4791, opened 2026-05-14
fix: fixed navbar logo redirection from change organisation page 

hrishi82:
Issue: https://github.com/superplanehq/superplane/issues/4771

**What changed:** 
Persist the active organization id in sessionStorage, use it for the header logo when the route has no org (e.g. when user is on Change Organisation page), and clear it on sign out.

**Why:** 
After clicking on “Change organization” from the settings menu, users land on "/" route without organizationId in the URL, so the logo targeted "/undefined". We also need a “back to last org” behavior which this PR implements

**How:** 
OrganizationScope (and org list clicks) call setCurrentOrgIdToSessionStorage; 
OrganizationMenuButton resolves the logo with organizationId || getCurrentOrgIdFromSessionStorage(); 
sign out calls clearCurrentOrgIdFromSessionStorage() before /logout.

**Related issues:** 
None

**Breaking changes:** 
None.


markoa:

PR #4794, opened 2026-05-14
fix: avoid broken logo link when no organization is selected

### LINKED ISSUE #4771: Navbar logo clickability navigates incorrectly

marton987:
Closes #4771

- Adds logoTarget = organizationId ? /${organizationId} : "/".
- Replaces the unconditional <Link to={/${organizationId}}> with a branch: when organizationId is set, render the <Link>; otherwise render a static <span aria-current="page"> (no link).

Behavior change: when there is no org, the logo no longer links to /${undefined} — it renders as a non-clickable element.

### DIFF
--- a/web_src/src/components/OrganizationMenuButton.tsx
+++ b/web_src/src/components/OrganizationMenuButton.tsx
@@ -40,6 +40,7 @@ export function OrganizationMenuButton({ organizationId, className }: Organizati
   );
   const [isMenuOpen, setIsMenuOpen] = useState(false);
   const menuRef = useRef<HTMLDivElement | null>(null);
+  const logoTarget = organizationId ? `/${organizationId}` : "/";
 
   const handleMenuButtonClick = () => {
     setIsMenuOpen((prev) => !prev);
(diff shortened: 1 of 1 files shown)


shiroyasha:
Good catch! Running CI.

shiroyasha:
Thanks for opening this PR! It seems that another PR was already merged that addresses this issue. Closing this one.

shiroyasha:

It looks like the issue this PR was addressing has already been fixed, so this change is no longer needed.

PR #4824, opened 2026-05-16
feat: enable cancelling runners

### LINKED ISSUE #4696: Runners: Support stopping a run

re-pixel:
Resolves https://github.com/superplanehq/superplane/issues/4696

Demo: https://drive.google.com/file/d/1oxWe9d3HIlKPES7Uqq7AT8R_3AqcA7ID/view?usp=sharing

### DIFF
--- a/pkg/components/runner/broker.go
+++ b/pkg/components/runner/broker.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -16,6 +17,9 @@ import (
 
 const (
 	brokerHTTPTimeout = 30 * time.Second
+
+	cancel409MaxAttempts  = 3
+	cancel409RetryBackoff = 200 * time.Millisecond
 )
 
 type BrokerClient struct {
@@ -43,7 +47,7 @@ func NewBrokerClient(httpClient core.HTTPContext) (*BrokerClient, error) {
 
 	return &BrokerClient{
 		httpClient: httpClient,
-		baseURL:    baseURL,
+		baseURL:    strings.TrimRight(strings.TrimSpace(baseURL), "/"),
 		fleetID:    fleetID,
 		authToken:  authToken,
 	}, nil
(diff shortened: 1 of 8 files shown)


pkg/components/runner/broker.go
```diff
@@ -16,6 +17,9 @@ import (
 
 const (
 	brokerHTTPTimeout = 30 * time.Second
+
+	cancel409MaxAttempts  = 3
+	cancel409RetryBackoff = 200 * time.Millisecond
```
shiroyasha, line 23: Check this.

pkg/components/runner/broker.go
```diff
@@ -43,7 +47,7 @@ func NewBrokerClient(httpClient core.HTTPContext) (*BrokerClient, error) {
 
 	return &BrokerClient{
 		httpClient: httpClient,
-		baseURL:    baseURL,
+		baseURL:    strings.TrimRight(strings.TrimSpace(baseURL), "/"),
```
shiroyasha: Check if this is necessary or llms are overly cautious.

pkg/components/runner/broker.go
```diff
@@ -184,14 +188,101 @@ func (t *Task) effectiveExitCode() int {
 }
 
 func (t *Task) IsInTerminalState() bool {
-	return t.Status == "succeeded" || t.Status == "failed"
+	switch strings.ToLower(strings.TrimSpace(t.Status)) {
+	case "succeeded", "failed", "canceled", "cancelled":
```
shiroyasha: Chose one spelling (US).

pkg/components/runner/broker.go
re-pixel: Task broker returns 409 when the task is not yet written in the database, so this is for safety in cases where the tasks are run and then immediately cancelled.

pkg/components/runner/broker.go
re-pixel: overcaution

pkg/components/runner/broker.go
re-pixel: canceled stays, cancelled removed

PR #4825, opened 2026-05-16
fix: validate 5-field cron ranges correctly

immanuwell:
Tiny frontend validation fix for schedule cron fields.

5-field cron was checked like it had a seconds column, so ranges shifted left. kinda sneaky.

Repro:
- open a cron configuration field
- submit `0 9 31 * *`
- before: rejected as an invalid hour, even tho day 31 is valid
- submit `0 24 * * *`
- before: accepted, even tho hour 24 is not valid

Now 5-field cron gets mapped onto the same internal positions as 6-field cron before range checks.

Tests:
- `npm run test:run -- src/lib/components.spec.ts src/lib/cron.spec.ts`
- `make format.js`

Related: searched issues/PRs for cron schedule validation, no direct match found.


shiroyasha (approved):

shiroyasha:
Looks good! Thank you for opening this PR, and congrats on your first merged PR.

PR #4834, opened 2026-05-17
fix: preserve zoom when switching to runs view

suryaravula1:
## Summary
- preserve the current viewport when switching between canvas and runs view
- skip the initial runs fit-all when reusing the existing viewport
- add unit tests for the viewport sync helper

## Verification
- npm run test:run -- src/pages/workflowv2/lib/runs-viewport.spec.ts

## Notes
- I also attempted a full frontend build, but this fresh clone is missing generated api-client files, so the build fails repo-wide before reaching this change.


shiroyasha:

PR #4843, opened 2026-05-17
fix: show runner commands in logs

### LINKED ISSUE #4804: Runner: Show which command was executed in the logs
Display the executed command before its output in the logs, so it's clear which step produced which output.

Itz-snj:
Closes #4804

## Summary
- print each runner command before its output in task logs
- preserve the original command execution while shell-quoting the displayed command header
- cover the broker payload and quoting behavior with runner tests

## Tests
- go test ./pkg/components/runner

### DIFF
--- a/pkg/components/runner/commands.go
+++ b/pkg/components/runner/commands.go
@@ -29,6 +29,22 @@ func normalizeCommands(commands string) []string {
 	return out
 }
 
+func commandsWithLogHeaders(commands []string) []string {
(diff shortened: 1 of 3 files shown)


shiroyasha:
Thank you for opening this PR! It is an interesting approach.

We will take a different approach and expose them on the runners. 
It is a cleaner approach.

I'll close this PR as it will no longer be necessary.

PR #4848, opened 2026-05-18
fix: preserve runs view zoom

### LINKED ISSUE #4800: Switching to runs view resets zoom level
## Description

Whenever switching to the runs view, the canvas zooms in unexpectedly. The zoom level should be preserved from the previous view.

## Steps to Reproduce

1. Set a specific zoom level on the canvas
2. Switch to the runs view
3. Observe the zoom level changes (zooms in)

## Expected Behavior

The zoom level should remain the same when switching between views.

## Actual Behavior

The canvas zooms in when switching to the runs view, which is disruptive and requires manually adjusting the zoom back.

AssaadHalabi:
Fixes #4800.

## Summary
- Preserve the current canvas viewport when entering runs mode so zoom does not reset.
- Keep the runs viewport across mode toggles, and only refit when the selected run's node set changes.
- Add unit coverage for the runs viewport refit decision.

## Tests
- `git diff --check`
- Not run: `make format.js`, `make check.build.ui` (Docker engine and local npm/make are unavailable in this environment).


shiroyasha:

Issue #1668, opened 2026-01-15
Hitting backspace while editing an expression should not delete a node


markoa:
Solution: change the keyboard shortcut to Cmd (on Mac) / Ctrl (on Windows) + Backspace

PR #4855, opened 2026-05-18
fix: restore keyboard delete shortcut for canvas components

### LINKED ISSUE #3243: Deleting selected component via Del key on keyboard
## Problem
Currently, there is no clear way to quickly delete a selected component on the canvas using a keyboard shortcut (Del key). Users expect to be able to delete the selected component by pressing 'Delete' on their keyboard for efficiency, similar to most UI/diagram editors.

Some related issues exist (e.g., changed shortcut to Cmd/Ctrl+Backspace to avoid conflicts while editing expressions), but a standard, discoverable way to delete a component directly with the Delete key does not seem to be present.

## Proposed Solution
- Pressing the 'Delete' key on the keyboard should remove the currently selected component from the canvas.
- Should not trigger deletion if focus is inside a text input or when editing fields (to prevent accidental removal while typing).
- Should work on both Windows (Del key) and macOS (Delete key) keyboards.
- Should NOT affect deletion behavior within text input fields or code editors within the component/configuration.

## Additional notes
- Should check for conflicting shortcut behaviors (see #1668 for previous change involving backspace).
- Should ideally provide a confirmation if component has children or is linked to others.

---
**References:**
- [#1668](https://github.com/superplanehq/superplane/issues/1668) for context on past keyboard deletion/binding issues.
- [Issue search for related items](https://github.com/superplanehq/superplane/search?q=is%3Aissue+%28delete+OR+del+OR+backspace%29+%28component+OR+node+OR+block%29+%28keyboard+OR+shortcut%29&type=issues)

**Request:**
Implement the ability to remove selected components using the Del key, with appropriate context awareness and conflict avoidance.

jwl3g14:
Closes #3243

## Summary

Pressing `Delete` or `Backspace` with components selected on the canvas now removes them, restoring the UX requested in #3243.

Re-enables the keyboard shortcut that was disabled in #1717 (via `deleteKeyCode={null}`) to work around #1668, but does so with a focus gate so typing in inputs, textareas, contenteditable fields, or the Monaco expression editor still edits text rather than deleting components — the original #1668 concern remains addressed.

## Changes

- **New hook `useDeleteShortcut`** (`web_src/src/ui/CanvasPage/useDeleteShortcut.ts`) — listens for plain `Delete` or `Backspace` on `window`, skips when focus is inside any editable element (`input`, `textarea`, `select`, `[contenteditable]`, `.monaco-editor`). Follows the same pattern as the existing `useBuildingBlocksShortcut`.
- **Wired up in `CanvasPage`** — fires `onNodesDelete` / `onNodeDelete` for currently-selected nodes, with a confirmation dialog matching the existing multi-select toolbar trash button. Reads selection from `useReactFlow().getNodes()` to avoid the one-render lag that `stateRef.current.nodes` has.
- **Removed `autoFocus` from the Name input** in `componentSidebar/SettingsTab.tsx` — clicking a component should select it without stealing keyboard focus into a text input. Without this, even with the new keyboard handler, Delete would just edit characters in the Name field instead of deleting the node.
- **Context-aware confirmation copy** — *"this component"* vs *"the N selected components"*. Applied to both the new keyboard path and the existing multi-select toolbar trash dialog (was always plural before; now consistent).

## Test plan

- [x] Unit tests cover the fire/skip matrix (11 tests in \`useDeleteShortcut.spec.tsx\`):
  - Plain Delete fires
  - Plain Backspace fires
  - Any modifier (Cmd/Ctrl/Alt/Shift) suppresses
  - Other keys ignored
  - Focus inside input / textarea / select / contenteditable / Monaco suppresses
  - Disabled flag suppresses
  - Listener removed on unmount
- [x] Manual smoke test against latest \`main\`:
  - Click a component → press \`Delete\` → confirms → deletes
  - Click into Name field → press \`Backspace\` → edits text, doesn't delete component
  - Press \`Delete\` inside an expression editor (Monaco) → edits text, doesn't delete component
  - Multi-select Cmd-click + \`Delete\` → confirms with correct count → deletes
  - In read-only / live mode → shortcut inert

## Known limitation (NOT introduced by this PR)

There's a pre-existing bug in \`workflowv2/index.tsx\`'s \`handleNodesDelete\` (plural handler) where the first delete attempt sometimes silently fails and a second press is needed. This affects the existing multi-select toolbar trash button on \`main\` and is inherited by this PR's keyboard shortcut for the same reason. Reproducible on plain upstream/main without any of these changes. Will file a separate issue.

### DIFF
--- a/web_src/src/ui/CanvasPage/index.tsx
+++ b/web_src/src/ui/CanvasPage/index.tsx
@@ -70,6 +70,7 @@ import { CustomEdge } from "./CustomEdge";
 import { Header } from "./Header";
 import { RightSideControls } from "./RightSideControls";
 import { useBuildingBlocksShortcut } from "./useBuildingBlocksShortcut";
+import { useDeleteShortcut } from "./useDeleteShortcut";
 import type { CanvasPageState } from "./useCanvasState";
 import { useCanvasState } from "./useCanvasState";
 import type { TriggerActionModal } from "@/pages/workflowv2/mappers/types";
@@ -1874,7 +1875,7 @@ function CanvasContent({
   canCreateIntegrations?: boolean;
   onLogView?: () => void;
 }) {
-  const { fitView, screenToFlowPosition, getViewport, getInternalNode, setViewport } = useReactFlow();
+  const { fitView, screenToFlowPosition, getViewport, getInternalNode, getNodes, setViewport } = useReactFlow();
   const { zoom } = useViewport();
   const isReadOnly = readOnly ?? false;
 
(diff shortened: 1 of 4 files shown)


shiroyasha:
Thank you for opening this PR! Looks interesting.

Before we would be able to legally use this PR on our open source repository, we would need you to sign the DCO https://github.com/superplanehq/superplane/blob/main/docs/contributing/commit_sign-off.md

shiroyasha:
I'll close this PR for now, even though I would like to have it merged.

@jwl3g14 thank you for opening it. When you sign the DCO, please re-open the PR. 🙏

PR #4865, opened 2026-05-18
feat: polish runs canvas view UX

ropsii:
Align run detail modal status chips with canvas event badges; restrict selection and modal opens to run participants; muted non-participant node chrome; full- opacity connectors in runs view; fix hover-card Radix import and remove erroneous radix-ui package.


forestileao:
👍

PR #4867, opened 2026-05-18
fix: make run detail URLs clickable

### LINKED ISSUE #4858: Run View: Links don't work in node popup (e.g. Semaphore workflow link)
When clicking on a node in the run view, links in the popup (e.g. link to the Semaphore workflow) are not clickable or do not work correctly.

forestileao:
## Summary
- render run detail URLs as anchors in the run node popup
- preserve existing plain-text rendering for non-URL values
- add coverage for clickable URL details in the modal spec

Fixes #4858


forestileao:
👍

PR #4890, opened 2026-05-19
feat: Agent session resources, dual send/stop, build-mode rubric

bender-rodriguez-unit1:
### Mounted reference files
Agent sessions now mount 52 reference files (workflow guides + vendor component docs). The agent reads them on demand instead of relying on skills or web fetches.

Configured via `ANTHROPIC_SESSION_RESOURCES_FILE` env var pointing to a JSON file mapping Anthropic file IDs to mount paths.

### Send while streaming
Users can send messages while the agent is working (e.g. while sub-agents research). Both Send and Stop buttons appear during streaming.

### Build mode spec confirmation
In Build mode, clicking "Start Building" on a rubric sends a short confirmation instead of echoing the full spec back. The agent already has context — no need to repeat it.

Plan mode behavior unchanged (outcome + grading loop).


pkg/config/config.go
```diff
@@ -42,3 +43,26 @@ func LoadAnthropicAgentConfig() AnthropicAgentConfig {
...
+
+// SessionResource defines a file to mount in agent sessions.
+type SessionResource struct {
+	FileID    string `json:"file_id"`
+	MountPath string `json:"mount_path"`
+}
+
+// LoadSessionResources reads the session resources config from ANTHROPIC_SESSION_RESOURCES_FILE env var.
+func LoadSessionResources() []SessionResource {
+	path := os.Getenv("ANTHROPIC_SESSION_RESOURCES_FILE")
```
forestileao: Let's follow the pattern of calling env variables from config so they are not spread in the codebase

forestileao: We need to check why we need this

pkg/agents/anthropic/session_resources.json
```diff
@@ -0,0 +1,262 @@
...
+  },
+  {
+    "type": "file",
+    "file_id": "file_011CbBVL1GhCrkv2Wt42NXLb",
+    "mount_path": "ref/canvas-yaml-spec.md"
+  },
+  {
+    "type": "file",
+    "file_id": "file_011CbBVL3AHtjp14Hxj8hRSD",
+    "mount_path": "ref/components-and-triggers.md"
```
forestileao: What if these file ids change, cant we get them from Anthropic api instead? they seems static here

forestileao:
👍

PR #4891, opened 2026-05-19
feat: add support for HEAD HTTP method in Core component

### LINKED ISSUE #4866: HTTP component: support HEAD request method
The HTTP component should support the HEAD request method in addition to the existing methods (GET, POST, PUT, DELETE, PATCH).

**Use case:** Health checks, URL validation, and checking resource metadata without downloading the full response body.

**Expected behavior:** Users can select HEAD as a method option in the HTTP component configuration.

WashingtonKK:
Closes: https://github.com/superplanehq/superplane/issues/4866

This update introduces the HEAD method to the HTTP component, allowing users to make requests that do not return a body. The HEAD method is useful for health checks, URL validation, and inspecting response headers without downloading the full response payload.

### DIFF
--- a/docs/components/Core.mdx
+++ b/docs/components/Core.mdx
@@ -489,7 +489,11 @@ The HTTP component allows you to make HTTP requests to external APIs and service
 
 ### Supported Methods
 
-- GET, POST, PUT, DELETE, PATCH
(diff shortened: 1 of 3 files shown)


shiroyasha:
LGTM

PR #4911, opened 2026-05-20
fix: JSON Data Encoding

andrecalil:
## Summary

Fixes canvas event JSON handling so numeric payload values are not silently converted to float64 and later rendered in scientific notation or with precision loss.

Changes:

- Replaced `CanvasEvent.Data` JSON storage with a number-preserving `models.JSONValue`.
- Decode webhook payloads with `json.Decoder.UseNumber()`.
- Preserve number tokens when events are emitted, re-emitted, and produced from node executions.
- Added protobuf serialization handling for `json.Number`, stringifying only unsafe/exponent-prone values at the Struct boundary.
- Kept non-number JSON types unchanged: objects, arrays, strings, booleans, and nulls.
- Added tests for storage scanning, webhook parsing, expression interpolation, and canvas event serialization.


pkg/grpc/actions/canvases/structpb_compatible.go
```diff
@@ -0,0 +1,84 @@
...
+	if shouldSerializeJSONNumberAsString(raw) {
+		return raw
+	}
+
+	number, err := value.Float64()
+	if err != nil {
+		return raw
+	}
+
+	return number
```
mchalapuk, line 52: I'm trying to understand this logic. Do we have tests for those branches in the code?

pkg/grpc/actions/canvases/structpb_compatible.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		return out
+	default:
+		return value
+	}
+}
+
+func jsonNumberForStructpb(value json.Number) any {
+	raw := value.String()
+	if shouldSerializeJSONNumberAsString(raw) {
```
mchalapuk: Does this mean that we will serialize some numbers as strings and some as numbers? Are we sure that we want that? It seems to me like something very hard to remember when developing new features.

pkg/models/json_value.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+
+	"gorm.io/driver/mysql"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/schema"
+)
+
+// JSONValue stores arbitrary JSON while preserving JSON numbers as json.Number
+// when values are scanned back from the database.
```
mchalapuk: I thought the goal is to serialize the IDs to JSON without engineering notation before writing them into the DB. This comment suggests the other way around. 

Could you elaborate? Does this PR fix reading or writing to the DB? Or maybe both? Just trying to understand.

mchalapuk (commented):
This review is more about me trying to understand what's going on in the code than anything else.

The PR seems to me like something that would work.

I'm wondering if there's a simpler solution that doesn't involve a special structure for handling JSON in the code and automatically converting the number values to strings.

Wouldn't it be simpler to explicitly convert those strings to numbers when we're creating the JSON structures instead of this special type and auto-convert logic?

pkg/models/json_value.go
andrecalil: Both, but the important bug source is read/decode boundaries. Webhook decoding now preserves incoming numeric tokens before writing, and CanvasEvent.Data now preserves numeric tokens when scanning stored JSON back from the DB. Writing then naturally keeps json.Number as its original JSON token. If a caller already gives us a float64, we can’t recover the original token, so the fix is to prevent those accidental float64s at ingress and DB scan time.

pkg/grpc/actions/canvases/structpb_compatible.go
andrecalil: You made a good point, there's a new version coming soon.

pkg/models/json_value.go
mchalapuk: But if writes are fixed then there's no extra work to be done when reading, no?

pkg/grpc/actions/canvases/structpb_compatible.go
andrecalil: Now we do

pkg/triggers/webhook/webhook.go
```diff
@@ -317,6 +319,23 @@ func (w *Webhook) HandleWebhook(ctx core.WebhookRequestContext) (int, *core.Webh
 	return http.StatusOK, nil, nil
 }
 
+func unmarshalPayload(data []byte, value any) error {
```
lucaspin: This is duplicated in pkg/models/json_value.go

andrecalil:
@cursoragent update your review

pkg/workers/contexts/process_queue_context.go
```diff
@@ -87,7 +87,7 @@ func BuildProcessQueueContext(
 		RootEventID:   queueItem.RootEventID.String(),
 		EventID:       event.ID.String(),
 		SourceNodeID:  event.NodeID,
-		Input:         event.Data.Data(),
+		Input:         map[string]any{event.NodeID: normalizeExpressionValue(event.Data.Data())},
```
lucaspin, line 90: This change is suspicious. Why is this needed?

andrecalil: To support expressions (hence the `sourceNodeID`) and to normalize different number types.
What do you find suspicious? Feel free to suggest a different approach.

lucaspin: > What do you find suspicious?

I could be misunderstanding this, but to me this looks suspicious to me because all the other changes seem to be about decoding, but for this one, you are actually changing what the `ProcessQueueContext.Input` value is. @forestileao can you take a look?

shiroyasha: What is the state here? @lucaspin @andrecalil @forestileao

andrecalil: I was waiting for Pedro to chime in while I looked at other issues/tasks.

PR #4919, opened 2026-05-20
feat: re-land editing sidebar UX (rebased from #4621)

forestileao:
Summary

  Re-applies the editing-sidebar UX changes from the reverted #4621,
   minus the two behaviors that led to the revert:

  - Skipped: the "show only installed integrations" filter — it hid
  categories during demos when the relevant integration wasn't yet
  connected.
  - Skipped: the icon-rerender flicker when expanding a category —
  block items now stay mounted inside <details> (the browser handles
   show/hide via CSS), and BlockItem is wrapped in React.memo.
  - Removed entirely: the + Integrations button + the
  connect-integrations modal, by request — including all related
  state, props, and the IntegrationGridSection component.

  Sidebar UX

  - Pointer-based resize for both BuildingBlocksSidebar (clamp
  280–640px, default 460) and componentSidebar (clamp 300–800px,
  default 380). Width persisted to localStorage and clamped on read.
  - Categories in BuildingBlocksSidebar: Core → Debugging → Memory
  ordered first, the rest alphabetical. No more filtering by
  configured integrations.
  - CategorySection simplified — drops drag-preview cloning, hover
  state, the GripVerticalIcon, the integration-status Plug
  indicator, and the conditional ItemGroup mount. Icon helper logic
  extracted into resolveCategoryIcon / renderCategoryIcon.

  Canvas append handles

  - Single-channel append (handles.tsx): tightened position (right: 
  -15, top: 18, drop transform: none).
  - Multi-channel append (multiRightHandle.tsx): per-channel layout.
   Channel-name text width is measured via canvas.measureText with
  the matching font; all rows align to the widest channel's line so
  the "+" buttons are flush regardless of name length.
  - appendHandle.tsx: new plus-hovered highlight state, horizontal
  nudge so the plus sits inside the connector, hover-driven border
  darken, preview offsets parameterised via CSS vars
  (--sp-append-preview-connector-offset-y,
  --sp-append-preview-offset-y).
  - canvas-reset.css: drops the old hitbox ::before, adds the
  plus-hovered rules to gate the hover styling.
  - RightSideControls spacing gap-1.5 → gap-2.5.

  CanvasPage / workflowv2 plumbing

  - Starter placeholder UX: clicking the floating Add Component
  button when nothing else is open now creates a free-positioned
  placeholder via onPlaceholderAdd({ position }) and opens the
  building-blocks sidebar over it. onPlaceholderAdd's sourceNodeId /
   sourceHandleId are now optional so this path doesn't need an
  edge.
  - handlePlaceholderAdd in workflowv2/index.tsx only adds an edge
  when both source fields are present.

### DIFF
--- a/test/e2e/approvals_test.go
+++ b/test/e2e/approvals_test.go
@@ -11,6 +11,7 @@ import (
 	q "github.com/superplanehq/superplane/test/e2e/queries"
 	"github.com/superplanehq/superplane/test/e2e/session"
 	"github.com/superplanehq/superplane/test/e2e/shared"
+	"github.com/superplanehq/superplane/test/support"
 )
 
 func TestApprovals(t *testing.T) {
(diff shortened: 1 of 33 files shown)


forestileao:
👍

PR #4939, opened 2026-05-21
fix: isolate edit mode sidebar state

forestileao:
## Summary
- clear component sidebar URL state when switching canvas versions or entering/leaving edit mode
- prevent node run-history loading from the component sidebar while the canvas is in edit mode
- add a CanvasPage regression test for live-only sidebar event loading

## Validation
- npm run test:run -- src/ui/CanvasPage/index.spec.tsx
- make format.js
- make check.build.ui


forestileao:
👍

PR #4951, opened 2026-05-21
fix: hide agent tab when feature is disabled

forestileao:
## Summary
- expose managed-agent feature availability separately from sidebar availability
- hide the Agent tab when managed agents are disabled, while keeping Runs and Versions available
- default the sidebar to Runs instead of trying to load Agent when stale localStorage opens the sidebar without the feature

## Tests
- cd web_src && npm run test:run -- src/components/CanvasToolSidebar/index.spec.tsx src/components/CanvasToolSidebar/useCanvasToolSidebarState.spec.tsx
- make format.js
- make check.build.ui


forestileao:
:+1:

PR #4953, opened 2026-05-21
fix: allow agent follow-up while running

forestileao:
## Summary
- keep the Agent composer send button enabled while the agent is streaming/running
- only disable Send while the send mutation itself is pending
- add unit and e2e coverage for sending a follow-up message during an active turn

## Validation
- make format.js
- make format.go
- npm run test:run -- src/components/CanvasToolSidebar/index.spec.tsx
- docker compose -f docker-compose.dev.yml exec app gotestsum --format short --junitfile junit-report.xml --packages="./test/e2e" -- -p 1 -timeout 30m -run '^TestAgentE2E$/^sends_a_follow-up_message_while_a_turn_is_still_running$'
- docker compose -f docker-compose.dev.yml exec app gotestsum --format short --junitfile junit-report.xml --packages="./test/e2e" -- -p 1 -timeout 30m -run '^TestAgentE2E$'
- make check.build.ui

### DIFF
--- a/test/e2e/agent_test.go
+++ b/test/e2e/agent_test.go
@@ -100,6 +100,28 @@ func TestAgentE2E(t *testing.T) {
 		steps.assertHidden(q.TestID("agent-stop-button"))
 	})
 
+	t.Run("sends a follow-up message while a turn is still running", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


forestileao:
👍

PR #4964, opened 2026-05-22
feat: list value support for memory components

andrecalil:
## Summary

Adds a list iteration mode to the canvas-level memory components so a single execution can write multiple rows from a list expression.

- The Add, Update, and Upsert memory components gain an "Input is a list" toggle backed by `iterateList`, `listSource`, and `itemVariable` configuration fields. Each iteration evaluates value/match expressions in a scope where the configured item variable is bound to the current list element, and a single `memory.added` / `memory.updated` / `memory.upserted` event is emitted with per-item results and a total `count`.
- Shared list semantics are factored into a new `pkg/components/memorywrite` helper covering list-mode validation, expression evaluation, and per-item value resolution.
- The execution context gains `RunWithScope(expression, scope)` so components can evaluate expressions with extra bindings without leaking into the global scope; existing `Run(expression)` callers keep working through default-scope wrappers added across the integration test fakes.
- Frontend mappers for the three memory components surface the new fields, and `docs/components/Core.mdx` documents list mode for each component.

Targets `main` directly (no longer stacked on the Console rebrand branch).

## Test plan

- [ ] `make test PKG_TEST_PACKAGES=./pkg/components/memorywrite/... ./pkg/components/addmemory/... ./pkg/components/updatememory/... ./pkg/components/upsertmemory/...`
- [ ] `make test PKG_TEST_PACKAGES=./pkg/workers/contexts/... ./pkg/integrations/grafana/...`
- [ ] `make lint && make check.build.app`
- [ ] `make check.build.ui`
- [ ] Add memory with list value (append + replace); update/upsert variants
- [ ] List source / item variable expressions evaluated per element
- [ ] Single `memory.added` / `memory.updated` / `memory.upserted` event with `count`
- [ ] Existing non-list memory writes keep working unchanged


pkg/core/component.go
```diff
@@ -46,6 +46,7 @@ type ExecutionContext struct {
 
 type ExpressionContext interface {
 	Run(expression string) (any, error)
+	RunWithScope(expression string, scope map[string]any) (any, error)
```
shiroyasha: This line is probably the biggest question mark for me in this PR. I don't take core system changes lightly.

pinging @andrecalil @lucaspin @forestileao

lucaspin: I don't like the name. "Scope" is not clear enough, I had to look at the code to see what scope is, and it looks like it's just extra variables being made available to the expression. `RunWithExtraVariables` or something like that would be better.

andrecalil: Updated to `RunWithExtraVariables`

PR #4968, opened 2026-05-22
feat: add canvas visual diff

forestileao:
## Summary
- Add visual draft diff styling for added, edited, and removed canvas nodes and edges
- Add a hoverable See diff action on edited nodes that opens the split YAML diff for only that node
- Keep position-only node moves out of the edited visual status

## Tests
- make format.js
- make check.lint.ui
- make check.build.ui
- cd web_src && npm run test:run -- src/pages/workflowv2/draftNodeDiff.spec.ts src/ui/CanvasPage/Block.spec.tsx


forestileao:
👍

PR #4969, opened 2026-05-22
feat: polish agent sidebar and other ux goodies

ropsii:
Improve canvas agent UX with refined transcript, composer, and widget styling, coordinated sidebar resizing, and a Cursor-style app switcher for quick navigation between recently opened canvases.


forestileao:
👍

PR #4986, opened 2026-05-25
fix: prevent stale draft diff state

forestileao:
## Summary
- Prevent clean drafts from showing nodes as edited due to representation-only YAML differences
- Clear stale draft specs after publishing so deleted node and edge diff visuals do not reuse old draft state

## Tests
- npm run test:run -- src/pages/workflowv2/draftNodeDiff.spec.ts src/pages/workflowv2/index.spec.ts
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #4988, opened 2026-05-25
feat: register and load console.yaml when installing from GitHub

### LINKED ISSUE #4979: Load console.yaml when installing an application from GitHub

re-pixel:
Resolves: https://github.com/superplanehq/superplane/issues/4979

### Summary

Besides `canvas.yaml`, SuperPlane now sees `console.yaml` file in the root directory of GitHub repos. It uses that file to load the console view, all the panels and the layout.

If the `console.yaml` is missing, the app starts with a blank console.

If the `console.yaml` is present and malformed, the installation is aborted. This behaviour is to be discussed.

### DIFF
--- a/docs/prd/console-and-widgets.md
+++ b/docs/prd/console-and-widgets.md
@@ -411,6 +411,17 @@ Rules:
 
 Frontend YAML parsing lives in `dashboardYaml.ts`. Backend YAML parsing and validation lives in `canvas_dashboard_yml.go`. Keep error behavior and accepted shapes aligned.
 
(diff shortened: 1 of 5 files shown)


re-pixel:
@andrecalil @shiroyasha i implemented this so that we abort the installation and return an error if the console.yaml is malformed. Do you think that it is better to try the best-effort approach and load just the canvas and inform the user that the console.yaml is malformed and leave it blank.

shiroyasha:
This is a great start. Lets merge and test it.

PR #5002, opened 2026-05-25
fix: align canvas mode overlays with sidebar layout

forestileao:
## Summary
- render Console and Memory overlays inside CanvasPage's main canvas region so the left tool sidebar consumes layout width consistently
- adjust overlay positioning to fill that region instead of offsetting from the page header
- add a CanvasPage layout regression test for mode overlays

## Tests
- cd web_src && npm run test -- src/ui/CanvasPage/index.spec.tsx
- make format.js
- make check.build.ui


forestileao:
Closing because the sidebar overlay behavior has been addressed on main.

shiroyasha:
Ah! Sorry, didn't know that you planned to work on it today.

PR #5005, opened 2026-05-25
feat: add global command palette

forestileao:
## Summary
- Add a global Cmd/Ctrl+K command palette for app navigation, creation actions, and account/admin utilities
- Add contextual canvas commands and nested pages for organization settings, canvas settings, canvas search, and installation admin routes
- Make canvas components searchable directly from the root command palette when viewing a canvas
- Keep the canvas toolbar magnifying-glass control and wire it to open the global command palette
- Wire Cmd/Ctrl+/ to create a canvas when an organization is active and permissions allow it
- Split the palette into focused modules for actions, data/model state, pages, list items, routes, and types

## Tests
- make format.js
- cd web_src && npm run test -- src/components/GlobalCommandPalette/index.spec.tsx
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5025, opened 2026-05-26
feat: introduce files tab in the app page

lucaspin:
Introduce a new files tab in the app page. This is the first pull request in the effort to have the app backed by a Git repository. This new tab is where the files in that repository will live. In this pull request, we are just introducing that tab, and moving the canvas and console YAMLs there, so this is a frontend-only change for now. We use https://trees.software/ to display the file tree.


lucaspin:
👍

PR #5026, opened 2026-05-26
fix(ui): no overlapping between sidebar and files tab

lucaspin:
Fixes a small issue in the new files tab, where an open left sidebar overlaps with it. Just updates the files tab to use the same approach as the other tabs to not allow them to overlap.


lucaspin:
👍

PR #5034, opened 2026-05-27
fix: update template app metadata

forestileao:
## Summary
- add manifest-level lead icons for template apps
- use manifest icons in home template list and detail modal
- rewrite template initial agent messages and next-step copy
- update the agent boot context unit test for the new next-step copy

## Tests
- make format.js
- make check.build.ui
- docker compose -f docker-compose.dev.yml exec app bash -c "cd web_src && npm run test:run"


forestileao:
👍

PR #5049, opened 2026-05-28
fix: prevent undefined organization logo link

forestileao:
## Summary
- Route the SuperPlane logo to organization selection when no organization is active
- Keep active organization logo navigation unchanged
- Add regression coverage for both logo link states

## Tests
- cd web_src && npm run test:run -- src/components/OrganizationMenuButton.spec.tsx
- make check.lint.ui
- make check.build.ui
- git diff --check


forestileao:
👍

PR #5050, opened 2026-05-28
fix: expose canvas version role permissions

forestileao:
## Summary
- Add canvas version edit permission to the role create/edit UI
- Add canvas version publish permission to match RBAC policy

## Tests
- make format.js
- make check.lint.ui
- make check.build.ui
- git diff --check


forestileao:
👍

PR #5051, opened 2026-05-28
fix: share builder stage for demo image

forestileao:
## Summary
- move the demo image target into the root Dockerfile so it uses the same dev-base and builder stages as the standard image
- keep `runner` as the Dockerfile default target and make the standard image release script explicitly build `--target runner`
- update the demo image release script to build the root Dockerfile `demo` target
- remove the duplicated demo Dockerfile that drifted from the shared builder setup

## Validation
- `bash -n release/superplane-image/build.sh`
- `bash -n release/superplane-demo-image/build.sh`
- `git diff --check`
- `docker buildx build --platform linux/amd64 --progress plain --provenance=false -f Dockerfile --build-arg BASE_URL=https://app.superplane.com -t superplane-default-local-test --load .`
- `docker buildx build --target demo --platform linux/amd64 --progress plain --provenance=false -f Dockerfile --build-arg BASE_URL=https://app.superplane.com -t superplane-demo-local-test --load .`


forestileao:
👍

PR #5057, opened 2026-05-28
fix: preserve console panel edits

forestileao:
## Summary
- add a panel YAML diff action while editing console panels
- reuse the canvas YAML diff modal with panel-specific labels
- optimistically update the console dashboard cache while saves are in flight so tab switches keep recent panel edits visible
- roll back the optimistic console cache update if save fails

## Tests
- make format.js
- make check.lint.ui
- make check.build.ui
- npm run test:run -- src/hooks/useCanvasData.spec.ts src/pages/workflowv2/dashboard/PanelEditorDialog.spec.tsx


forestileao:
👍

PR #5061, opened 2026-05-29
feat: repository provisioning

lucaspin:
### Changes

- New repositories table
- New RepositoryProvisioner worker, controlled by START_REPOSITORY_PROVISIONER environment variable
- Canvas lifecycle updates

### Flow

- When canvas is created, a new pending repository record is created for the canvas.
- RepositoryProvisioner scans for pending repositories - it also listens to RabbitMQ messages about canvas creations to speed up things a bit. For every new pending repository, it calls the provider's CreateRepository()
- If successful, moves repository to ready state. Otherwise, moves it to error state.


lucaspin:
👍

PR #5066, opened 2026-05-29
feat: canvas repository APIs

lucaspin:
Introduce the new canvas repository APIs so the UI and CLI can manage files.

### New endpoints

- GET /api/v1/canvases/{canvas_id}/repository
- GET /api/v1/canvases/{canvas_id}/repository/files
- POST /api/v1/canvases/{canvas_id}/repository/commits
- GET /api/v1/canvases/{canvas_id}/repository/file?path=

### Downloading single file

For downloading a single file, we do not use the proto + gRPC gateway approach. There is no way to use that and not download the file into memory instead of streaming it. For that endpoint specifically, we register a route directly in the HTTP server handlers.

### Lazily provisioning repositories

When GetCanvasRepository is called for an existing canvas that does not yet have a repository provisioned, we create a pending repository for it. This lazily provisions repositories for all existing canvases that are still in use.


shiroyasha (dismissed):
🚀

lucaspin:
👍

PR #5068, opened 2026-05-29
feat(ui): manage repository files

lucaspin:
Updating UI to consume APIs exposed in [#5066](https://github.com/superplanehq/superplane/pull/5066), allowing users to manage repository files in the UI. This is just the first iteration of this, so UX needs a bit of work still.


lucaspin:
👍

PR #5076, opened 2026-05-29
fix: sync anthropic agent prompt on startup

forestileao:
## Summary
- embed the current Anthropic managed agent system prompt in pkg/agents/anthropic/agent_prompt.md
- add startup sync so the Anthropic agent prompt is updated from code before provider initialization
- add tests for embedded prompt, no-op sync, and update sync

## Validation
- make format.go
- make lint
- make test PKG_TEST_PACKAGES=./pkg/agents/anthropic *(blocked by existing generated proto mismatch: missing pbCanvases canvas repository FullMethodName symbols)*
- make check.build.app *(blocked by the same generated proto mismatch)*


forestileao:
👍

PR #5078, opened 2026-05-29
chore: update self-hosted release to use supergit

lucaspin:
Git provider is now required for SuperPlane to function, and the release installation artifacts we have (demo image, tarball, Helm chart) have not been updated to use [supergit](https://github.com/superplanehq/supergit). This pull request fixes that.


lucaspin:
👍

PR #5092, opened 2026-05-30
feat: Add Ansible component for playbooks and ad-hoc commands

dkozlov:
## What

Adds a built-in **`ansible`** action component that runs Ansible **playbooks** and **ad-hoc commands** from a Canvas.

## Why

SuperPlane ships execution components (`ssh`, `runner`, `http`) but nothing for configuration management / infrastructure automation. Ansible is a natural fit: a node that, given a playbook (or an ad-hoc module + host pattern) and an inventory, runs Ansible and routes the workflow on `success` / `failed`, surfacing the play recap (ok/changed/unreachable/failed) and stdout/stderr.

## How

- **One component, two modes** (`Playbook` / `Ad-hoc`) via conditional config fields, modeled on the existing `ssh` component.
- The SuperPlane process acts as the **Ansible control node**; it reaches managed hosts through the provided inventory (SSH, or `ansible_connection=local` for a self-contained run).
- Playbook mode runs with the JSON stdout callback so the **per-host recap** is captured; routes `success` (exit 0) / `failed` (non-zero). If Ansible can't run at all (missing binary, timeout), the run finishes in the **error** state.
- **Security:** argv is built as a slice (no shell — no `sh -c`), module/var names and argv values are validated, and no secrets are logged or emitted.
- Adds `ansible` to the `dev-base` image as a dedicated layer (`DEBIAN_FRONTEND=noninteractive`).
- Unit tests cover validation, argv construction (playbook + ad-hoc, become/extra-vars/limit/verbosity), recap parsing, and `success`/`failed` channel routing.

## Verified

Built and run end-to-end against a local dev instance: a `Start → Ansible (playbook) → Smoke test (ad-hoc) → Display` canvas executes, the playbook performs real file operations on `localhost`, the recap is captured (`ok/changed/...`), and routing follows `success`/`failed`.

## Notes

- The component shells out via `os/exec` on the node running the executor — appropriate for self-hosted/dev; flagged for reviewers.

## Files

- `pkg/components/ansible/` — `ansible.go`, `runner.go`, `example.go`, `example_output.json`, `ansible_test.go`
- `pkg/registryimports/registryimports.go` — register the component
- `Dockerfile` — install `ansible` in the dev image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,16 @@ WORKDIR /app
 
 RUN mkdir -p "${PLAYWRIGHT_BROWSERS_PATH}"
 RUN playwright install chromium-headless-shell --with-deps
+
+# Ansible is required by the built-in `ansible` component, which uses this
+# container as the Ansible control node. Kept as a dedicated layer so it does
+# not invalidate the cache for the expensive toolchain layers above.
+# DEBIAN_FRONTEND=noninteractive avoids the tzdata install prompt.
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
+  DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends ansible && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
+
 RUN rm -rf /opt/install /opt/install-scripts /tmp/*
 
 CMD [ "/bin/bash",  "-c", "sleep infinity" ]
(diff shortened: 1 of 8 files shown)


shiroyasha:
Very interesting! Thank you for opening this PR.

We were just discussing the need for Ansible components.
 
One thing we are not yet sure is if we would support them in this format, or rather as a command that runs inside of a runner where we would be able to clone the repository that contains the ansible commands.

We'll build a POC of the runner backed approach in the coming weeks. Lets review this PR again once we have it.

PR #5095, opened 2026-05-31
fix: format public server listen address with net.JoinHostPort

alan747271363-art:
## Summary
- uses `net.JoinHostPort` for the public server listen address
- adds coverage for IPv4 and IPv6 listen address formatting

## Validation
- `git diff --check HEAD~1..HEAD`
- Go tests were not run locally because this Windows environment does not have `go`/`gofmt` available. The targeted command to run in a Go environment is `go test ./pkg/public -run Test__ListenAddr -count=1`.

### DIFF
--- a/pkg/public/server.go
+++ b/pkg/public/server.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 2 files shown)


shiroyasha:

I really appreciate you taking the time to look into this. From what I can tell, this PR doesn’t seem to be connected to an existing issue or a known installation problem. Please let me know if I’m missing some context here.

Since the PR has been stale for a while, I’ll close it for now. That said, I’d be happy to reopen it or take another look if you can share more details about the problem it solves.

Thanks again for contributing!

PR #5096, opened 2026-05-31
feat: update CLI to expose app commands

lucaspin:
Renames the CLI top-level canvases command to apps and reorganizes subcommands to match the product model: an app is the unit that contains a canvas, console, and files. App lifecycle commands (create, delete, change-requests) live at the app level.

#### Before

```
superplane canvases list|active|create|delete|get|init|update|change-requests …
superplane console get|set
```

#### After

```
superplane apps list|active|create|delete|change-requests …
superplane app canvas get|init|update
superplane app console get|set
superplane app files tree|show
```

### Other changes

Most of the commands remain with the same flags, except for a few.

#### Create command

The create command was about creating canvases, and now it is about creating apps. Some renaming in the flags are in order, to make it clearer what they are.

```
# Before
superplane canvases create \
  --file canvas.yaml \
  --auto-layout ... \
  --auto-layout-scope ... \
  --auto-layout-node ..

# Now
superplane apps create \
  --canvas-file canvas.yaml \
  --canvas-auto-layout ... \
  --canvas-auto-layout-scope ... \
  --canvas-auto-layout-node ..
```


lucaspin:
👍

Issue #5033, opened 2026-05-27
GitHub status event trigger


markoa:
I want to implement SuperPlane equivalent of Palantir's [Bulldozer](https://github.com/palantir/bulldozer) and need this trigger component.

> Bulldozer subscribes to status and review webhook events and automatically merges pull requests once all required status checks pass and the configured merge criteria (typically the presence of a `merge-when-ready` label) are met. Bulldozer is stateless and evaluates merge eligibility on every event, taking no action until all conditions are satisfied.

markoa:
Implemented in two referenced PRs.

PR #5104, opened 2026-06-01
fix: clarify GitHub commit status trigger

markoa:
Clarifies the GitHub status trigger now that it has been exercised against real PR checks.

  - Rename the user-facing trigger label from “On Status” to “On Commit Status”
  - Clarify that the trigger listens to GitHub commit status events from the legacy Commit Statuses API
  - Update generated component docs to explain that commit statuses are separate from (yet to be supported) Checks API check runs
  - Improve event details by replacing “Sender” with “Status creator”, adding “Commit author”, and reordering high-
  signal fields
  - Include the status state in event titles, e.g. `ci/build succeeded - d6f3c8a`

Related to #5033.


pkg/integrations/github/components/statuses/on_commit_status.go
```diff
@@ -26,15 +26,17 @@ func (t *OnStatus) Name() string {
 }
 
 func (t *OnStatus) Label() string {
-	return "On Status"
+	return "On Commit Status"
```
lucaspin, line 29: The component name should be changed to `github.onCommitStatus` too, right?

markoa: yes will do

markoa: renamed

lucaspin (approved):

PR #5113, opened 2026-06-01
feat: unify new app page and polish home onboarding UI

ropsii:
Replace the new app modal with a shared /apps/new experience, refresh apps catalog and card styling, and refetch canvas lists after template install so the homepage shows newly created apps.


lucaspin:
👍

PR #5123, opened 2026-06-01
feat: Memory Manual Entry

### LINKED ISSUE #5122: Memory: Allow manual entry
- When in edit mode, the user should be able to create memory banks or update memory banks that had been manually created. The edit view is going to be a simple JSON input (with proper syntax highlight for usability).
- The user should not be able to add/edit memory banks created via runs/nodes (only delete, as today).
- Only users with edit permissions should be able to create memory entries (like the delete today enforces).

andrecalil:
## Summary
- When in edit mode, the user should be able to create memory banks or update memory banks that had been manually created. The edit view is going to be a simple JSON input (with proper syntax highlight for usability).
- The user should not be able to add/edit memory banks created via runs/nodes (only delete, as today).
- Only users with edit permissions should be able to create memory entries (like the delete today enforces).

Resolves #5122

### DIFF
--- a/db/migrations/20260601220232_add-canvas-memories-source.up.sql
+++ b/db/migrations/20260601220232_add-canvas-memories-source.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE canvas_memories
+  ADD COLUMN IF NOT EXISTS source TEXT NOT NULL DEFAULT 'node';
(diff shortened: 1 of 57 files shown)


shiroyasha:
Is Memory Bank == Memory namespace?

andrecalil:
> Is Memory Bank == Memory namespace?

Yes! Sorry, I've been using `bank` lately.

shiroyasha:
Lets stick to one naming convention in the whole app and code. Our coding agents will get confused if we start using synonims. 

I would go with namespaces right now. We can change it to bank later.

andrecalil:
> Lets stick to one naming convention in the whole app and code. Our coding agents will get confused if we start using synonims.
> 
> I would go with namespaces right now. We can change it to bank later.

done

PR #5132, opened 2026-06-02
feat: stream agent output

### LINKED ISSUE #5032: Agent output streaming
Right now agent is "thinking" for a very long time then immediately renders entire output. Users nowadays expect streaming output.

markoa:
Implements streaming agent output for the canvas agent conversation.

  The main fix is to align our Managed Agents flow with Anthropic’s documented sequence: subscribe to the session event stream first, then send the user message after the stream is open. This avoids missing early provider output and lets the UI render assistant text as it arrives instead of staying on “Thinking…” until the full response is available.

  ## What changed

  - Moved user-message forwarding into the agent stream worker’s stream-open path.
  - Kept RabbitMQ payloads identifier-only; message content and fresh JWT preambles are re-read/generated server-
  side.
  - Added an `afterOpen` hook to the provider stream interface.
  - Added Anthropic stream mapping that accumulates streamed assistant text into one updatable assistant message
  row.
  - Updated worker persistence so repeated streamed assistant chunks update the same row.
  - Updated the frontend thinking indicator to hide once assistant output exists for the current turn.
  - Expanded unit, support-provider, worker, Anthropic provider, and E2E coverage.
 
Resolves #5032.

### DIFF
--- a/pkg/agents/anthropic/anthropic_test.go
+++ b/pkg/agents/anthropic/anthropic_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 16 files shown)


shiroyasha:
Closing as stale.

Lets use this as an inspiration when we continue this effort.

PR #5142, opened 2026-06-02
chore: use VITE_ASSET_BASE_URL for vite build

lucaspin:
Preparation for moving asset loading to assets.superplane.com


lucaspin:
👍

PR #5143, opened 2026-06-02
chore: specify VITE_ASSET_BASE_URL during image build

lucaspin:
Specify VITE_ASSET_BASE_URL during image build. Also, FRONTEND_PREBUILT to control if Vite build should run or is already available.


lucaspin:
👍

PR #5144, opened 2026-06-03
feat: add runner fleet selection

0vertake:
## Summary
- Add Runner fleet discovery through SuperPlane so the UI can load fleets from the task-broker.
- Persist the selected `fleet_id` in Runner configuration and send it when creating broker tasks.
- Add a two-step Runner picker for architecture and runner size, with product labels instead of raw instance types.
- Gate Runner availability with `RUNNER_ENABLED` while preserving per-org experimental feature behavior.

## Test plan
- `go test ./pkg/components/runner`
- `go test ./pkg/features`
- `make lint && make check.build.app`
- `cd web_src && npx vitest run src/ui/configurationFieldRenderer/RunnerFleetFieldRenderer.spec.ts`
- Local smoke: real task-broker + runner completed a task successfully, and `/api/v1/runner/fleets` returned registered fleets.

### DIFF
--- a/pkg/components/runner/broker.go
+++ b/pkg/components/runner/broker.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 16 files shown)


0vertake:
@shiroyasha

shiroyasha:
The listing from the task-broker became complicated. We don't have architectural support for it in SuperPlane. Closing this one and hardcoding it the values here:
https://github.com/superplanehq/superplane/pull/5146

PR #5156, opened 2026-06-03
fix: UI uses /apps on routes

lucaspin:
Updates the UI to use /{organizationId}/apps/{appId} instead of /{organizationId}/canvases/{canvasId}. API paths (/api/v1/canvases/...) remain unchanged for now — only client-side routing was updated.

### Route changes

- App page: apps/:appId
- App settings: apps/:appId/settings
- Legacy redirects from /canvases/:canvasId → /apps/:appId (query params preserved), so old bookmarked /canvases/... URLs redirect automatically.


lucaspin:
👍

PR #5160, opened 2026-06-04
feat: reorganize runs UX with sidebar drill-down and bottom detail pane

ropsii:
Replace the run node detail modal with a resizable bottom pane, add run sidebar drill-down with execution chain navigation, and polish runs list and filter UI.


markoa:
Pass this on to your agent to address + ask to squash all commit into one and force push for clean history:

---

I got the following feedback in code review:

- [P2] Handle deep-link detail requests after mount
  - File: `web_src/src/components/CanvasToolSidebar/RunsTabPanel.tsx:52-53`
  - When a `?view=runs&run=...` deep link is applied after `RunsTabPanel` has already mounted, for example browser navigation from another view into runs mode, `initialOpenDetail` can flip to true later but this initializer is never re-read. The sidebar can therefore stay on the list instead of opening the intended run detail.
  - The adjacent effect also ignores the first non-null `selectedRunId`, so URL-driven selection from `null` has the same behavior.
  - Please sync `sidebarView` from `initialOpenDetail` / `selectedRunId` when the deep-link request arrives.

- [P3] Split the detail component instead of disabling lint
  - File: `web_src/src/ui/Runs/RunNodeDetailContent.tsx:1`
  - This new file starts by disabling `max-lines-per-function` and `complexity` because `RunNodeDetailContent` combines tab state, keyboard handling, navigation, data derivation, and all tab rendering in one component.
  - That makes the revamped UI harder to maintain. Please extract the tab-state logic and header/details/tab views so the lints remain useful and future changes are safer.

markoa (approved):

PR #5180, opened 2026-06-05
chore: Add clean code skill

markoa:
- Add the `clean-code` agent skill under `.agents/skills/clean-code`, vendored from https://github.com/markoa/clean-code-skill
- Allow repo-local agent skills to be tracked while keeping other `.agents` state ignored
- Update `AGENTS.md` to explicitly require clean code and test-first development by default


markoa:
@shiroyasha pls note the addition of `.agents/` directory which was previously gitignored. this is a convention that we can't change afaik.

PR #5182, opened 2026-06-05
chore: rename /dashboard API to /console

lucaspin:
Cleaning up after the dashboard->console rename, updating the API to use the correct term.


lucaspin:
👍

PR #5186, opened 2026-06-05
chore(ui): move files tab code into pages/app/files

lucaspin:
Next one after https://github.com/superplanehq/superplane/pull/5185. Moving files tab specific code into pages/app/files. Also, renames components and files to not mention workflow/canvas anymore.


lucaspin:
👍

PR #5192, opened 2026-06-06
fix: expression preview no longer shows error when no runtime context available

### LINKED ISSUE #5148: Expression preview tool doesn't work
In my experience it consistently reports various types of errors for various valid expressions.

Purkulosh:
Closes #5148


markoa:
Thanks for the contribution @Purkulosh! After looking into this a bit more I found that more work was needed to really polish the UX. So I will close this in favor of https://github.com/superplanehq/superplane/pull/5364.

PR #5153, opened 2026-06-03
poc: git source of truth

mchalapuk:
Requires [this version of supergit](https://github.com/superplanehq/supergit/pull/2).

PR #5194, opened 2026-06-07
feat: multi-draft apps

mchalapuk:
Replaces the single implicit per-user draft with multiple named draft branches per app. Each draft is a workflow_versions row with branch_name and display_name; the builder auto-saves to whichever draft is active. This is PR 1 of the git-first stack — DB-native only; no staging, Commit UX, canonical YAML, or git backend yet.

Screenshot 2026-06-07 at 06 54 47

### Database & migration
- Adds `workflows.next_draft_display_number`, `workflow_versions.branch_name`, and `workflow_versions.display_name`
- Backfills existing drafts with synthetic `drafts/{uuid}` branch names and `Draft #N` display names
- Enforces draft branch uniqueness and draft/published branch constraints

### Backend
- `CreateDraftBranch` — clone live into a new draft version row
- `ListDraftBranches` / `DeleteDraftBranch` — list and delete drafts by branch metadata
- Stops reusing a single per-user draft in save paths; version listing and describe expose draft branch fields

### UI
- Versions sidebar — Drafts section with create, switch, and delete
- `StartEditingDropdown` — Continue existing draft / Create new draft when drafts exist
- `useActiveDraftBranch` — URL branch param + local state for active draft
- Active draft drives which version row auto-save and edit mode target
- Per-draft client spec cache when switching branches
- Draft delete confirmation dialog, blue edit-mode chrome, blue Publish button
- Publish invalidates `draftBranches` and clears branch URL state so drafts disappear from the list without refresh
- Fixes for post-create / version-switch navigation (Edit disabled, branch highlight, switching to live/historical versions while editing)

### Tests
- Backend: `canvas_version`, `list_canvas_versions`, publish/update paths
- Frontend: `StartEditingDropdown`, `DraftBranchRow`, `useActiveDraftBranch`, header edit actions
- E2E: `TestCanvasMultiDraft` — create two drafts, switch between them in the sidebar

## Related Items

POC: [superplane PR#5153](https://github.com/superplanehq/superplane/pull/5153)


pkg/grpc/actions/canvases/console_version_helpers.go
```diff
@@ -32,6 +32,15 @@ func resolveConsoleVersionID(tx *gorm.DB, canvas *models.Canvas, requestedVersio
 	return liveVersion.ID, nil
 }
 
+func canReadOwnedRegisteredDraftInTransaction(
+	_ *gorm.DB,
```
lucaspin: Why use _ parameters here? This function could probably be inlined too

pkg/grpc/actions/canvases/create_canvas_version.go
```diff
@@ -49,22 +58,26 @@ func CreateCanvasVersion(ctx context.Context, organizationID string, canvasID st
...
+		var createErr error
+		version, createErr = models.CreateDraftBranchFromLiveInTransaction(
 			tx,
-			canvas.ID,
+			canvasUUID,
 			userUUID,
-			liveVersion.Nodes,
-			liveVersion.Edges,
+			strings.TrimSpace(displayName),
+			append([]models.Node(nil), liveVersion.Nodes...),
```
lucaspin: Why append []models.Node(nil)?

pkg/grpc/actions/canvases/list_canvas_versions.go
```diff
@@ -99,6 +89,50 @@ func ListCanvasVersionsPaginated(
 	}, nil
 }
 
+func listDraftCanvasVersions(
+	_ context.Context,
```
lucaspin: If context is not used, do not pass it as parameter

pkg/grpc/actions/canvases/list_canvas_versions_test.go
```diff
@@ -1,13 +1,238 @@
...
+	"google.golang.org/grpc/status"
 )
 
 func TestGetCanvasVersionLimit(t *testing.T) {
 	require.Equal(t, uint32(DefaultLimit), getCanvasVersionLimit(0))
 	require.Equal(t, uint32(20), getCanvasVersionLimit(20))
 	require.Equal(t, uint32(MaxCanvasVersionLimit), getCanvasVersionLimit(MaxCanvasVersionLimit+1))
 }
+
+func TestCreateCanvasVersion(t *testing.T) {
```
lucaspin: The TestCreateCanvasVersion test is in the file for tests for the list operation. We should move it to the appropriate file

pkg/grpc/actions/canvases/list_canvas_versions_test.go
```diff
@@ -1,13 +1,238 @@
...
+		require.NotNil(t, version)
+		require.NotNil(t, version.GetMetadata())
+		assert.Equal(t, "Release prep", version.GetMetadata().GetDisplayName())
+
+		stored := findRegisteredDraftBranch(t, canvasUUID, version.GetMetadata().GetBranchName())
+		assert.Equal(t, "Release prep", stored.DisplayName)
+	})
+}
+
+func TestDeleteCanvasVersion(t *testing.T) {
```
lucaspin: Same comment as above

pkg/grpc/actions/canvases/update_canvas_version.go
```diff
@@ -215,6 +204,31 @@ func UpdateCanvasVersionWithUsage(
 	}, nil
 }
 
+func ensureVersionIsOwnedRegisteredDraftInTransaction(
+	_ *gorm.DB,
```
lucaspin: Same comment about parameters being passed and not used

pkg/grpc/actions/canvases/update_console.go
```diff
@@ -78,19 +78,8 @@ func UpdateConsole(
...
-		if version.OwnerID == nil || *version.OwnerID != userUUID {
-			return status.Error(codes.PermissionDenied, "version owner mismatch")
-		}
-
-		if _, draftErr := models.FindCanvasDraftByVersionInTransaction(tx, canvas.ID, userUUID, version.ID); draftErr != nil {
-			if errors.Is(draftErr, gorm.ErrRecordNotFound) {
-				return status.Error(codes.FailedPrecondition, "version is not your current edit version")
-			}
-			return draftErr
+		if err := ensureVersionIsOwnedRegisteredDraftInTransaction(tx, canvas.ID, userUUID, version); err != nil {
```
lucaspin: So this means I can see the drafts from other users, but I cannot update them? Should we hide them in that case in the UI too?

pkg/models/canvas_version.go
```diff
@@ -281,42 +336,70 @@ func PromoteToLiveInTransaction(tx *gorm.DB, version *CanvasVersion, nodes []Nod
...
 
-	// Reuse existing draft if one already exists for this user+canvas.
-	existing, findErr := FindCanvasDraftInTransaction(tx, workflowID, userID)
-	if findErr == nil {
-		existing.Nodes = datatypes.NewJSONSlice(nodes)
-		existing.Edges = datatypes.NewJSONSlice(edges)
-		existing.UpdatedAt = &now
-		if err := tx.Save(existing).Error; err != nil {
+	if nodes == nil {
+		nodes = append([]Node(nil), liveVersion.Nodes...)
```
lucaspin: Same comment about append with []Node(nil)

pkg/models/canvas_version.go
```diff
@@ -281,42 +336,70 @@ func PromoteToLiveInTransaction(tx *gorm.DB, version *CanvasVersion, nodes []Nod
...
+
+	number := canvas.NextDraftDisplayNumber
+	if number < 1 {
+		number = 1
+	}
+
+	if err := tx.
+		Model(&Canvas{}).
+		Where("id = ?", canvasID).
+		Update("next_draft_display_number", number+1).
```
lucaspin, line 365: The next_draft_display_number counter could be confusing if we only show drafts owned by a single user

pkg/models/canvas_version.go
```diff
@@ -337,6 +422,16 @@ func SaveCanvasDraftInTransaction(
...
 }
 
+func SaveCanvasDraftInTransaction(
+	tx *gorm.DB,
+	workflowID uuid.UUID,
+	userID uuid.UUID,
+	nodes []Node,
+	edges []Edge,
+) (*CanvasVersion, error) {
+	return CreateDraftBranchFromLiveInTransaction(tx, workflowID, userID, "", nodes, edges)
```
lucaspin: Why not use CreateDraftBranchFromLive directly?

pkg/grpc/actions/canvases/update_console.go
mchalapuk: Draft listing is now scoped to the authenticated user at the API layer (matching read/update permissions), so other users' drafts no longer appear in the sidebar.

pkg/models/canvas_version.go
mchalapuk: There are a few alternatives here, each having its downsides:
1. Don't count the drafts at all. Use random names instead. Could be even more confusing.
2. Leave old draft versions in the DB (soft delete them) and count drafts created by a single user when creating a new one. This might be less confusing but requires leaving data in the DB forever.

My vote is to leave the implementation as it is for now. The main goal for having it was to prevent the draft counter from going back to 1 after all drafts are deleted.

lucaspin (approved):

PR #5209, opened 2026-06-08
fix: align agent mention cursor padding

### LINKED ISSUE #5207: Agent: @ mentions with wrong cursor padding
The keybaord cursor is on wrong position after adding  node mentions using `@` in app.

Image

Image

And since this is a multiline scrollable input text area, it also is onwrong position with new lines. check the image:

Image

forestileao:
## Summary
- keep the agent mention textarea and backdrop on identical text metrics
- remove mention highlight padding/font weight so highlighted text no longer shifts the caret alignment
- resize the agent input until a max height and sync internal textarea scroll with the mention backdrop
- clamp agent input bottom scrolling to 96% before syncing the mention backdrop, avoiding the full-bottom padding mismatch
- reduce typing work in the agent composer by skipping mention candidate building while the dropdown is closed, unmounting the closed dropdown, and memoizing the composer toolbar path
- keep the agent composer under the UI lint budget by relying on dropdown cleanup instead of a duplicate effect
- filter hidden transcript messages before creating turn wrappers so empty rows do not appear above Thinking or command rows
- let long user prompts scroll normally instead of pinning them above current agent output
- add regression coverage for mention highlighting, input scroll sync, 96% bottom-scroll clamping, mention candidate gating, hidden transcript rows, and long prompt stickiness

Fixes #5207

## Tests
- make format.js
- make check.lint.ui
- npm run test:run -- MentionTextarea.spec.tsx useMentionCandidates.spec.ts BackdropContent.spec.tsx AgentConversationTranscript.spec.tsx
- git diff --check
- make check.build.ui

### DIFF
--- a/web_src/src/components/AgentSidebar/BackdropContent.spec.tsx
+++ b/web_src/src/components/AgentSidebar/BackdropContent.spec.tsx
@@ -1,0 +1,14 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { BackdropContent } from "./BackdropContent";
+
+describe("BackdropContent", () => {
+  it("renders mention highlights without changing text metrics", () => {
+    render(<BackdropContent text="@Create Release deploy" mentions={[{ label: "Create Release", startIndex: 0 }]} />);
+
+    const mention = screen.getByText("@Create Release");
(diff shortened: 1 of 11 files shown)


forestileao:
👍

PR #5213, opened 2026-06-08
fix: publish drafts with invalid integration refs

### LINKED ISSUE #5212: bug: publishing canvas draft fails with stale integration references
## Summary
Publishing a canvas draft can fail with an internal error when draft nodes contain integration references that do not exist or do not belong to the organization.

The draft can already persist those nodes with node-level validation errors like `integration not found or does not belong to this organization`, but publish currently tries to insert those invalid integration IDs into `workflow_nodes.app_installation_id`, causing a database foreign key violation.

## Expected behavior
Publishing should succeed and keep the affected nodes with node-level errors instead of failing the whole publish operation. Invalid or missing integration references should be ignored/cleared for persistence, similar to other validation errors.

## Actual behavior
The UI shows a generic `internal error` and publish fails.

Production log:

```text
failed to publish canvas version: ERROR: insert or update on table "workflow_nodes" violates foreign key constraint "workflow_nodes_app_installation_id_fkey" (SQLSTATE 23503)
```

## Reproduction notes
A draft generated by the agent included nodes with placeholder integration IDs such as:

```yaml
integration:
  id: "00000000-0000-0000-0000-000000000001"
errorMessage: "integration not found or does not belong to this organization"
```

Publishing that draft fails because the placeholder integration ID is persisted as an app installation reference.

forestileao:
## Summary
- allow draft publish to continue when a node already has an invalid or stale integration reference
- validate draft node integration IDs before writing `workflow_nodes.app_installation_id`
- persist missing, malformed, or cross-organization integration references as `NULL` database FKs while preserving the node-level error message
- keep affected nodes published in error state instead of failing the whole publish with a foreign key violation
- add regression coverage for added and updated nodes that carry stale integration IDs

Fixes #5212

## Tests
- make format.go
- make test PKG_TEST_PACKAGES=./pkg/grpc/actions/canvases/changesets
- git diff --check

### DIFF
--- a/pkg/grpc/actions/canvases/changesets/canvas_publisher.go
+++ b/pkg/grpc/actions/canvases/changesets/canvas_publisher.go
@@ -2,6 +2,7 @@ package changesets
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -214,6 +215,10 @@ func (p *CanvasPublisher) addNode(ctx context.Context, change *pb.CanvasChangese
 	//
 	// TODO: handle blueprint nodes once blueprints are enabled again
 	//
+	appInstallationID, err := p.getNodeIntegrationID(node)
+	if err != nil {
+		return err
+	}
 
 	now := time.Now()
 	newNode := models.CanvasNode{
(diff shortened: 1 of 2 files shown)


forestileao:
👍

PR #5234, opened 2026-06-09
feat: add gcp monitoring on-alert trigger

WashingtonKK:
## Summary

Adds a **trigger** to the GCP integration that fires when a Cloud Monitoring **alerting policy opens or closes an incident** — the event counterpart to the existing alerting-policy actions.

| Type | Component | Description |
|------|-----------|-------------|
| trigger | `gcp.monitoring.onAlert` | Fires when a Cloud Monitoring alerting policy opens/closes an incident |


pkg/integrations/gcp/monitoring/on_alert.go
```diff
@@ -0,0 +1,405 @@
...
+			return fmt.Errorf("failed to decode metadata: %w", err)
+		}
+	}
+
+	if ctx.Integration == nil {
+		return fmt.Errorf("a connected GCP integration is required to set up the On Alert trigger")
+	}
+	if err := ctx.Integration.RequestWebhook(struct{}{}); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
WashingtonKK: The shared webhook is a better alternative to the per node webhook. Each node auto creates a notification channel which routes notifications on the policies to it.

pkg/integrations/gcp/monitoring/on_alert.go
```diff
@@ -0,0 +1,419 @@
...
+		metadata.NotificationChannel = channelName
+	} else {
+		// The channel already exists: resync its URL and Basic-auth password so a
+		// moved node webhook URL or a rotated webhook secret keeps signed
+		// deliveries flowing, instead of Cloud Monitoring POSTing to a stale URL
+		// or with an outdated password.
+		if err := updateWebhookChannel(client, metadata.NotificationChannel, webhookURL, string(secret)); err != nil {
+			return err
+		}
+	}
```
forestileao, line 161: Shouldn't we call webhook Context to request a webhook or merge config?

pkg/integrations/gcp/monitoring/on_alert.go
```diff
@@ -0,0 +1,419 @@
...
+			"username": webhookAuthUsername,
+			"password": secret,
+		},
+	}
+	url := fmt.Sprintf("%s/%s?updateMask=labels.url,labels.username,labels.password", monitoringBaseURL, channelName)
+	if _, err := client.PatchURL(context.Background(), url, body); err != nil {
+		return fmt.Errorf("%s", apiErrorMessage("failed to update notification channel", roleHintChannelEditor, err))
+	}
+	return nil
+}
```
forestileao, line 333: Shouldnt this be in webhook handler?

WashingtonKK: The channel provisioning is node-specific (it embeds the per-node URL and secret), so it has to live in the trigger's Setup/Cleanup, not the integration WebhookHandler (which has no per-node URL/secret).

pkg/integrations/gcp/monitoring/on_alert.go
WashingtonKK: the trigger uses the node webhook context (ctx.Webhook.Setup()/GetSecret()) for the real delivery URL+secret. RequestWebhook(struct{}{}) only provisions the integration-level webhook record; since each node owns its own Cloud Monitoring channel → node URL, an empty config with CompareConfig→true is correct here.

forestileao:
👍

PR #5244, opened 2026-06-09
feat: add CLI draft lifecycle commands and --draft-id selector

### LINKED ISSUE #5237: Git SoT Phase #3: CLI draft management
A draft is a `CanvasVersion` in `STATE_DRAFT`, owned by a user, carrying both `canvas.yaml` and `console.yaml`. Users can own **multiple** drafts per app, but the CLI currently hides this: every write resolves to "the" draft (the first one owned by the user via `EnsureCurrentUserDraftVersionID` / `FindCurrentUserDraftVersionID`), and there is no way to create, list, address, or delete a specific draft.

We need to add explicit draft lifecycle commands and a `--draft-id` selector so operators (and agents) can manage drafts deterministically. The required server endpoints already exist and the Go SDK (`CanvasVersionAPI`) already exposes them, so this is purely CLI wiring — no proto or codegen changes.

## Concepts & vocabulary

- A "draft" = `CanvasVersion` with `state = STATE_DRAFT`, owned by a user.
- Metadata available from the API: `id`, `canvas_id`, `owner` (UserRef), `state`, `display_name`, `created_at`, `updated_at`.
- User-facing term in the CLI is **draft**; the underlying id is a version id.
- The selector flag is `--draft-id`.

## Command surface

New lifecycle group (alias `draft`):

```bash
superplane apps drafts create [app] [--name "<display name>"]
superplane apps drafts list   [app] [--all]
superplane apps drafts delete <draft-id> [app]
```

New selector flag on existing content commands:

```bash
superplane apps canvas  get    [app] --draft-id <id>
superplane apps canvas  update -f <file> --draft-id <id>
superplane apps console get    [app] --draft-id <id>
superplane apps console set    [app] [file] --draft-id <id>
```

`[app]` is optional everywhere and falls back to the active app (`common.ResolveAppNameOrIDArg`), matching the rest of the CLI.

## Selector behavior

1. **`--draft` removed** (replaced with `--draft-id`).
2. **No `--draft-name` lookup.** Drafts are selected by id only; `display_name` is not unique, so name-based selection is intentionally out of scope. `drafts list` shows names to help users find the id.
3. **`drafts delete` matches the `apps delete` convention.** No confirmation prompt and no `--yes` flag — it deletes directly. (See "drafts delete" below.)
4. Existing `change-requests create` accepts `--draft-id` instead of `--version-id` (consistency).

Notes:

- `--draft-id` implies draft mode; 
- For `canvas update` / `console set`, `--draft-id` **never auto-creates**. If the id does not resolve to one of the user's drafts, the command errors.

## `drafts create`

- **Args:** optional `[app]`.
- **Flags:** `--name` (maps to `display_name`).
- **Behavior:** `CanvasVersionAPI.CanvasesCreateCanvasVersion(appID).Body({display_name})`. Server forks a draft branch from live.
- **Text output:**

```
Draft created for app <app-id>
Draft ID: <version-id>
Name:     <display-name or "(none)">
```

- **Structured output (`-o json/yaml`):** render the returned `CanvasesCanvasVersion`.
- **Errors:** propagate API errors (e.g. templates are read-only → `FailedPrecondition`).

## `drafts list`

- **Args:** optional `[app]`.
- **Flags:** `--all` (list all owners; default = current user's drafts only). `--all` is best-effort — the server returns what the caller is authorized to see.
- **Behavior:** `CanvasesListCanvasVersions(appID).State("STATE_DRAFT")`, then filter by owner unless `--all`.
- **Text output:** table sorted by `updated_at` desc:

```
DRAFT ID                              NAME            OWNER            UPDATED
b2c1...                               wip-rollback    you             2m ago
9f0a...                               (none)          alice@acme.io   1h ago
```

  (Mark the current user as `you`.) Empty → `No drafts found for app <app-id>.`
- **Structured output:** array of versions (reuse renderer).

## `drafts delete`

Matches the `apps delete` convention (decision 3): direct delete, no prompt, no `--yes`.

- **Args:** required `<draft-id>`, optional `[app]`.
- **Flags:** none.
- **Behavior:** `CanvasVersionAPI.CanvasesDeleteCanvasVersion(appID, draftID)`. The server enforces draft-only, registered-draft-branch, and owner-match, so the CLI passes the id through and surfaces precise errors.
- **Text output:** `Draft deleted: <draft-id>`.
- **Structured output:** `{"id": "<draft-id>", "deleted": "true"}` (same shape as `apps delete`).
- **Error mapping:**
  - `PermissionDenied` (owner mismatch) → "you can only delete your own drafts".
  - `FailedPrecondition` (not a draft / not a registered branch) → "only draft versions can be deleted".
  - `NotFound` → "draft <id> not found".

mchalapuk:
Closes #5237.

Adds explicit draft lifecycle management to the CLI and wires a shared `--draft-id` selector through canvas, console, and change-request commands.

### superplane apps drafts

- `drafts create [app] [--name]` — fork a new draft from live
- `drafts list [app] [--all]` — table of drafts (owner, name, relative updated time)
- `drafts delete <draft-id>` — discard a draft (direct delete, no prompt)

### Content commands

- `canvas get|update` and `console get|set` accept `--draft-id` instead of`--draft`
- `--draft-id` implies draft mode (no auto-publish / no auto change-request)
- change-requests create accepts `--draft-id` instead of `--version-id` (consistency)
- `canvas get` text output now includes `Source` and `Version ID` when reading a draft

### CLI
- [x] `superplane apps drafts create <app> --name "wip"` — prints draft id
- [x] `superplane apps drafts list <app>` — shows owned drafts; `--all` shows other owners
- [x] `superplane apps drafts delete <draft-id> <app>` — deletes without confirmation
- [x] With multiple owned drafts: `canvas get --draft-id <id>` succeeds
- [x] `canvas update --draft-id <id> -f canvas.yaml` updates the targeted draft without auto-publish
- [x] `console set --draft-id <id> -f console.yaml` updates the targeted draft
- [x] `change-requests create --draft-id <id>` works


pkg/cli/commands/apps/common/common.go
```diff
@@ -36,6 +36,61 @@ func ResolveAppNameOrIDArg(ctx core.CommandContext, arg string) (string, error)
...
+			return FindAppID(ctx, ctx.API, active)
+		}
+	}
+	return findAppIDByDraftID(ctx, draftID)
+}
+
+// findAppIDByDraftID scans the drafts of every app the user can list and
+// returns the id of the app whose drafts contain draftID. It is a best-effort
+// fallback used when no app is given and none is active.
+func findAppIDByDraftID(ctx core.CommandContext, draftID string) (string, error) {
```
lucaspin: This seems like a dangerous fallback to have. Could lead to a lot of API requests

pkg/cli/commands/apps/canvas/update.go
```diff
@@ -16,6 +16,8 @@ import (
 type updateCommand struct {
 	file            *string
 	draft           *bool
```
lucaspin: Why do we need this flag still?

pkg/cli/commands/apps/common/draft.go
```diff
@@ -0,0 +1,188 @@
...
+	resolvedVersionID := ""
+	if draftID != nil {
+		resolvedDraftID = strings.TrimSpace(*draftID)
+	}
+	if versionID != nil {
+		resolvedVersionID = strings.TrimSpace(*versionID)
+	}
+
+	if resolvedDraftID != "" && resolvedVersionID != "" && !strings.EqualFold(resolvedDraftID, resolvedVersionID) {
+		return "", fmt.Errorf("--draft-id and --version-id must match when both are set")
```
lucaspin: Why keep both flags? Let's just use one

pkg/cli/commands/apps/common/draft.go
```diff
@@ -0,0 +1,188 @@
...
+		Execute()
+	if err != nil {
+		return nil, err
+	}
+
+	return response.GetVersions(), nil
+}
+
+// ListOwnedDraftVersions returns draft versions owned by userID.
+func ListOwnedDraftVersions(ctx core.CommandContext, appID, userID string) ([]openapi_client.CanvasesCanvasVersion, error) {
```
lucaspin: Why is this needed? API returns only the user's drafts, no?

pkg/cli/commands/apps/common/common.go
mchalapuk: Yeah the appId shouldn't be needed when operating on drafts.

pkg/cli/commands/apps/canvas/update.go
mchalapuk: Not needed if we completely remove the old --draft options. I'll do that.

pkg/cli/commands/apps/common/draft.go
mchalapuk: Wee'll keep `--draft-id` then.

pkg/cli/commands/apps/common/draft.go
mchalapuk: Yes, this is not needed.

The same redundancy exists in common.go: FindCurrentUserDraftVersionID / FindOwnedDraftVersionID go through draftVersionIDForOwner, which re-applies the identical owner check client-side.

pkg/cli/commands/apps/common/common.go
mchalapuk: Removed the all-apps draft scan in commit 0c0a3c5. `drafts delete` now resolves the app from the optional `[app]` argument or the active app, so there's no longer a fan-out of list calls across every app.

pkg/cli/commands/apps/canvas/update.go
mchalapuk: Removed the `--draft` flag in commit 0c0a3c5. Draft mode is now driven solely by `--draft-id` across canvas/console get/update/set.

pkg/cli/commands/apps/common/draft.go
mchalapuk: Dropped `--version-id` in commit 0c0a3c5; `--draft-id` is now the single selector and `MergeDraftOrVersionID` is gone.

pkg/cli/commands/apps/common/draft.go
mchalapuk: Removed the redundant client-side owner filtering in commit 0c0a3c5 (`ListOwnedDraftVersions`, `FindOwnedDraftVersionID`, `draftVersionIDForOwner`). The API already scopes draft listing to the authenticated user, so the resolver relies on that.

PR #5268, opened 2026-06-10
feat: commit UI

### LINKED ISSUE #5328: Git SoT Phase #4: Commit UI
Phase <span>#</span>4 of the Git Source-of-Truth (Git SoT) stack.

Introduce a git-like stage → commit → publish editing loop for multi-draft apps, backed by server-side draft staging. Canvas, console, and repository file edits should stage into an uncommitted layer first, then be explicitly committed into the draft version, and finally published to live.

POC: https://github.com/superplanehq/superplane/pull/5153

This phase is also DB-native only — it intentionally does not include staging, Commit UX, or a git backend yet. Those land in later phases of the stack.

### Server
- **Server-side staging layer**: A `workflow_staging` table/model keyed by the draft `workflow_versions.id` that stores staged YAML per path. Rows cascade away when the draft version is deleted. Staged deletion is soft (keep the row, mark deleted) so an effective read can distinguish "unstaged" from "staged removal"; discard is a hard delete that reverts staging.
- **Staging RPCs**: `StageCanvasRepositoryFile`, `DiscardCanvasStaging`, and `CommitCanvasStaging`. Effective spec reads come through the repository file `GET` with `?stage=true` (staging overlay → committed draft row).
- **Authorization**: staging / auto-layout / discard map to `update_version` (agent-allowed); commit maps to `update` (user-only). Usage limits (e.g. node limits) are enforced at commit time, not stage time.

### UI
- Canvas and console edits autosave into staging (`canvas.yaml` / `console.yaml`) instead of writing the version row directly.
- Header actions: **Reset** (revert uncommitted staging to last committed draft, with confirm dialog), **Commit** (orange — materialize the staged draft version), **Publish** (blue — promote the committed draft to live).
- Visual indicators: orange dots/badges for uncommitted staged changes, blue dots/badges for committed-but-not-yet-published changes; per-draft change dots on tabs and header.
- Files tab routes through repository file staging with Reset/Commit.
- Staged edits are isolated per draft (switching drafts must not leak staged edits).
- Console diff/X-ray and "Show diff" modal reflect the effective draft console (committed + staged), while publishable indicators diff committed vs live.

### CLI
Never operate on staging. `superplane apps canvas get --draft-id` reads only committed changes; `superplane apps canvas update --draft-id` always commits.

### Agent
canvas custom tool always operates on staging (stages spec-file edits, auto-layouts the staged canvas) and never commits or publishes — commit/publish remain user-driven.

## Out of Scope
- No git backend.
- No git repository access for users.

### LINKED ISSUE #5355: Adding a draft with CLI doesn't immediately show it in the UI
The new draft is visible on the list after refresh.

Same behavior for deleting a draft through CLI.

mchalapuk:
Introduces server-side draft staging (`workflow_staging`) and a **stage** → **Commit** → **Publish** editing loop for multi-draft apps.

- UI: Canvas/console edits stage to `workflow_staging`; Commit (orange) materializes the draft version row; Publish (blue) promotes to live. Orange/blue indicators show uncommitted vs ready-to-publish state.
- CLI and agent tools: Never operate on staging; they commit directly to the draft version row.
- Agent: Always operates on staging; never commits.

Closes #5328
Resolves #5355

### Backend

- New `workflow_staging` table and model.
- Staging RPCs: `StageCanvasRepositoryFile`, `DiscardCanvasStaging`, `CommitCanvasStaging`.
- Effective spec reads via repository file `GET` with `?stage=true` (staging overlay → committed draft row).
- Auth: staging/discard/commit-staging/auto-layout and direct repository commits map to `update_version` (agent-allowed).

### UI

- Spec autosave stages `canvas.yaml` / `console.yaml` instead of writing the version row directly.
- Header: **Reset** (uncommitted), **Commit**, **Publish**; draft change dots (orange/blue).
- Files tab wired through repository file staging.
- `DescribeCanvas` kept for metadata/runtime status; draft spec from effective YAML.

### CLI

- No changes in commands.
- Never operates on staging. `superplane apps canvas get --draft-id` reads only committed changes. `superplane apps canvas update --draft-id` always commits.

### Agents

- `update_draft` commits directly to the UI staging area of the agent's private draft version row via `StageCanvasRepositoryFiles`. Agent never commits or publishes.

### UI — staging loop

- [x] Open a draft, add/move a node → orange uncommitted indicator; live unchanged.
- [x] Commit → staging cleared; draft version has the edit; blue “ready to publish” when appropriate.
- [x] Publish → node appears on live canvas; runs still work.
- [x] Reset with uncommitted staging → reverts to last committed draft; confirm dialog.
- [x] Switch between two drafts → staged edits isolated per draft.
- [x] Files tab: `canvas.yaml` / `console.yaml` match export modal; edit → stage → commit

### CLI

- [x] `superplane apps canvas get <app> --draft-id <id>` -o yaml → committed draft only
- [x] `superplane apps canvas update --draft-id <id>` -f canvas.yaml → commits

### Agents

- [x] Builder agent commits to its private draft (staging)
- [x] After agent edit, UI shows draft in versions list; user then **Commits**


docs/contributing/agent-tools.md
```diff
@@ -9,7 +9,7 @@ to the CLI for common app operations.
 Prefer adding an action to `superplane_app` when the capability is app-related:
 
 - reading app/canvas or Console YAML
-- updating the draft app
+- staging edits on the draft app (`workflow_staging`; agents never commit or publish)
```
mchalapuk: This is wrong. The scope of this PR changed.

docs/contributing/agent-tools.md
```diff
@@ -63,7 +63,7 @@ Action rules:
 
 - Always stay scoped to the current `AgentSessionContext`.
 - Reject or ignore attempts to operate on another canvas.
-- Never publish drafts from an agent action.
+- Never commit or publish drafts from an agent action. Stage edits only; users commit and publish from the UI.
```
mchalapuk: Same here.

pkg/authorization/interceptor_test.go
```diff
@@ -57,7 +57,11 @@ func TestCanvasAuthorizationRulesSeparateDraftAndLiveActions(t *testing.T) {
 		{pbCanvases.Canvases_ListCanvasVersions_FullMethodName, "read"},
 		{pbCanvases.Canvases_DeleteCanvasVersion_FullMethodName, "update_version"},
 		{pbCanvases.Canvases_ApplyCanvasVersionChangeset_FullMethodName, "update_version"},
-		{pbCanvases.Canvases_CommitCanvasRepositoryFiles_FullMethodName, "update"},
+		{pbCanvases.Canvases_CommitCanvasRepositoryFiles_FullMethodName, "update_version"},
+		{pbCanvases.Canvases_StageCanvasRepositoryFile_FullMethodName, "update"},
+		{pbCanvases.Canvases_DiscardCanvasStaging_FullMethodName, "update"},
+		{pbCanvases.Canvases_CommitCanvasStaging_FullMethodName, "update"},
+		{pbCanvases.Canvases_ApplyCanvasAutoLayout_FullMethodName, "update"},
```
mchalapuk: All those should be "update_version".

db/migrations/20260610003157_workflow-staging.up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+
+CREATE TABLE workflow_staging (
```
mchalapuk: This needs a better name if a row represents a file.

protos/canvases.proto
```diff
@@ -610,8 +669,18 @@ message DescribeCanvasVersionRequest {
...
-  // Version metadata only; workflow content is in canvas.yaml via the repository files API.
+  // Version metadata only; workflow content is read via the repository files API
+  // (use ?stage=true on the file GET to read effective staged content).
   CanvasVersion version = 1;
+  // Populated for draft versions to drive uncommitted/ready-to-publish indicators.
+  StagingState staging_state = 2;
+}
+
+// StagingState reports the uncommitted spec edits held in workflow_staging for a draft version.
+message StagingState {
```
mchalapuk: Let's call this StagingSummary?

docs/contributing/canvas-change-requests.md
```diff
@@ -6,11 +6,13 @@ This guide describes the current canvas versioning and change-request flow in Su
...
 When effective canvas versioning is enabled:
 
-1. Update a draft version.
-2. Create a change request from that draft.
+1. Edit a draft version through the UI **stage → Commit** loop (canvas/console/files edits autosave into `workflow_staging`; **Commit** materializes them onto the draft version row) or commit directly via the CLI or agent tools.
+2. Create a change request from that committed draft.
 3. Review the change request and collect approvals.
 4. Publish the change request when approval requirements are satisfied.
 
+Uncommitted staging must be committed before proposing a change request or publishing a draft directly (when change management is disabled).
```
forestileao, line 14: We only enforce "commit first" in the UI right now. An API caller can publish or open a change request straight from the committed draft even when `workflow_staging` has newer edits, so the rule just doesn't exist on the server. Can we add a `HasWorkflowStaging` FailedPrecondition guard before publish and change-request creation so this holds for everyone, not only the UI?

Refs: `publish_canvas_version.go:134-147`, `create_canvas_change_request.go:109-116`.

pkg/public/repository_file_download.go
```diff
@@ -70,7 +71,11 @@ func (s *Server) handleRepositoryFileDownload(w http.ResponseWriter, r *http.Req
 
 	if canvases.IsRepositorySpecFilePath(path) {
 		ctx := authentication.SetUserIdInMetadata(r.Context(), user.ID.String())
-		content, readErr := canvases.ReadRepositorySpecFile(
+		readSpecFile := canvases.ReadRepositorySpecFile
+		if stage {
```
forestileao, line 75: This `stage=true` path only checks `canvases:read` and then starts reading staged content. So if staging rows are still around on a version that later becomes visible, any org reader can pass `?stage=true` and read someone else's uncommitted work. Can we gate staged reads behind the draft owner — or at least require the same permission as staging writes — instead of the plain read path?

Refs: `repository_spec_files.go:101-107`, `workflow_staging.go:204-222`.

pkg/grpc/actions/canvases/commit_canvas_staging.go
```diff
@@ -0,0 +1,197 @@
...
+		return nil, status.Errorf(codes.Internal, "failed to load staging: %v", err)
+	}
+
+	specOps, gitOps := stagedCommitOperations(rows)
+
+	// Git files and spec rows live in separate stores, so they cannot share a
+	// transaction. Commit git files first: if the git commit fails (for example
+	// on a missing repository), the request returns before any spec change is
+	// written, keeping the version row consistent with the failed commit.
+	if len(gitOps) > 0 {
```
forestileao, line 50: This can leave us half-committed. We write the git files first and then parse/apply the spec — if the spec step fails, the user sees "commit failed" but git has already moved and staging is still sitting there, so retries get messy. Can we apply/validate the spec first, or add a compensating cleanup so a failed commit doesn't strand git and the draft row out of sync?

web_src/src/pages/app/useDraftStagingActions.ts
```diff
@@ -0,0 +1,181 @@
...
+  onCanvasDraftRestoredToCommitted,
+}: UseDraftStagingActionsOptions) {
+  const queryClient = useQueryClient();
+
+  const handleCommitStaging = useCallback(async () => {
+    if (!hasEditableVersion || !activeCanvasVersionId) {
+      return;
+    }
+
+    setIsPreparingVersionAction(true);
```
forestileao: I think this can silently drop a quick Files edit. `setIsPreparingVersionAction(true)` suspends file staging (and kills its debounce timer), but `ensureVersionActionDraftReady` only waits on canvas autosaves. So if someone edits a file and immediately hits Commit, the latest file content may never reach `workflow_staging`. Can we flush/await file staging before we set the suspend flag (or before calling `commitCanvasStaging`)?

Refs: `index.tsx:5350-5358`, `useEditorStagingSync.ts:28-33`.

web_src/src/pages/app/files/useRepositoryFileStaging.ts
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    stagedPathsRef.current = new Set();
+  }, [versionId]);
+
+  useEffect(() => {
+    if (!enabled || !canvasId || !versionId) {
+      return;
+    }
+
+    const timer = setTimeout(() => {
```
forestileao, line 140: This debounce needs a generation/abort guard. Clearing the timer stops future runs, but it doesn't cancel an async run already inside `matchesCommittedRepositoryFileContent` / `mutateAsync`. A stale run can land after a newer edit or revert and clobber `stagedPathsRef` or re-stage old content. Can we tag each run with a generation and ignore completions that aren't the latest?

web_src/src/pages/app/useDraftStagingIndicators.ts
```diff
@@ -0,0 +1,284 @@
...
+  hasConsoleStagingChanges: boolean;
+  hasFilesStagingChanges: boolean;
+  draftChangeIndicators: ConsoleVersionDiff["draftChangeIndicators"];
+}) {
+  return {
+    editTabTone: draftEditTabToneFromStaging(hasStagingChanges, isEditing),
+    hasUncommittedCanvasDraftChanges: isEditing && hasCanvasStagingChanges,
+    hasUncommittedConsoleDraftChanges: isEditing && hasConsoleStagingChanges,
+    hasUncommittedFilesDraftChanges: isEditing && hasFilesStagingChanges,
+    hasCommittedCanvasDraftChanges:
```
forestileao, line 118: The committed-files path is only half-wired. `CanvasModeToggle` takes a committed prop all the way down, but this hook only returns committed flags for Canvas and Console — so after a file-only commit the blue Files dot can never light up. Can we either return `hasCommittedFilesDraftChanges` here or drop the dead committed-files plumbing so the UI state stays honest?

Refs: `useDraftStagingIndicators.ts:267-283`, `Header.tsx:90-95`, `Header.tsx:301-306`.

forestileao (commented):
A few things I want to flag before this goes in, mostly around the commit/staging rules only being enforced in the UI, plus a couple of races in the file-staging hooks. Details inline.

web_src/src/pages/app/useDraftStagingActions.ts
mchalapuk: Interesting. To be investigated.

web_src/src/pages/app/useDraftStagingIndicators.ts
mchalapuk: This dot behavior is correct. Drafts are not yet mapped to branches.  Arbitrary files are not actually part of the draft.  Currently, they are always committed to main. Next PR will change that.

pkg/grpc/actions/canvases/commit_canvas_staging.go
mchalapuk: Cleanup can also fail, but probably it's the way to go for now.

docs/contributing/canvas-change-requests.md
mchalapuk: We added commit to UI because we don't want to create commits per auto-save. API called might have a completely different request granularity. It's all right that we enable both paths.

pkg/public/repository_file_download.go
mchalapuk: We definitely need to gate behind the draft owner.

db/migrations/20260610003157_workflow-staging.up.sql
mchalapuk: Renamed the table to `workflow_staged_files` (each row is a staged file path) and added a follow-up migration to rename existing `workflow_staging` tables in dev DBs. Addressed in 06aad0705.

protos/canvases.proto
mchalapuk: Renamed the proto message and response fields to `StagingSummary` / `staging_summary`, and regenerated pb + OpenAPI + TS client. Addressed in 06aad0705.

pkg/public/repository_file_download.go
mchalapuk: Added `ensureStagedReadAllowed()` on staged spec reads and `ReadStagedRepositoryFile`, so effective staged content is only returned to the draft owner (covers the public `?stage=true` path too). Added `TestStagedReadRequiresDraftOwner`. Addressed in 06aad0705.

pkg/grpc/actions/canvases/commit_canvas_staging.go
mchalapuk: Git files are still committed first, but we now snapshot pre-commit content and attempt a compensating revert commit if spec apply fails afterward (best-effort, logged on revert failure). Addressed in 06aad0705.

web_src/src/pages/app/files/useRepositoryFileStaging.ts
mchalapuk: Debounced runs now carry a generation counter; stale async completions bail out before mutating server state or `stagedPathsRef`. Addressed in 06aad0705.

shiroyasha (approved):

forestileao (approved):

PR #5285, opened 2026-06-10
feat: List Memory Variables for HTML/Markdown Widgets

### LINKED ISSUE #5263: Support memory lists in markdown/html interpolation

andrecalil:
## Summary
Support memory list variables for HTML/Markdown widgets.

Resolves #5263


shiroyasha:
Looks good 👍

web_src/src/pages/app/console/markdownInterpolation.ts
```diff
@@ -14,11 +14,23 @@ const TEMPLATE_RE = /\{\{[\s\S]*?\}\}/;
...
+ * Authors who want a separator between elements use the `join(list, sep)`
+ * builtin explicitly. Non-array objects still fall back to JSON so a stray
+ * `{{ row }}` reference shows something inspectable instead of `[object
+ * Object]`.
+ */
 function stringifyMarkdownValue(value: unknown): string {
   if (value === null || value === undefined) return "";
   if (typeof value === "string") return value;
   if (typeof value === "number" || typeof value === "boolean") return String(value);
+  if (Array.isArray(value)) return value.map(stringifyMarkdownValue).join("");
```
forestileao: this changes stringification for every array-valued interpolation, not just list-mode memory variables. Existing panels that render array fields or {{ parseJson(...) }} used to get JSON like ["a","b"], but now silently render ab. Since this PR adds join(list, sep), please keep the default array fallback as JSON and use join(..., "") for intentional fragment concatenation.

web_src/src/pages/app/console/VariablePreview.tsx
```diff
@@ -45,10 +45,145 @@ export function VariablePreview({
...
+          {joinSnippet}
+        </button>
+      </div>
+    );
+  }
+  const first = items[0] as Record<string, unknown>;
+  return (
+    <ul className="max-h-40 space-y-0.5 overflow-y-auto pr-1">
+      {fieldKeys.map((key) => {
+        const snippet = `{{ join(${name}.map(item, item.${key}), ", ") }}`;
```
forestileao: The generated list snippet uses item.${key} for every field key. Memory fields can contain dashes, dots, spaces, or start with digits, which will generate invalid CEL or access the wrong path.

What about emitting bracket access for non-identifier keys?
 e.g. item["deploy-status"], or reuse an accessor formatter.

forestileao (approved):

PR #5288, opened 2026-06-10
feat: decouple canvas nav views and polish agent toggle

ropsii:
Move Runs and Versions into first-class header tabs with dedicated side panels, reserve the tool sidebar for the agent, and refine the mode toggle grouping and agent button styling.


forestileao:
👍

PR #5290, opened 2026-06-10
feat: enable attaching images to agent chat

### LINKED ISSUE #5136: Support adding images to agent chat
eg. while building a canvas I want to take a screenshot of an error log and paste it inside agent chat so it can fix it

re-pixel:
Demo: https://drive.google.com/file/d/1bcmEU7CWOvt3aJRHPBG02LW93rtRkLmK/view?usp=sharing

Resolves [#5136](https://github.com/superplanehq/superplane/issues/5136). Lets users paste or attach images (e.g. a screenshot of an error) in the canvas agent chat so the agent can act on them.

**Backend**
- New `AgentChatImage { media_type, data }` proto type; `images` field added to the send-message request and stored message.
- New `images` jsonb column on `agent_session_messages` (migration included).
- Images forwarded to the Anthropic provider as `image` content blocks alongside the text.
- gRPC handler validates media type (png/jpeg/gif/webp), base64, non-empty, ≤8 per message, and a combined ≤2.5 MB image payload per message (sized to keep the request under the gRPC limit); allows image-only messages.
- Chat history does **not** embed base64: message responses carry image metadata only, and bytes are served out-of-band via a new cookie-authed endpoint (`GET /api/v1/agents/chats/{chatId}/messages/{messageId}/images/{index}`, scoped to the session's owner). This keeps history responses small enough to stay under the gRPC/HTTP response size limit.

**Frontend**
- Paste images into the composer or pick them via a new attach button.
- Removable thumbnail previews before sending; transcript images load lazily from the image endpoint via `` (URL built from chat/message/index), rather than inline data.
- Paste only intercepts the clipboard when a supported, within-limit image is present, so unsupported/oversized pastes (and accompanying text) fall through normally.


re-pixel:
@lucaspin managed to align with main, would appreciate if you could take a look

lucaspin:
@forestileao can you take a look here?

protos/agents.proto
```diff
@@ -114,6 +114,17 @@ message SendAgentChatMessageRequest {
...
+  // agent alongside the text so it can reason about them.
+  repeated AgentChatImage images = 4;
+}
+
+// AgentChatImage is a base64-encoded image attached to a chat message.
+message AgentChatImage {
+  // IANA media type, e.g. "image/png" or "image/jpeg".
+  string media_type = 1;
+  // Base64-encoded image bytes, without a data URI prefix.
+  string data = 2;
```
forestileao: Will check if we cannot use bytes here

protos/agents.proto
```diff
@@ -114,6 +114,17 @@ message SendAgentChatMessageRequest {
...
   AgentMode mode = 3;
+  // Images attached to the message (e.g. a pasted screenshot). Sent to the
+  // agent alongside the text so it can reason about them.
+  repeated AgentChatImage images = 4;
+}
+
+// AgentChatImage is a base64-encoded image attached to a chat message.
+message AgentChatImage {
+  // IANA media type, e.g. "image/png" or "image/jpeg".
+  string media_type = 1;
```
forestileao: this should be an enum

forestileao (approved):

PR #5300, opened 2026-06-10
fix: keep secrets working after organization rename

0vertake:
## Summary
- Fix secrets actions to validate organization/domain IDs instead of panicking on malformed values.
- Ensure secrets UI requests pass the explicit organization ID header instead of relying on URL inference.
- Add regression coverage for secrets remaining usable after an organization rename.

## Test plan
- manual E2E test
- `make test.e2e E2E_TEST_PACKAGES=./test/e2e`


bender-rodriguez-unit1:

PR #5312, opened 2026-06-11
chore: remove legacy template canvases from database

### LINKED ISSUE #5303: Remove legacy canvas templates
Before we had apps, we used [these templates](https://github.com/superplanehq/superplane/tree/main/templates/canvases). Now, we don't use them anymore, but we still have code for them, and we still store them in the database. We should clean that up.

lucaspin:
Closes https://github.com/superplanehq/superplane/issues/5303

Next one after https://github.com/superplanehq/superplane/pull/5301. Removing legacy template canvases from the database. Since they do not have any data associated (events, executions, ...), we only need to remove the workflow_nodes and workflows records.


lucaspin:
👍

PR #5334, opened 2026-06-11
feat: good default run titles

### LINKED ISSUE #4432: Good default run titles for each built-in trigger component
For example, the `github.onPush` trigger would already come with run title set to `{{ root().data.head_commit.message }}` and if you want you can change it.

markoa:
## Summary

  - Add trigger-specific default run title expressions for built-in triggers
  - Prefill the global `customName` / Run title field from those defaults when users enable customization
  - Add a registry-backed test to require every built-in trigger to define a default run title
  - Document the convention in the component implementation checklist

  ## Notes

  This keeps the implementation centralized in `pkg/grpc/actions/run_title_defaults.go`, instead of requiring every trigger to define the global field separately.

  The first commit maps current title behavior where practical. The follow-up commit tightens a few obviously weak defaults, including static manual/schedule/webhook titles and GitHub SHA-based titles.

---

Resolves #4432


markoa:
example: when you flip the switch to set a custom title you'll how the default is made so it's easier to customize:
CleanShot 2026-06-11 at 16 17 01@2x

PR #5302, opened 2026-06-10
fix: recover agent sessions after provider stalls

forestileao:
## Summary
- return a recoverable busy error when Anthropic is waiting for tool results instead of marking the chat failed
- recreate archived/unavailable Anthropic provider sessions and update the local session external ID before retrying once
- allow failed agent sessions to recover on the next user send and show persisted failed status in the sidebar

## Validation
- make format.go && make format.js
- make test PKG_TEST_PACKAGES=./pkg/agents/...
- make test PKG_TEST_PACKAGES=./pkg/grpc/actions/agents
- make test PKG_TEST_PACKAGES=./pkg/workers
- cd web_src && npm test -- CanvasToolSidebar/index.spec.tsx --run
- make check.build.ui
- make lint && make check.build.app

PR #5335, opened 2026-06-11
fix: unstick agent sessions stuck in streaming state

forestileao:
## Summary

Two-part fix for the bug where the agent sidebar shows "Agent is running…" forever and the stop button has no effect.

The UI gates on `agent_sessions.status === "streaming"`. When the stream worker goroutine dies mid-turn (rolling deploy / SIGKILL / OOM / panic past recover / TCP cut on the SSE), nothing transitions the row off `streaming`, so the user is locked out until `FailStuckStreamingSessions` runs ~30 min later. Stop was inert because `Service.InterruptSession` posted `user.interrupt` to Anthropic but threw the success away without touching the DB.

Confirmed against the real stuck session in the report: Anthropic said `status: idle`, our row said `streaming`. Pure state divergence.

### 1. Stop button authoritatively resets local state

`InterruptSession` now:
- Calls the provider best-effort. `ErrProviderSessionUnavailable` → upstream gone, treat as already interrupted. Any other provider error → log and still proceed; the next `SendMessage` reconciles via the existing recovery paths from #5302.
- Closes any tool rows still in `started` (mirrors the worker's `closeOpenTools`; kept duplicated to avoid a `workers → agents` import cycle).
- `UpdateAgentSessionStatus(idle)` + publishes `turn_completed`/`status=idle`. The frontend's existing websocket dispatcher flips `setStatus("idle")` — no UI changes needed.

Race-safe with a live worker: bumping `updated_at` makes the worker's `UpdateAgentSessionStatusIfUnchanged(idle, turnStartedAt)` a no-op, then it re-reads, sees `status != "streaming"`, returns cleanly.

### 2. Heartbeat-based stuck detection

Worker writes `last_active_at = now()` every 30s while streaming (uses `UpdateColumn` so GORM doesn't auto-bump `updated_at` — that column is the per-turn optimistic concurrency key for `UpdateAgentSessionStatusIfUnchanged`). `FailStuckStreamingSessions` now keys on `last_active_at` instead of `updated_at`, with grace dropped from 30 min → 2 min and cleanup ticker from 5 min → 1 min. Worst-case stuck time ~3 min, covers every worker-death mode (the only signal is the absence of a heartbeat write). An initial heartbeat fires before the first tick so a backed-up queue can't let the cutoff elapse before the worker claims the session. The model fn is guarded by `status='streaming'` so a goroutine that survives past a reset can't keep an idle row looking active.

## Test plan

- [x] `make test PKG_TEST_PACKAGES=./pkg/agents/...` (90 tests; adds 4 InterruptSession scenarios)
- [x] `make test PKG_TEST_PACKAGES=./pkg/workers/...` (180 tests; updates cleanup test + adds heartbeat keeps-alive and non-streaming no-op tests)
- [x] `make test PKG_TEST_PACKAGES=./pkg/grpc/actions/agents` (existing interrupt tests untouched, still pass)
- [x] `make format.go && make lint`
- [ ] Manual: trigger an agent turn, kill the worker mid-stream, hit Stop — composer un-gates immediately and a `turn_completed`/`status=idle` WS event lands.

### DIFF
--- a/db/migrations/20260611153001_add-heartbeat-at-to-agent-sessions.up.sql
+++ b/db/migrations/20260611153001_add-heartbeat-at-to-agent-sessions.up.sql
@@ -1,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE agent_sessions ADD COLUMN heartbeat_at TIMESTAMPTZ NULL;
+
+COMMIT;
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -132,7 +132,8 @@ CREATE TABLE public.agent_sessions (
     status character varying(40) DEFAULT 'idle'::character varying NOT NULL,
     last_active_at timestamp with time zone,
     created_at timestamp with time zone DEFAULT now() NOT NULL,
-    updated_at timestamp with time zone DEFAULT now() NOT NULL
+    updated_at timestamp with time zone DEFAULT now() NOT NULL,
+    heartbeat_at timestamp with time zone
 );
 
 
@@ -2190,7 +2191,7 @@ SET row_security = off;
 --
 
 COPY public.schema_migrations (version, dirty) FROM stdin;
-20260611001815	f
+20260611153001	f
 \.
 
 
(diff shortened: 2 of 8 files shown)


forestileao:
👍

PR #5338, opened 2026-06-11
fix: GET /api/v1/canvases returns app summaries

lucaspin:
No need to return all the information for the app when listing them. Just a summary is enough.


lucaspin:
👍

PR #5343, opened 2026-06-11
feat: Introducing Coolify Connectors

### LINKED ISSUE #5342: Introduce Coolify connectors

andrecalil:
## Coolify integration

A new connector for [Coolify](https://coolify.io) self-hosted instances. Authenticates via Bearer API token against `<host>/api/v1`, validated on Sync by hitting `/version`.

**Configuration**
- `Base URL` — Coolify host (e.g. `https://coolify.example.com`)
- `API Token` — Coolify API token (sensitive)

**Actions**

| Action | What it does |
|---|---|
| List Applications | Fetches all applications from `/applications` for use in workflows or as a resource picker source |
| List Services | Fetches all services from `/services` |
| Control Application | Start / Stop / Restart a chosen application by UUID |
| Control Service | Start / Stop / Restart a chosen service by UUID |
| Deploy Application | Fire-and-forget trigger of a new deploy via `/deploy`; emits the returned deployment metadata without polling |

Resource pickers for the `Control` and `Deploy` actions are populated by `ListResources` (application + service), so users select from a live list instead of typing UUIDs.

**Frontend**
- Coolify SVG icon registered in the integration icon map
- Component mapper (`web_src/src/pages/app/mappers/coolify`) wires all actions through a shared base mapper for canvas/run presentation

### Extra
Added a small fix for list variables in memory upsert, the match wasn't processing the variable.

Resolves #5342


pkg/components/updatememory/update_memory.go
```diff
@@ -123,7 +127,7 @@ func (c *UpdateMemory) Configuration() []configuration.Field {
 			Name:        "iterateList",
 			Label:       "Input is a list",
 			Type:        configuration.FieldTypeBool,
-			Description: "When enabled, iterate over a list expression and update once per element (matchList stays global)",
+			Description: "When enabled, iterate over a list expression and update once per element. Both matchList and valueList values are evaluated per element with the iteration variable in scope (e.g. item.uuid)",
```
lucaspin, line 130: This fix should be in a separate PR

andrecalil: Please give it another go

pkg/integrations/coolify/control_application.go
```diff
@@ -0,0 +1,184 @@
...
+	spec.Application = strings.TrimSpace(spec.Application)
+	if spec.Application == "" {
+		return ControlApplicationSpec{}, fmt.Errorf("application is required")
+	}
+
+	if err := validateLifecycleOperation(spec.Operation); err != nil {
+		return ControlApplicationSpec{}, err
+	}
+
+	return spec, nil
```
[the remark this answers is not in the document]
andrecalil: The issue as described does not exist in the current code.

Both decode functions already do spec.Operation = strings.TrimSpace(spec.Operation) — they trim the operation and assign the trimmed value back to the spec field (line 146), exactly like they do for application/service (line 141). This happens before validation and before Execute reads spec.Operation.

So:

Setup with surrounding whitespace (e.g. "restart ") trims to "restart", which passes validation.
Execute builds the URL from the already-trimmed spec.Operation, producing a valid path like /api/v1/applications/{uuid}/restart.
The description's premise ("only validate operation on a trimmed copy" while Execute uses the untrimmed value) would only be true if line 146 were just if err := validateLifecycleOperation(spec.Operation); err != nil { without the assignment. The assignment is present, so the operation is consistently trimmed.

validateLifecycleOperation does also trim internally on a local copy, but that's harmless/defensive and irrelevant here since the spec field itself is already trimmed.

No code change is needed. If you'd like, I can run the package tests to confirm the whitespace-trimming behavior, or add a regression test asserting that an operation with surrounding whitespace is accepted and produces a clean URL.

lucaspin (approved):

PR #5381, opened 2026-06-12
feat: persist runner execution logs and serve history via API 

0vertake:


db/migrations/20260612131308_add-runner-execution-logs.up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS workflow_node_execution_logs (
```
lucaspin: Why are we storing these in the database? Shouldn't we use S3?

0vertake: Yes, we should use S3. I'm working on it right now, although this PR will also have to await further architecture decisions from Igor

shiroyasha:
## PR Risk Review

**Risk:** 42/100 (medium)
**Review approved:** No
**Check passed:** No

### Summary
New API endpoint proxies logs from an external broker service using environment variable credentials, with several correctness and security concerns that should be addressed before merging.

### Concerns
- Broker auth token is read from environment variables on every request via os.Getenv without caching or validation - if TASK_BROKER_AUTH_TOKEN is empty, requests are sent with an empty Bearer token which could be a security issue or cause confusing errors
- uuid.MustParse in ListNodeExecutionLogs handler will panic on invalid organization ID from context, causing a server crash instead of returning an error
- fetchLogsFromBroker fetches ALL logs from the broker every time, then applies pagination client-side - this is inefficient and could be a performance/memory issue for executions with large log volumes
- No context propagation to the HTTP request to the broker - if the gRPC request is cancelled, the broker HTTP call will continue until its 60s timeout
- The proto CanvasNodeExecutionLog message defines fields (id, canvas_id, node_id, execution_id, run_id) that are never populated by the implementation, which could confuse API consumers
- No rate limiting or circuit breaking on the outbound HTTP calls to the broker service
- Test coverage doesn't verify behavior when broker returns errors or when environment variables are missing/empty

PR #5406, opened 2026-06-12
fix: use FOR NO KEY UPDATE locking

### LINKED ISSUE #5424: Replace FOR UPDATE to FOR NO KEY UPDATE where appropriate

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/5424

Fix worker row locks to use FOR NO KEY UPDATE instead of FOR UPDATE when only non-key columns are updated. These worker locks still serialize competing workers via SKIP LOCKED, but no longer block concurrent inserts into child tables that reference the locked rows through foreign keys. This reduces PostgreSQL deadlock risk during canvas event routing, node queue processing, execution processing, and run finalization.


lucaspin:
👍

Issue #5386, opened 2026-06-12
DB: Slow query — UPDATE app_installation_secrets (avg 96ms, 222k calls)


superleomori:
Upsert on `app_installation_secrets` is averaging **96ms** and is called **222,774 times** — very high frequency.\n\n```sql\nUPDATE "app_installation_secrets" SET "organization_id"=$1,"installation_id"=$2,"name"=$3,...\n```\n\nCheck for missing indexes on lookup columns and whether this can be batched.

shiroyasha:
More info:

<img width="3622" height="2506" alt="Image" src="https://github.com/user-attachments/assets/e37e4cac-5b51-4878-a89a-7b14bf9c07f7" />

superleomori:
**Update from Lucas Pinheiro:**

> Only thing that comes to mind is a faulty integration implementation. We probably need some logs around that - if we still don't have them - and probably some metrics. Like "GitHub" integration is scheduling X number of requests, updating secrets at an X rate, …

The 222k calls to `UPDATE app_installation_secrets` is abnormally high and likely points to a bug in the integration layer — possibly a GitHub integration that is re-scheduling or re-triggering secret updates in a loop.

**Next steps:**
- Add logs around `app_installation_secrets` updates to identify which integration is triggering them and at what rate
- Add metrics: request scheduling rate per integration, secret update rate per integration
- Look for a loop/retry bug in the GitHub integration implementation

mchalapuk:
## Call paths that write to `app_installation_secrets`

Every writer funnels through one of two methods, both ending in GORM `Save()`/`Create()` on `models.IntegrationSecret`:
- `IntegrationContext.SetSecret` (`pkg/workers/contexts/integration_context.go:320`)
- `IntegrationSecretStorage.{Create,CreateMany,Update}` (`pkg/workers/contexts/integration_secret_storage.go`)

Grouped by entry point:

| # | Trigger | Frequency | Path → writer |
|---|---------|-----------|---------------|
| 1 | **Background resync timer** | **High (dominant)** | `IntegrationRequestWorker` (1s ticker) → `Sync()` → token refresh → `Integration.SetSecret` |
| 2 | OAuth callback / connect | Rare (user) | HTTP `…/callback` → `HandleCallback` → `Integration.SetSecret` + first `ScheduleResync` (bootstraps <span>#</span>1) |
[...]

PR #5435, opened 2026-06-14
feat: add per-integration secret write metrics and logging

mchalapuk:
Related to https://github.com/superplanehq/superplane/issues/5386.

Adds attribution for writes to `app_installation_secrets` to help diagnose the high-frequency upserts.

- New OTel counter `integration.secret.writes.total` labeled by `app_name`, `source`, `installation_id`, and `operation` (`create`/`update`), recorded in `IntegrationContext.SetSecret` and `IntegrationSecretStorage.{Create,Update}`.
- A structured log line per write (`app_name`, `installation_id`, `secret_name`, `source`, `operation`). The secret value is never logged.
- Threads a `source` label through `IntegrationContext` and `IntegrationSecretStorage` as a required constructor parameter (no default), set at the `sync`, `integration_action`, `execution`, `webhook`, `oauth_callback`, `setup`, `user_update`, `cleanup`, and `list_resources` entry points.

This complements the existing `db.rows.affected.count{table, operation}` (which shows raw volume) by attributing each write to a specific integration and call path, so we can confirm which integration/source drives the churn.


pkg/workers/contexts/integration_secret_storage.go
```diff
@@ -186,5 +194,28 @@ func (s *IntegrationSecretStorage) Update(name string, value string) error {
...
+}
+
+// recordSecretWrite emits the metric and structured log for a write to
+// app_installation_secrets. The secret value is never logged.
+func (s *IntegrationSecretStorage) recordSecretWrite(name, operation string) {
+	telemetry.RecordIntegrationSecretWrite(
+		context.Background(),
+		s.integration.AppName,
+		s.source,
+		s.integration.ID.String(),
```
lucaspin: Should we use integration ID here? Thinking about high cardinality issues

shiroyasha (approved):

PR #5441, opened 2026-06-14
chore: remove child executions APIs + cleanup UI logic

### LINKED ISSUE #5431: Remove /children API endpoint from canvases.proto

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/5431

Next one after https://github.com/superplanehq/superplane/pull/5429. No more blueprints means we don't have child executions anymore. Removing the /children API + cleaning up the child executions logic from the UI.


lucaspin:
👍

PR #5454, opened 2026-06-14
fix(canvas): recover gracefully when a draft is deleted out from under the UI

### LINKED ISSUE #5464: Canvas publish/commit breaks with "version not found" when a draft is deleted
## Problem
When a canvas draft no longer exists — deleted by the agent, another browser tab, or the user — publishing or committing against it returns a **"version not found"** (404 / gRPC `NOT_FOUND`) error with no recovery. The UI keeps a stale draft version id in refs/URL, the canvas/console commit state desyncs, and every retry re-fails, leaving the editor stuck.

## Expected
The UI should detect that the draft is gone and recover gracefully — return to the live canvas, clear the stale version id and URL params, and show an informational message — instead of stranding the user.

## Notes
The agent backend already self-heals (it re-resolves or recreates the draft per tool call); this is a frontend resilience gap.

Reported during agent testing (canvas committed → publish state desync; "discard"/"publish" actions breaking).

forestileao:
Closes #5464

## What
Publishing or committing against a draft that no longer exists — deleted by the agent, another browser tab, or the user — returned a **"version not found"** 404 with no recovery, leaving the canvas/console commit state desynced and every retry failing.

## Changes
- **`ensureDraftVersionExists`** (`useCanvasData.ts`) — refetches draft branches to validate a draft id before commit/publish.
- **`isNotFoundError`** (`workflowPageHelpers.ts`) — generic 404/NOT_FOUND detection; `isCanvasLoadNotFoundError` delegates to it.
- **`useDraftRecovery`** (new) — owns the publish + exit-to-live + recovery lifecycle; publish-success and missing-draft recovery share one path (clear stale ref/spec/URL params → return to live).
- Commit/reset (`useDraftStagingActions`) and the agent draft opener (`useAgentDraftEditor`) detect NotFound and recover/prune the dead id instead of erroring.

Pre-check + a catch branch cover the delete-between-check-and-mutation race. A recovered miss shows an informational toast, not the fatal red error.

## Tests
`useCanvasData.spec.ts` — `ensureDraftVersionExists` present / absent / missing-args.

## Notes
Frontend-only. The agent backend already self-heals (re-resolves/recreates the draft per tool call) — this fixes the UI side that stranded on a stale id.

### DIFF
--- a/web_src/src/hooks/useCanvasData.spec.ts
+++ b/web_src/src/hooks/useCanvasData.spec.ts
@@ -11,6 +11,7 @@ const {
   canvasesCommitCanvasStaging,
   canvasesDiscardCanvasStaging,
   canvasesDescribeCanvasVersion,
+  canvasesListCanvasVersions,
 } = vi.hoisted(() => ({
   canvasFoldersUpdateCanvasFolder: vi.fn(),
   canvasesListRuns: vi.fn(),
@@ -18,6 +19,7 @@ const {
   canvasesCommitCanvasStaging: vi.fn(),
   canvasesDiscardCanvasStaging: vi.fn(),
   canvasesDescribeCanvasVersion: vi.fn(),
+  canvasesListCanvasVersions: vi.fn(),
 }));
 
 vi.mock("../api-client/sdk.gen", async (importOriginal) => {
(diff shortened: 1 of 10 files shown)


forestileao:
👍

PR #5455, opened 2026-06-14
feat(agent): surface async agent activity in the chat

### LINKED ISSUE #5465: Agent chat: async activity is invisible and recoverable errors are dropped
## Problem
The agent chat only exposes a binary `streaming`/`idle` state, so users can't tell what the agent is actually doing in the background ("a lot of async things happening that are not visible"). Two concrete gaps:

- **Recoverable provider errors are silently dropped.** `session.error` events from the managed-agents stream are discarded, so a transient hiccup the agent recovers from leaves no trace for the user.
- **Canvas edits look like generic commands.** When the agent edits the canvas via the `update_draft` tool, the transcript just says "Running command", giving no signal that the draft is being modified.

## Expected
- Surface recoverable `session.error` events as a non-fatal, informational notice (distinct from a hard failure), cleared on the next turn / new message.
- Label the agent's `update_draft` tool calls as "Editing canvas" when an edit is the tool actually running.

Reported during agent testing.

forestileao:
Closes #5465

## What
Agent activity was a binary `streaming`/`idle` state, and recoverable provider errors (`session.error`) were silently dropped — users couldn't tell what the agent was doing or when it hit a transient hiccup.

## Changes
**Backend**
- New non-terminal `ProviderEventSessionNotice`. `session.error` (previously dropped in `anthropic/events.go`) now maps to it; `session.status_terminated` stays the only terminal path.
- The stream worker publishes a `session_notice` event **without** changing session status, persisting a DB row, or stopping the stream.

**Frontend**
- `session_notice` flows through the WS hook to a transient **amber** inline notice, distinct from the fatal red error, auto-cleared on the next turn boundary.
- The agent's `superplane_app` `update_draft` tool calls now render as **"Editing canvas"** instead of "Running command".

## Tests
- `anthropic_test.go` — `session.error` emits one `session_notice` and the stream continues.
- `useAgentSessionWebsocket.spec.ts` — `session_notice` dispatches to `onNotice` without touching status.

## Notes
Backend Go build/tests couldn't run in my environment because the branch's generated protobuf/swagger is stale (pre-existing, unrelated) — run `make gen` then `go test ./pkg/agents/... ./pkg/workers/...` to exercise the new mapping.

### DIFF
--- a/pkg/agents/anthropic/anthropic_test.go
+++ b/pkg/agents/anthropic/anthropic_test.go
@@ -825,10 +825,14 @@ func TestStreamEvents_SessionErrorDoesNotStopStream(t *testing.T) {
 		return nil
 	}))
 
-	require.Len(t, received, 2)
-	assert.Equal(t, agents.ProviderEventAssistantMessage, received[0].Type)
-	assert.Equal(t, "Recovered", received[0].Text)
-	assert.Equal(t, agents.ProviderEventTurnCompleted, received[1].Type)
+	// session.error surfaces as a non-terminal notice; the stream keeps going
+	// and the recovered assistant message + turn completion still arrive.
+	require.Len(t, received, 3)
+	assert.Equal(t, agents.ProviderEventSessionNotice, received[0].Type)
+	assert.Equal(t, "An internal service error occurred", received[0].ErrorMessage)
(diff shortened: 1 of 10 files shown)


forestileao:
👍

PR #5458, opened 2026-06-14
fix: canvas edit-mode staging bugs (cross-tab sync, diff badge, commit flash, changeManagement diff)

### LINKED ISSUE #5450: Edit mode: [Reset][Commit] re-enabled briefly after commit
**Description:**
In canvas edit mode, the secondary action controls toggle between two mutually exclusive states: `[Reset] [Commit]` while there are staged (uncommitted) changes, and `[Discard] [Publish]` once everything is committed. After clicking **Commit**, there is a window where the `[Reset] [Commit]` controls are re-enabled but still showing — instead of immediately switching to `[Discard] [Publish]`. This is normally hard to notice, but becomes clearly visible when the app is under high load / slow.

**Steps to reproduce:**
1. Enter edit mode on a canvas and make a change so staged changes exist (`[Reset] [Commit]` shown).
2. Click **Commit**. The controls are correctly disabled while the request is in flight.
3. Observe the controls right after the request completes (easiest to reproduce when the app is slow / under load).

**Expected / Actual:**
Expected: Once the commit request completes, the controls should never show an enabled `[Reset] [Commit]`. They should go straight from the disabled/pending state to `[Discard] [Publish]`.
Actual: After the request completes the controls are re-enabled but still show `[Reset] [Commit]` for a few seconds, then switch to `[Discard] [Publish]`.

**Likely cause:**
The toggle is driven by two independent state sources that update at different times:
- `commitStagingPending` (`isPreparingVersionAction || commitCanvasStagingMutation.isPending`) — controls the disabled/spinner state, and is cleared in the `finally` block of `handleCommitStaging` as soon as the mutation resolves (`web_src/src/pages/app/useDraftStagingActions.ts`).
- `hasStagingChanges` — decides whether `[Reset] [Commit]` (`showStagingActions`) or `[Discard] [Publish]` is rendered, and is recomputed asynchronously from staging indicators/baselines/queries (`web_src/src/pages/app/useDraftStagingIndicators.ts`, `resolveEditingStagingFlags`).

When the commit completes, `commitStagingPending` flips to `false` (re-enabling the buttons) before `hasStagingChanges` flips to `false`, so `EditModePublishDiscardActions` (`web_src/src/ui/CanvasPage/HeaderSecondaryActions.tsx`) still renders the enabled `[Reset] [Commit]` branch until the staging indicators catch up. The gap is amplified under load.

A fix should keep the controls in the disabled/pending state (or otherwise avoid showing an enabled `[Reset] [Commit]`) until the staging state has actually flipped to committed, so the two sources never disagree.

### LINKED ISSUE #5451: Canvas diff badge ignores uncommitted changes
**Description:**
On the Canvas page (edit mode), the diff summary badge — the `+added ±updated -removed` indicator that opens the "Diff X-Ray" / "See Full Diff" hover card (`DiffSummaryHoverCard`) — does not appear when the only changes are uncommitted (staged) edits. As soon as there are staged-but-not-yet-committed changes, the badge should reflect them, but it shows nothing until the changes are committed into a draft version.

**Steps to reproduce:**
1. Enter edit mode on a canvas.
2. Make an edit (add/update/remove a node) so there are staged changes, but do **not** click Commit.
3. Look at the canvas header diff indicator.

**Expected / Actual:**
Expected: The diff badge shows the counts for the uncommitted (staged) changes, so users can see and open the diff for work they haven't committed yet.
Actual: No diff badge is shown while changes are only staged/uncommitted. The badge only appears once the changes are committed (i.e. once there is a committed draft that differs from live).

**Likely cause:**
The diff badge is rendered in `web_src/src/ui/CanvasPage/HeaderSecondaryActions.tsx`:

```tsx
{onCanvasTab && hasUnpublishedDraftChanges && draftVisualDiff?.diffCounts ? (
  <DiffSummaryHoverCard diffCounts={draftVisualDiff.diffCounts} ... />
) : null}
```

The same pattern gates the console badge on `hasUnpublishedConsoleDraftChanges`. These `hasUnpublished*DraftChanges` flags come from `getDraftChangeIndicators` (`web_src/src/pages/app/lib/version-action-state.ts`), which is based on `hasDraftDiffVersusLive` — i.e. the **committed** draft compared to live. They do not account for uncommitted staged changes (`hasStagingChanges` / `hasUncommitted*DraftChanges`).

Meanwhile `draftVisualDiff.diffCounts` (`web_src/src/pages/app/useDraftVisualDiff.ts`) is computed from the rendered draft nodes/edges vs live, so it would already reflect uncommitted edits — it's just hidden because the surrounding visibility condition is committed-only.

The fix should make the badge visibility include uncommitted/staged changes (e.g. gate on `hasUncommittedCanvasDraftChanges`/`hasUncommittedConsoleDraftChanges` in addition to the committed flags, or on a combined "has any draft changes" condition), so the diff indicator shows up whenever there is a non-empty diff regardless of commit state.

### LINKED ISSUE #5453: Uncommitted draft changes don't sync across tabs
**Description:**
When the same draft is open in two browser tabs (same user, same canvas/draft version), uncommitted (staged) changes made in one tab are not synced to the other tab in real time. The second tab keeps showing the stale draft state until it is manually reloaded or some other action triggers a refetch.

**Steps to reproduce:**
1. Open a canvas draft in edit mode in two browser tabs (Tab A and Tab B).
2. In Tab A, make an edit so there are staged/uncommitted changes (do not commit).
3. Switch to Tab B without reloading.

**Expected / Actual:**
Expected: Tab B reflects the uncommitted changes from Tab A reasonably quickly (the staging layer is shared on the server, so both tabs should converge).
Actual: Tab B does not show the uncommitted changes. It stays on the old draft state until a manual reload or an unrelated refetch occurs. (Once the change is committed, the committed draft does propagate.)

**Likely cause:**
Uncommitted changes are persisted to a per-version server-side staging layer and surfaced locally via the `canvasKeys.versionStaging` React Query cache (`useCanvasVersionStaging` in `web_src/src/hooks/useCanvasData.ts`). The tab that makes the edit updates its own query cache / server staging, but there is no cross-tab propagation:

- The canvas websocket (`web_src/src/hooks/useCanvasWebsocket.ts`) only handles committed/lifecycle events — `canvas_updated`, `canvas_version_updated`, `canvas_deleted`, `repository_branch_updated`. There is no staging-updated event, so staging mutations in one tab never reach other tabs.
- There is no `BroadcastChannel` / `storage`-event mechanism syncing staging state between tabs in the same browser either.

As a result the staging layer is shared on the server, but other tabs are never told to refetch it. A fix would need a way to notify other tabs/sessions that staging changed (e.g. emit a staging-updated realtime event and invalidate `canvasKeys.versionStaging` + the draft spec, and/or a same-browser BroadcastChannel) so open tabs converge on the latest uncommitted draft state.

### LINKED ISSUE #5456: Weird  changeManagement diff artifacts on Canvas tab
Image

mchalapuk:
## Summary

Fixes a cluster of canvas edit-mode staging bugs, plus a related diff-rendering artifact.

Fixes #5453.

**Uncommitted changes don't sync across tabs.** Staging writes (canvas/console/file edits, discard, auto-layout) persist server-side but emitted no realtime event, so other tabs editing the same draft never refreshed. The backend now publishes a `canvas-staging-updated` event on every staging write and broadcasts it as a `staging_updated` websocket event. The client refreshes the staged caches and re-applies the effective staged spec to the active draft's rendered graph. A per-tab echo guard (`canvasStagingEcho.ts`) ignores the originating tab's own broadcast so it doesn't clobber in-flight local edits, and tabs with their own unsaved edits defer the refresh.

Fixes #5450.

**`[Reset][Commit]` briefly re-enabled after commit.** The `versionStaging` query is now optimistically cleared on commit success so `hasStagingChanges` flips false immediately instead of lagging behind the commit-pending flag. The staging controls stay disabled until staging settles (the lingering commit spinner is dropped in favor of the disabled-control signal).

Fixes #5451.

**Diff badge ignores uncommitted staged changes.** The canvas/console diff badges are now gated on the uncommitted staged-change flags in addition to the committed unpublished-change flags, so staged edits surface a diff badge before they're committed.

Fixes #5456.

**Weird `changeManagement` diff artifacts on Canvas tab.** The live canvas spec from the API carries proto3 default fields (`enabled: false`, empty `userId`/`roleName`) while the locally-edited draft omits them, producing spurious `changeManagement` edits in the diff. `buildCanvasYamlFromWorkflow` now canonicalizes change management before serialization (dropping falsy `enabled` and empty approver fields) so only real change-management edits appear.

## Changes

**Backend (Go)**
- `pkg/grpc/actions/messages/canvas.go`: `CanvasStagingUpdatedRoutingKey` + `PublishStagingUpdated()`.
- `pkg/workers/eventdistributer/canvas.go`: `staging_updated` event + `HandleCanvasStagingUpdated`.
- `pkg/workers/event_distributer.go`: register the new routing key.
- Publish the event after a successful staging write in `workflow_staging.go`, `discard_canvas_staging.go`, and `apply_canvas_auto_layout.go` (commit is already covered by `canvas_version_updated`).

**Frontend (TS)**
- `useCanvasWebsocket.ts`: handle `staging_updated`, expose `onCanvasStagingEvent` callback, invalidate staged caches.
- `index.tsx`: wire `onCanvasStagingEvent` to resync the active draft with staging echo suppression.
- `useCanvasData.ts`: optimistically clear `versionStaging` on commit success.
- `HeaderSecondaryActions.tsx`: gate diff badges on uncommitted flags; keep commit controls disabled until staging settles.
- `canvas-yaml-staging.ts`: canonicalize change management before YAML serialization.
- New `lib/canvasStagingEcho.ts` echo guard.

## Test plan

- [x] Open the same canvas draft in two browser tabs (A and B) in edit mode. Make a staged change in A (move a node / edit a component / discard / auto-layout). Confirm B reflects the change without a manual refresh, and that A's own in-flight local edits are not clobbered by the echo.
- [x] With staged changes, throttle the network (DevTools → Slow 3G) and click Commit. Confirm the `[Reset][Commit]` controls never flash back to an enabled state during the post-commit refetch window — they stay disabled until staging clears.
- [x] Make a staged (uncommitted) edit on both the canvas and the console. Confirm the diff badge appears before committing, and that the hover card shows the correct counts.
- [x] Open a canvas whose `changeManagement` is disabled/default and enter edit mode. Confirm the diff view shows no spurious `changeManagement` (`enabled`/`userId`/`roleName`) artifacts. Then enable change management and confirm a real `enabled: true` change still appears in the diff.

### DIFF
--- a/web_src/src/hooks/useCanvasWebsocket.ts
+++ b/web_src/src/hooks/useCanvasWebsocket.ts
@@ -28,6 +28,8 @@ type CanvasWebsocketPayload = {
 
 type CanvasLifecycleEventName = "canvas_updated" | "canvas_version_updated" | "canvas_deleted";
 
+type CanvasStagingEventName = "staging_updated";
+
 type WebsocketPayload =
   | CanvasesCanvasNodeExecution
   | CanvasesCanvasEvent
@@ -53,6 +55,7 @@ export function useCanvasWebsocket(
   shouldApplyCanvasUpdate?: () => boolean,
   processRuntimeEvents = true,
   enabled = true,
+  onCanvasStagingEvent?: (payload: CanvasWebsocketPayload, eventName: CanvasStagingEventName) => boolean | void,
 ): void {
   const nodeExecutionStore = useNodeExecutionStore();
   const queryClient = useQueryClient();
@@ -113,7 +116,10 @@ export function useCanvasWebsocket(
       const payload = data.payload;
       const isCanvasLifecycleEvent =
         data.event === "canvas_updated" || data.event === "canvas_version_updated" || data.event === "canvas_deleted";
-      if (!isCanvasLifecycleEvent && !processRuntimeEvents) {
+      // Staging events fire while editing a draft (not the live version), so they
+      // must bypass the runtime-event gate that is disabled outside the live view.
+      const isCanvasStagingEvent = data.event === "staging_updated";
+      if (!isCanvasLifecycleEvent && !isCanvasStagingEvent && !processRuntimeEvents) {
         return;
       }
 
@@ -225,6 +231,39 @@ export function useCanvasWebsocket(
           queryClient.invalidateQueries({ queryKey: canvasKeys.list(organizationId) });
           break;
         }
+        case "staging_updated": {
+          // A draft's staging layer changed in another tab (or this one). Refresh
+          // the staged caches so the diff badge, console and files tabs reflect
+          // the uncommitted changes.
+          const stagingMessage = payload as Partial<CanvasWebsocketPayload>;
+          if (!stagingMessage.canvasId || stagingMessage.canvasId !== canvasId || !stagingMessage.versionId) {
+            break;
+          }
+
+          const shouldInvalidateStagingQueries =
+            onCanvasStagingEvent?.(stagingMessage as CanvasWebsocketPayload, "staging_updated") !== false;
+          if (!shouldInvalidateStagingQueries) {
+            break;
+          }
+
+          const versionId = stagingMessage.versionId;
+          queryClient.invalidateQueries({ queryKey: canvasKeys.versionStaging(canvasId, versionId) });
+          // versionStagedDetail, consoleStaged and staged repositoryFileContent
+          // keys all end with "staged" and include the version id, so a single
+          // predicate refreshes the editor's effective draft reads without
+          // touching the committed caches.
+          queryClient.invalidateQueries({
+            predicate: (query) => {
+              const key = query.queryKey;
+              return (
+                Array.isArray(key) &&
+                key[key.length - 1] === "staged" &&
+                (key as readonly unknown[]).includes(versionId)
+              );
+            },
+          });
+          break;
+        }
         default:
           break;
       }
@@ -237,6 +276,7 @@ export function useCanvasWebsocket(
       onWorkflowEvent,
       onExecutionEvent,
       onCanvasLifecycleEvent,
+      onCanvasStagingEvent,
       shouldApplyCanvasUpdate,
       processRuntimeEvents,
       organizationId,
--- a/web_src/src/hooks/useCanvasWebsocket.ts
+++ b/web_src/src/hooks/useCanvasWebsocket.ts
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useRef } from "react";
 import useWebSocket from "react-use-websocket";
-import { useQueryClient, type InfiniteData } from "@tanstack/react-query";
+import { useQueryClient, type InfiniteData, type QueryClient } from "@tanstack/react-query";
 import type {
   CanvasesCanvasNodeExecution,
   CanvasesCanvasEvent,
@@ -43,6 +43,20 @@ interface QueuedMessage {
     payload: WebsocketPayload;
   };
   timestamp: number;
+}
+
+// Refreshes the staged caches for a draft version. versionStagedDetail,
+// consoleStaged and staged repositoryFileContent keys all end with "staged" and
+// include the version id, so a single predicate refreshes the editor's
+// effective draft reads without touching the committed caches.
+function invalidateStagedCanvasQueries(queryClient: QueryClient, canvasId: string, versionId: string): void {
+  queryClient.invalidateQueries({ queryKey: canvasKeys.versionStaging(canvasId, versionId) });
+  queryClient.invalidateQueries({
+    predicate: (query) => {
+      const key = query.queryKey;
+      return Array.isArray(key) && key[key.length - 1] === "staged" && (key as readonly unknown[]).includes(versionId);
+    },
+  });
 }
 
 export function useCanvasWebsocket(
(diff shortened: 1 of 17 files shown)


web_src/src/hooks/useCanvasWebsocket.ts
```diff
@@ -43,6 +45,20 @@ interface QueuedMessage {
...
+function invalidateStagedCanvasQueries(queryClient: QueryClient, canvasId: string, versionId: string): void {
+  queryClient.invalidateQueries({ queryKey: canvasKeys.versionStaging(canvasId, versionId) });
+  queryClient.invalidateQueries({
+    predicate: (query) => {
+      const key = query.queryKey;
+      return Array.isArray(key) && key[key.length - 1] === "staged" && (key as readonly unknown[]).includes(versionId);
+    },
+  });
+}
+
```
forestileao, line 61: the amount of requests frotnend will make because of the invalidation is okay?

mchalapuk: Yes. This is react query mechanism which caches the yaml files so that we're not requesting the same contents multiple times.

That said, it must be invalidated sometimes, otherwise it would never fetch a new version.

andrecalil (approved):
Only thing I noticed is that `canvasStagingEcho` could have a unit test, as it seems rather key

forestileao (approved):

PR #5463, opened 2026-06-14
fix: decouple integration sync HTTP calls from the DB transaction

mchalapuk:
## Summary

Reduces the latency of the slow, extremely frequent `UPDATE app_installation_secrets` query by no longer holding a DB transaction open across external HTTP calls in the integration sync worker.

Related to #5386.

## Root cause

Token-based integrations (GCP WIF, AWS STS, GitLab, PagerDuty, ServiceNow, Azure, …) refresh their credentials on a schedule: each `Sync` re-stores the secret(s) and reschedules itself, so the same secret rows are rewritten on a tight, self-perpetuating loop.

The problem reported in #5386 was the **average latency per `UPDATE` (~96ms)** — a single-row update by primary key should be sub-millisecond. The latency was inflated because `LockAndProcessRequest` wrapped the entire `Sync`/`HandleHook` (external HTTP **and** the secret writes) in a single transaction. Concretely, for an integration like GCP the `SetSecret` `UPDATE` runs and then several more external HTTP calls execute before commit, so:

- the secret row stays write-locked across the network round-trips, and any concurrent writer to that row blocks (and that lock-wait is counted into the `UPDATE`'s time); and
- the long-open transactions hold back Postgres's xmin horizon, so autovacuum can't reclaim the dead tuples generated by the churn → table/index/TOAST bloat → every `UPDATE` touches more pages and gets slower.

## What was done

Reworked `IntegrationRequestWorker` to a 3-phase flow so the external HTTP call never runs inside a DB transaction:

- **Phase 1 (short tx):** claim the request.
- **Phase 2 (no tx):** run `Sync`/`HandleHook` with a non-transactional context (`database.Conn()`) and `registry.HTTPContext()`.
- **Phase 3 (short tx):** persist instance state and mark the request `completed`.

### Claiming via a `run_at` lease (visibility timeout)

Because Phase 2 runs without a row lock, we need another way to stop the 1s poll loop from dispatching the same request again while it is in flight. Rather than introduce a separate `processing` state plus a reaper to recover crashed workers, claiming uses a lease on the existing `run_at` column (the SQS visibility-timeout pattern):

- `LeaseIntegrationRequest(tx, id, lease)` locks the row with `SKIP LOCKED`, gated on `state = 'pending' AND run_at <= now`, then pushes `run_at` to `now + lease`.
- The poll (`ListIntegrationRequests`) already filters `state = 'pending' AND run_at <= now`, so a leased request (future `run_at`) drops out automatically — no new state.
- On success, Phase 3 marks the request `completed`.
- **If the worker dies mid-process**, nothing completes the request; once the lease expires (`run_at <= now`) it becomes due again and the normal poll re-picks it — automatic recovery, restart-independent, with no separate reset mechanism.

The lease duration (`requestLeaseDuration`, 5m) is both the worst-case recovery latency for a crashed request and a lower bound that must exceed the longest expected `Sync`/`HandleHook`, so a request still legitimately in flight is never re-leased.

No migration: `state` stays `varchar`, no new state value is introduced, and the existing partial index `(state, run_at) WHERE state = 'pending'` already serves the poll query.


pkg/workers/integration_request_worker.go
```diff
@@ -39,7 +39,21 @@ func NewIntegrationRequestWorker(encryptor crypto.Encryptor, registry *registry.
...
+// stuckProcessingTimeout is how long a request may stay in the "processing"
+// state before the reaper assumes the worker that claimed it died and resets
+// it back to "pending". It must be larger than the longest expected Sync/HandleHook
+// so we never reset a request that is still legitimately in flight on another replica.
+const stuckProcessingTimeout = 15 * time.Minute
+
 func (w *IntegrationRequestWorker) Start(ctx context.Context) {
+	// On startup, reset any requests stuck in "processing" state from a previous
+	// crash back to "pending" so they get retried.
+	if count, err := models.ResetStuckProcessingIntegrationRequests(stuckProcessingTimeout); err != nil {
```
lucaspin: Does this mean we only reset the ones stuck on a new deployment?

mchalapuk: A new deployment, a manual restart, or a crash-and-relaunch.

mchalapuk: A request only ends up orphaned in processing if the worker dies mid-HTTP-call crash/OOM/kill which means we'll restart anyway.

lucaspin:
Also, how often are these integrations re-syncing? If they are scheduling too much work, that could also be a problem, no?

mchalapuk:
> Also, how often are these integrations re-syncing? If they are scheduling too much work, that could also be a problem, no?

Yes — each token-based integration runs on its own loop, typically rescheduling at about half the token lifetime (30–60 min for most providers; Azure is fixed at 30 min). That does create steady background work (HTTP + writes), and with many integrations it adds up.

We didn’t change cadence in this PR on purpose: the main problem was the cost per sync — redundant writes due to random AES-GCM nonces, plus holding a DB transaction across the external HTTP call. This fix targets that.

If write volume is still high after deploy (we have new metrics to measure that), cadence tuning (e.g. refresh closer to expiry instead of at half-life, or raising minimum intervals) would be a reasonable follow-up once we have post-deploy metrics/logs.

pkg/workers/integration_request_worker.go
lucaspin: What about new deployments terminating workers that claimed a request, but didn't fully process it?

1. Worker claims a request A -> request A moves to processing
2. New deployment -> pod receives SIGTERM
3. Worker is not gracefully shutting down, so pod terminates before request A is processed
4. Request A is only gonna be re-processed when pod crashes or when new deployment happens

pkg/workers/integration_request_worker.go
```diff
@@ -71,31 +79,65 @@ func (w *IntegrationRequestWorker) Start(ctx context.Context) {
...
+			return err
 		}
 
-		return w.processRequest(tx, r)
+		claimed = r
+		return nil
 	})
+
+	if err != nil {
+		w.log("Request %s already being processed - skipping", request.ID)
```
mchalapuk: We're not checking what error we got. Are we sure this is the only thing that can happen?

mchalapuk (commented):
One question:

pkg/workers/integration_request_worker.go
mchalapuk: Good catch — not the only thing that can happen. `ErrRecordNotFound` covers another worker having leased it / the row no longer being due or pending. Update failures and transient DB errors are real failures and should not be logged as "already being processed".

Narrowed the skip to `errors.Is(err, gorm.ErrRecordNotFound)` and propagate everything else in b546d00cc.

pkg/workers/integration_request_worker.go
mchalapuk: Implemented a lease mechanism using `run_at` column. No reset mechanism needed.

lucaspin (approved):

Issue #5462, opened 2026-06-14
DB: N+1 query in SerializeCanvasVersion — FindMaybeDeletedUserByID called per version


superleomori:
## Problem

`SerializeCanvasVersion` calls `models.FindMaybeDeletedUserByID` for each version being serialized. When serializing a list of versions, this results in an N+1 DB query pattern — one extra query per version to resolve the owner name.

```go
func SerializeCanvasVersion(version *models.CanvasVersion, organizationID string) *pb.CanvasVersion {
    var owner *pb.UserRef
    if version.OwnerID != nil {
        ownerID := version.OwnerID.String()
        ownerName := ""
        if user, err := models.FindMaybeDeletedUserByID(organizationID, ownerID); err == nil && user != nil {
            ownerName = user.Name
        }
        owner = &pb.UserRef{Id: ownerID, Name: ownerName}
    }
    ...
}
```

## Impact

- `ListVersions` and any other endpoint serializing multiple `CanvasVersion` records fires N extra DB queries
- Contributes directly to the high query volume and slow response times observed on `ListVersions` (~500ms+)

## Fix

Pre-load user data before serialization — resolve all unique owner IDs in a single batch query, then pass the resolved map into the serializer instead of querying per-item.

PR #5477, opened 2026-06-15
fix: Eliminate N+1 queries for canvas version listing

shiroyasha:


shiroyasha:
fixes https://github.com/superplanehq/superplane/issues/5462

Issue #5425, opened 2026-06-13
Stability: MaybeFinalizeRunInTransaction can lead to deadlocks — needs redesign


superleomori:
## Problem

`MaybeFinalizeRunInTransaction` has a design that can lead to deadlocks under concurrent execution. Run finalization needs to be redesigned to be safe under concurrency.

## Reported by

Lucas Pinheiro

## Next Steps

- Review `MaybeFinalizeRunInTransaction` for lock ordering issues, competing transactions, or unsafe concurrent access patterns
- Redesign the finalization logic to be deadlock-safe (e.g. consistent lock ordering, optimistic locking, or moving finalization out of the transaction entirely)

PR #5479, opened 2026-06-15
fix: skip run finalization check if not needed

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/5425

No need to check if run is finished - locking it and blocking other transactions from creating new events, executions, queue items - if we just produced records (pending events or queue items) that would result in a non-finished run after the check. This is just a quick optimization to see how the numbers behave. Run finalization is still not great.


lucaspin:
👍

PR #5485, opened 2026-06-15
fix: use extended log entry when logging errors in EventRouter

lucaspin:
We should use the logger created with `logging.ForEvent(w.logger, event)` when logging errors in EventRouter


lucaspin:
👍

PR #5488, opened 2026-06-15
chore: Lint snake_case component configuration fields

### LINKED ISSUE #5247: Linter: warn when configuration fields use snake_case instead of camelCase
## Context

Configuration fields in canvas YAML should use camelCase, but it is easy to accidentally write snake_case (e.g. `my_field` instead of `myField`). There is currently no tooling to catch this.

## Request

Add a linter rule (CLI or CI check) that:
- Scans canvas YAML configuration fields
- Warns or errors when a field name uses snake_case instead of camelCase

## Why

- Prevents silent misconfiguration (snake_case fields may be silently ignored or cause unexpected behavior)
- Enforces a consistent naming convention across canvases
- Catchable at `superplane canvases update` time or in CI before a canvas is applied

shiroyasha:
## Summary
- Adds a linter that scans registered component, trigger, widget, and integration **configuration field definitions** for snake_case names
- Runs in CI via `make check.configuration.fields`
- Grandfathers 50 existing violations in `.configuration-fields-baseline.txt` (mostly runner components)
- Closes #5247

This targets Go `configuration.Field` names when building components and integrations — not canvas YAML user config.

New snake_case field names fail CI unless the baseline is updated. Each issue suggests the camelCase equivalent (e.g. `execution_mode` → `executionMode`).

## Test plan
- [ ] `make test PKG_TEST_PACKAGES=./pkg/lint/configurationfields` passes
- [ ] `make check.configuration.fields` passes
- [ ] Adding a new snake_case field to a component `Configuration()` fails CI
- [ ] Existing runner fields remain allowed via baseline until migrated


.configuration-fields-baseline.txt
```diff
@@ -0,0 +1,51 @@
+# Configuration fields on components, triggers, widgets, and integrations must use camelCase.
+action:display:configuration[2]:color_expression
```
lucaspin, line 2: This is not related to this, but leaving a comment here, since it's where I saw it for the first time: a weird that we append *expression on some field names too. No need to do that, expressions should be usable mostly everywhere. No need to append that in that configuration field name

shiroyasha:
Agreed — the `*expression` suffix on field names is a separate cleanup. This PR only flags snake_case; happy to track that naming separately.

PR #5492, opened 2026-06-15
fix: suppress stale agent draft auto-open toast

forestileao:
## Summary
- Keep the DraftActionsWidget publish flow unchanged.
- Suppress the stale agent-draft unavailable toast when background auto-open finds a draft was already published.
- Add regression coverage for the auto-open race.

## Validation
- npm run test:run -- src/pages/app/useAgentDraftEditor.spec.tsx
- make check.build.ui


forestileao:
👍

PR #5497, opened 2026-06-15
feat: add Anthropic agent memory stores

forestileao:
## Summary
- add org-scoped experimental feature flag support for Anthropic managed-agent memory stores
- persist user+canvas scoped memory-store mappings and attach them to new/recovered provider sessions
- keep memory provisioning fallback-safe, avoid provider calls inside DB transactions, and prepare mappings for existing sessions
- fix Anthropic memory-store creation endpoint to `/memory_stores`

## Tests
- `make format.go`
- `make test PKG_TEST_PACKAGES="./pkg/agents ./pkg/agents/anthropic ./pkg/models ./pkg/features"`
- `make lint`
- `make check.build.app`

### DIFF
--- a/pkg/agents/service.go
+++ b/pkg/agents/service.go
@@ -7,22 +7,26 @@ import (
 	"fmt"
 	"hash/fnv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/superplanehq/superplane/pkg/authorization"
 	"github.com/superplanehq/superplane/pkg/database"
+	"github.com/superplanehq/superplane/pkg/features"
 	"github.com/superplanehq/superplane/pkg/grpc/actions/messages"
 	"github.com/superplanehq/superplane/pkg/models"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 var ErrSessionForbidden = errors.New("agent session is owned by another user")
 
 type Service struct {
-	provider Provider
-	auth     authorization.Authorization
+	provider         Provider
+	auth             authorization.Authorization
+	memoryStoreLocks sync.Map
 }
 
 func NewService(provider Provider, auth authorization.Authorization) *Service {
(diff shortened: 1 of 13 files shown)


pkg/agents/service.go
```diff
@@ -397,6 +460,228 @@ func (s *Service) installRecoveredProviderSession(stale *models.AgentSession, pr
 	return recovered, nil
 }
 
+const (
+	agentMemoryStoreDescription  = "Durable SuperPlane agent memory for one user and one canvas. Store user preferences, project conventions, useful lessons, and prior mistakes. Do not store secrets, credentials, raw prompts, or large canvas YAML snapshots."
```
mchalapuk: shouldn't this say "one app"?

pkg/agents/service.go
```diff
@@ -397,6 +460,228 @@ func (s *Service) installRecoveredProviderSession(stale *models.AgentSession, pr
...
+	agentMemoryStoreInstructions = "Use this memory only for durable lessons, user preferences, app-specific conventions, and prior mistakes that will help future SuperPlane agent sessions for this same user and canvas. Do not store secrets, credentials, raw user prompts, API tokens, or large canvas YAML snapshots."
```
mchalapuk: And what draft are you working with?

mchalapuk: Are you sure about credentials and API tokens?

mchalapuk (commented):
I think we need specific instructions for the agent about using memory to store draft id somewhere in the system prompt.

forestileao:
Addressed Maciej's memory prompt feedback in 3587fb15: added a Durable Memory Store section to the system prompt, changed scope wording to one user + one app, instructed the agent to store the latest private draft/version ID returned by superplane_app as a hint, and clarified that secret values/credentials/API tokens must not be stored.

forestileao:
Closing this for now. After reviewing the current session model, memory stores do not directly solve the draft-selection problem: we keep one SuperPlane agent session per user + app, so Claude's managed-agent session already carries normal working context. The draft concern is better handled in the tool contract by letting update_draft receive a draft/version ID and validating it server-side. I am replacing this with a smaller targeted change around draft selection.

PR #5503, opened 2026-06-15
fix: no need to publish execution message in EventRouter

lucaspin:
Remove legacy CanvasExecutionMessage publishing from EventRouter — it only existed for blueprint parent completion, which was removed last week; for normal workflows it duplicated what NodeExecutor already publishes


lucaspin:
👍

PR #5507, opened 2026-06-16
fix: use dedicated executions exchange

### LINKED ISSUE #5506: Use a separate exchange for executions
We are using the canvas-exchange for everything (events, executions, runs, organization-related events, queue items, notifications), using separate routing keys for each event type. That is weird, and doesn't allow workers to subscribe only to the specific events they need. For example, NodeExecutor only needs to listen to `execution.pending` events, but it gets all messages, and has to load the execution and check if it is finished. More efficient to just use an executions exchange, and subscribe to `execution.pending` events instead.

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/5506


lucaspin:
👍

Issue #5377, opened 2026-06-12
Processing: deadlock detected (SQLSTATE 40P01) issues


lucaspin:
See [this Discord message](https://discord.com/channels/1409914582239023200/1513515878480875540/1514744297012986167) for more details.

lucaspin:
This is probably caused by https://github.com/superplanehq/superplane/issues/5425

lucaspin:
No deadlocks since the deployment of https://github.com/superplanehq/superplane/pull/5508:

PR #5508, opened 2026-06-16
fix: move run finalization to RunFinalizer worker

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/5377

The way we are finalizing runs can lead to deadlocks, as multiple workers can lock the workflow_run in different order, and in different transactions. Here, we move the run finalization logic to a new RunFinalizer worker.

### Other changes

RunFinalizer needs to react to terminal events - events that produced no queue items. Taking the opportunity here to also move the RabbitMQ events related to app events to a dedicated exchange, while adding the new event.terminal event.


AleksandarCole:
/docs-agree

@lucaspin ignore the next comments - testing my SP app.

lucaspin:
👍

Issue #5514, opened 2026-06-16
Perf: Authorization tail latency — p99 still hitting 300-400ms despite p50 at ~1ms


superleomori:
## Problem

Authorization p50 is comfortably at ~1ms, but the **p99 can still spike to 300-400ms**. This tail latency adds directly to the worst-case response time for every authenticated request.

## Impact

- Users at the p99 experience 300-400ms of latency just from auth — before any business logic runs
- Makes it hard to hit overall p99 targets for API endpoints

## Investigation

- Identify what causes the auth tail — likely intermittent DB latency, cache misses, or lock contention on permission lookups
- Check if the p99 spikes correlate with DB load spikes or specific permission check paths
- Consider adding a per-request auth cache to eliminate repeat lookups within a request lifecycle

## Related

- #5470 (Reduce authorization duration — overall auth optimization)

PR #5516, opened 2026-06-16
fix: cache allowed organization permission checks in auth interceptor

shiroyasha:
## Summary
- Add `organization_permissions.go` with the same cache semantics as experimental features: trust cached **allowed** for 15 minutes; reload from Casbin on miss or cached denied.
- Wire the gRPC authorization interceptor through `checkOrganizationPermission` instead of calling `CheckOrganizationPermission` directly.
- Keep `AuthService.CheckOrganizationPermission` uncached so role-management tests and other callers still hit Casbin immediately.

## Test plan
- [x] `make test PKG_TEST_PACKAGES=./pkg/authorization/...`
- [ ] Confirm `auth.check_permission` Casbin SQL spans drop on repeated canvas/org requests for the same user


shiroyasha:
related https://github.com/superplanehq/superplane/issues/5514

PR #5517, opened 2026-06-16
fix: stop runaway DockerHub token-refresh secret-write loop (#5386)

mchalapuk:
## Summary

Fixes the runaway `UPDATE app_installation_secrets` load reported in #5386 (~35k writes/day, avg ~96ms, dominating Cloud SQL Query Insights). The DockerHub token-refresh loop was leaking self-perpetuating refresh chains; this PR switches it to the deduplicated scheduler, hardens the request worker so a request can never duplicate on failure/restart, and adds gauges for early detection.

## Root cause

DockerHub drives a perpetual token refresh. It rescheduled itself via `ScheduleActionCall`, which — unlike `ScheduleResync` — has **no de-duplication**: it creates a new pending request without ever completing the current one. So every `Sync` (which runs on integration sync/edit/capability update *and* on the recurring refresh) permanently added a brand-new, never-terminating refresh chain.

```
Sync -> ScheduleActionCall("refreshAccessToken")   # creates B, does NOT complete A
HandleHook(A) -> refresh + ScheduleActionCall       # creates B', does NOT complete A
...chains multiply, none ever die
```

This was amplified by the worker's phasing: the successor request was created/committed in phase 2 (the external `Sync`/`HandleHook` call), but the parent was completed in a **separate** phase-3 transaction. If phase 3 failed or the worker died in between, the leased parent was retried after its lease expired and spawned a *duplicate* successor.

Each chain refreshes the token shortly before its JWT `exp`, and since all chains for an installation share the same token lifetime, their `run_at` is phase-locked into a tight window — producing synchronized ~9-minute write bursts. `SetSecret` always does a full-row `Save()` with a fresh AES-GCM nonce, so every refresh is a real, heavy UPDATE.

## How we reached the conclusion

- Cloud SQL Query Insights pinpointed `UPDATE app_installation_secrets` as the hot query.
- Structured logs (`source=integration_action`, `Integration secret write`) tied the writes to one DockerHub installation (`f55dd252-…`) on a recurring background cadence — not user activity — and showed the ~9-minute synchronized bursts.
- Direct DB inspection of `app_installation_requests` for that installation: **202 pending + 1.14M completed** `invoke-action` rows, all `refreshAccessToken`, with the 202 pending clustered into a ~2-minute `run_at` window. `202 chains × (~1440 min/day ÷ ~9 min) ≈ 32k UPDATEs/day`, matching the observed ~35k/day.
- Code review confirmed the `ScheduleActionCall` (no dedup) vs `ScheduleResync` (dedup via `completeCurrentRequestForInstallation`) distinction as the mechanism, and the phase-2/phase-3 split as the multiplier.

## How we confirmed it with a test

`Test__DockerHubRefreshLoopDoesNotAccumulate` (added in the preceding commit) drives the **real** DockerHub `Sync` against a DB-backed integration context, mocking only HTTP, and asserts the invariant the fix must hold: **at most one pending integration request per installation**. Against the unfixed code it **failed** — the pending count climbed from 1 to 2 after the second sync — reproducing the production accumulation in a unit test.

## The fix (and how the test confirms it)

- **DockerHub** (`pkg/integrations/dockerhub/dockerhub.go`): `Sync` and the `refreshAccessToken` hook now reschedule via `ScheduleResync`, which completes the current pending request before creating its successor. The hook is retained as a migration bridge so the 202 existing orphaned chains collapse onto the single deduplicated sync loop after deploy — code-only remediation, no manual SQL.
- **Worker hardening** (`pkg/workers/contexts/integration_context.go`, `pkg/workers/integration_request_worker.go`): the in-flight request is now carried on `IntegrationContext`, and `ScheduleResync`/`ScheduleActionCall` complete the parent **and** create the successor in one transaction. Phase-3 completion stays as an idempotent fallback for paths that schedule nothing (e.g. the Sync error path). This removes the duplicate-on-failure window.
- **Observability** (`pkg/telemetry/metrics.go`, `pkg/telemetry/periodic.go`): adds `app_installation_requests.pending.count` and `app_installation_requests.pending.max_per_installation` gauges. The max-per-installation is the real early warning — one runaway installation is invisible in a global total.

With the fix, the same `Test__DockerHubRefreshLoopDoesNotAccumulate` now **passes**: repeated syncs keep exactly one pending refresh request. This is the designed "fail before, pass after" guard against regressions.

Additional tests:
- `Test__DockerHubSyncSchedulesResyncRequest` — `Sync` schedules a `sync`-type (deduplicated) request, not an action call.
- `Test__DockerHubRefreshHookBridgesToResync` — an orphaned `refreshAccessToken` action request collapses onto a single sync request when it runs.
- `Test__IntegrationRequestWorker_NoDuplicateChainOnRetry` — reprocessing an already-completed leased request creates no duplicate chain (validates the atomic completion).

## Test plan

- [x] `Test__DockerHubRefreshLoopDoesNotAccumulate` fails on the pre-fix code and passes with the fix.
- [x] `pkg/workers`, `pkg/integrations/dockerhub`, `pkg/telemetry` suites pass (157 tests).
- [x] `make format.go`, `make lint` pass; changed packages build cleanly.
- [ ] After deploy: watch the new gauges and the `Integration secret write` log rate fall as the 202 chains collapse to 1; confirm via the pending-count for `f55dd252-…` and that `UPDATE app_installation_secrets` returns to baseline in Query Insights.

### DIFF
--- a/pkg/integrations/dockerhub/dockerhub.go
+++ b/pkg/integrations/dockerhub/dockerhub.go
@@ -102,8 +102,14 @@ func (d *DockerHub) Sync(ctx core.SyncContext) error {
 		return fmt.Errorf("failed to validate credentials: %w", err)
 	}
 
-	err = ctx.Integration.ScheduleActionCall("refreshAccessToken", map[string]any{}, *refreshIn)
-	if err != nil {
+	//
+	// Schedule the next refresh through ScheduleResync, which completes the
+	// current pending request before creating the next one. This keeps a single
+	// pending request per installation. Using ScheduleActionCall here instead
+	// would never complete the prior request, so every sync would leak another
+	// self-perpetuating refresh chain (issue #5386).
+	//
+	if err := ctx.Integration.ScheduleResync(*refreshIn); err != nil {
 		return fmt.Errorf("failed to schedule token refresh: %w", err)
 	}
 
@@ -131,12 +137,20 @@ func (d *DockerHub) Hooks() []core.Hook {
 func (d *DockerHub) HandleHook(ctx core.IntegrationHookContext) error {
 	switch ctx.Name {
 	case "refreshAccessToken":
+		//
+		// Backwards-compatibility bridge for issue #5386: the refresh loop used
+		// to be driven by self-rescheduling "refreshAccessToken" action calls,
+		// which accumulated into many orphaned chains. Any such in-flight action
+		// request is converted here onto the deduplicated resync loop (ScheduleResync
+		// completes the current request before scheduling the next), so the extra
+		// chains collapse to a single pending request over the following cycles.
+		//
 		refreshIn, err := refreshAccessToken(ctx.HTTP, ctx.Integration)
 		if err != nil {
 			return fmt.Errorf("failed to refresh access token: %w", err)
 		}
 
-		return ctx.Integration.ScheduleActionCall("refreshAccessToken", map[string]any{}, *refreshIn)
+		return ctx.Integration.ScheduleResync(*refreshIn)
 
 	default:
 		return fmt.Errorf("unknown hook: %s", ctx.Name)
--- a/pkg/telemetry/metrics.go
+++ b/pkg/telemetry/metrics.go
@@ -55,6 +55,9 @@ var (
 
 	pendingEventsGauge     metric.Int64Gauge
 	pendingExecutionsGauge metric.Int64Gauge
+
+	pendingIntegrationRequestsGauge                   metric.Int64Gauge
+	pendingIntegrationRequestsMaxPerInstallationGauge metric.Int64Gauge
 )
 
 // Operation values for integration secret writes.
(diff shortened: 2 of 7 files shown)


pkg/integrations/dockerhub/dockerhub.go
```diff
@@ -131,12 +137,20 @@ func (d *DockerHub) Hooks() []core.Hook {
...
+		// completes the current request before scheduling the next), so the extra
+		// chains collapse to a single pending request over the following cycles.
+		//
 		refreshIn, err := refreshAccessToken(ctx.HTTP, ctx.Integration)
 		if err != nil {
 			return fmt.Errorf("failed to refresh access token: %w", err)
 		}
 
-		return ctx.Integration.ScheduleActionCall("refreshAccessToken", map[string]any{}, *refreshIn)
+		return ctx.Integration.ScheduleResync(*refreshIn)
```
lucaspin, line 153: We shouldn't use Sync() for this. We should use hooks only for these background tasks. I think the reason why this was written like this was because integrations didn't have hooks yet, so the re-sync mechanism was added.

PR #5520, opened 2026-06-16
fix: react to execution.finished messages in node queue worker

### LINKED ISSUE #5388: DB: High-frequency query — SELECT DISTINCT workflow_nodes (2.6ms avg, 1.8M calls)
SELECT DISTINCT on `workflow_nodes` is called **1,816,841 times**. Even at 2.6ms avg, total load is significant.\n\n```sql\nSELECT DISTINCT "workflow_nodes"."workflow_id","workflow_nodes"."node_id","workflow_nodes"."parent_node_id",...\n```\n\nInvestigate whether results can be cached or the call frequency reduced.

lucaspin:
Fixes https://github.com/superplanehq/superplane/issues/5388

NodeQueueWorker polls the database every 1s. That is a lot of load on the DB. We can fix that by making the worker listen to execution.finished messages as well as the queue item created messages it already listens to:
- On queue item message: when a node is ready, create execution for the new queue item. Helpful when node has nothing in queue, nothing executing, and just got a new queue item.
- On execution.finished message: when a queue item is created, the node might already be executing something. If that happens, the node queue worker ignores the message, and right now, the poller is responsible for taking care of processing the next queue item when the node becomes ready quickly. If we react to the execution.finished message, we can react quickly without needing a short polling interval.

### DIFF
--- a/pkg/workers/node_queue_worker.go
+++ b/pkg/workers/node_queue_worker.go
@@ -32,17 +32,28 @@ type NodeQueueWorker struct {
 	semaphore   *semaphore.Weighted
 	logger      *log.Entry
 
-	rabbitMQURL string
-	consumer    *tackle.Consumer
+	rabbitMQURL               string
+	queueItemConsumer         *tackle.Consumer
+	executionFinishedConsumer *tackle.Consumer
 }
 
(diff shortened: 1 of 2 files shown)


lucaspin:
👍

Issue #5521, opened 2026-06-16
Remove /canvases/:id/events endpoint


lucaspin:
Now that we have /runs endpoint, we do not need this endpoint anymore.

PR #5522, opened 2026-06-16
chore: do not use /canvases/:id/events endpoint in UI anymore

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/5521

First step before removing that endpoint is updating the UI to use only the /runs endpoint.


lucaspin:
👍

PR #5537, opened 2026-06-17
fix: dedup ScheduleActionCall to stop integration secret-write loop (#5386)

### LINKED ISSUE #5386: DB: Slow query — UPDATE app_installation_secrets (avg 96ms, 222k calls)

mchalapuk:
Fixes #5386.

`IntegrationContext.ScheduleActionCall` was a plain insert with no de-duplication, so any self-rescheduling action accumulated a new chain on every run. DockerHub's `refreshAccessToken` loop grew to hundreds of orphaned chains in production, driving ~32k `UPDATE app_installation_secrets`/day.

Fix in the primitive: complete all matching pending invoke-action requests and create exactly one successor in a single transaction, keyed by `(installation, action, parameters jsonb)` so legitimately distinct calls (e.g. AWS `provisionRule` per detail type) are preserved. DockerHub needs no change and its orphaned chains self-drain.

The complete+create is serialized with a `FOR UPDATE` lock on the installation row, because the worker processes due requests in parallel. Without it, two simultaneous refreshes each complete-then-insert under their own READ COMMITTED snapshot, neither sees the other's new successor, and the chains never drain.

### Automated
- [x] Accumulation repro (fail-before / pass-after), integration-agnostic via the dummy integration: an integration whose `Sync` reschedules an action keeps exactly one pending request across N re-syncs.
- [x] Primitive dedup (`pkg/workers/contexts`): same `(action, params)` → 1 pending; distinct params → 2 (preserved); accumulated chains collapse to 1; rejects sub-second interval.
- [x] Concurrency: 8 goroutines rescheduling the same action converge to a single pending request — fails without the per-installation lock (leaves 8), passes with it.
- [x] No-duplicate-on-retry: reprocessing an already-completed leased request via the worker creates no second chain.

### Manual
- [x] On local dev env: seed duplicate pending `refreshAccessToken` rows in the DB; the worker collapses them back to 1.


pkg/workers/integration_request_worker_dockerhub_test.go
```diff
@@ -0,0 +1,188 @@
...
+//
+// DockerHub drives its token refresh through ScheduleActionCall("refreshAccessToken").
+// Re-running Sync (which happens on every integration sync/edit/capability update,
+// and on the recurring refresh) must not leave behind an extra scheduled refresh -
+// the loop must keep at most one pending request per installation.
+//
+// Before the ScheduleActionCall de-duplication fix this FAILS (the pending count
+// climbs past 1, as it did in production with 202 orphaned chains). After the fix
+// it PASSES, with the single pending request being the refreshAccessToken hook.
+func Test__DockerHubRefreshLoopDoesNotAccumulate(t *testing.T) {
```
lucaspin: I don't think we should have a integration-specific test in pkg/workers. This test should be more generic and test the behavior itself, not the dockerhub integration. We can use the dummy integration implementation we have in test/support to simulate the behavior

mchalapuk: Yes, it turned out to be scoped more widely than dockerhug in the end. The test is now integration-agenosic.

pkg/workers/contexts/integration_context.go
```diff
@@ -199,7 +199,33 @@ func (c *IntegrationContext) ScheduleActionCall(actionName string, parameters an
...
+	// different calls (e.g. AWS provisionRule for different detail types) are not
+	// affected.
+	//
+	// The installation row is locked first to serialize concurrent scheduling for
+	// it. The worker processes due requests in parallel, so without this lock two
+	// simultaneous refresh requests would each complete-then-insert under their own
+	// READ COMMITTED snapshot, neither seeing the other's new successor - leaving
+	// multiple chains that share a run_at and never drain.
+	//
+	return c.tx.Transaction(func(tx *gorm.DB) error {
```
lucaspin, line 218: Do we need the nested transaction? Why not use c.tx directly?

pkg/models/integration_request.go
```diff
@@ -116,3 +117,28 @@ func (r *IntegrationRequest) Complete(tx *gorm.DB) error {
...
+func CompletePendingActionRequestsInTransaction(tx *gorm.DB, installationID uuid.UUID, actionName string, parameters any) error {
+	params, err := json.Marshal(parameters)
+	if err != nil {
+		return err
+	}
+
+	return tx.Model(&IntegrationRequest{}).
+		Where("app_installation_id = ? AND state = ? AND type = ?",
+			installationID, IntegrationRequestStatePending, IntegrationRequestTypeInvokeAction).
+		Where("spec->'invoke_action'->>'action_name' = ?", actionName).
```
lucaspin, line 143: An update with this where clause makes me uneasy. Since the number of requests for an integration will be limited after this fix - but the number of records could be big - it may be better to find all by app_installation/state/type, and then complete them with `UPDATE app_installation_requests set ... where id in (...)`

pkg/workers/integration_request_worker_dedup_test.go
lucaspin, line 1: Why create a separate test file, and not use a sub-test in integration_request_worker_test.go?

pkg/workers/contexts/integration_context.go
mchalapuk: We do need it here. In the worker path `c.tx` is `database.Conn()` (autocommit, not an open transaction), so without the explicit `Transaction(...)` each statement would commit on its own and the `FOR UPDATE` lock on the installation row would be released right after the SELECT — defeating the per-installation serialization that stops two concurrent refreshes from each inserting a successor. The lock has to be held across complete+create.

When `c.tx` is already a transaction (the gRPC paths), `Transaction(...)` just opens a savepoint, which is harmless.

pkg/models/integration_request.go
mchalapuk: Good call — changed it to select the matching ids first (narrowed by the indexed `app_installation_id`/`state`/`type` columns) and then complete them with `UPDATE ... WHERE id IN (...)`, so the write is by primary key and only touches the few matched rows. Done in eb47d87.

pkg/workers/integration_request_worker_dedup_test.go
mchalapuk: Moved both tests (and the helper) into `integration_request_worker_test.go` and dropped the separate file. Done in eb47d87.

lucaspin (approved):

PR #5413, opened 2026-06-13
poc: git backend

mchalapuk:
## Summary

Migrates the canvas draft / commit / publish backend from DB-native storage to **git-backed storage**, treating git as the source of truth and `workflow_versions` as a materialized projection of git branches and commits. This is Phase <span>#</span>5 of the Git Source-of-Truth effort.

Closes #5330

### Git as source of truth
- Drafts are real **git draft branches**. Per-draft edits are held in `workflow_staged_files` (with `base_head_sha`) and materialized into version rows on commit.
- Change requests publish via **git merge into `main`**, then the live version is materialized from the resulting commit.
- `workflow_versions.id` stays a UUID and existing RPC names are preserved (`CreateCanvasVersion`, `PublishCanvasVersion`, etc.); drafts remain `workflow_versions` rows.

### Materialization (`pkg/canvas/materialize`)
- New seed / sync / backfill pipeline: seed repositories and `canvas.yaml` / `console.yaml`, sync live and draft branches into DB projections, reconcile deleted draft branches, and backfill existing pre-git-first canvases (draft rows → real git branches).
- New `repository_materializer` worker and a `repository_branch` event distributer; new `RepositoryBranchUpdated` message so the UI reacts to branch updates.

### Git provider
- Extended the provider interface and in-memory / codestorage / supergit implementations (branch create/merge/head, commit with expected-head, etc.).

### Change requests
- Consolidated CR diff / conflict / refresh logic into `pkg/canvas/changerequests`.

### Cleanup
- Removed the deprecated DB-native write paths now superseded by git-first: the `ApplyCanvasVersionChangeset` / `ValidateCanvasVersionChangeset` RPCs (proto, service methods, authorization entries) and the internal `UpdateCanvasVersion(WithUsage)` / `UpdateConsole` handlers, plus their tests.
- Relocated still-used helpers out of the deleted files and removed the now-dead client-side AI changeset apply path in the UI.

### Migration
- `20260612120000_git-first-canvas-versions`: adds git-first columns to `workflow_versions` / `workflow_staged_files` and a `repository_materialization_state` table.

## Test plan

- [x] create canvas → edit draft → commit → publish; works as before
- [x] Modify README.md → uncommited changes → commit → committedd to branch/draft; README.md in published version unchanged; blue dot next to Files tab → publish
- [x] Files Diff always compare against live version

PR #5538, opened 2026-06-17
feat: add git-first canvas version model and migration

mchalapuk:
Data layer for git-backed canvas versions, split out of #5413 as a prerequisite for the git-first materializer worker.

## Changes
- Migration: git-first columns on `workflow_versions` (`commit_sha`, `git_branch`, `materialization_status`, `materialization_error`) + indexes, `base_head_sha` on `workflow_staged_files`, and a new `repository_materialization_state` table.
- Models: new fields/constants and query/upsert helpers on `CanvasVersion`, the new `RepositoryMaterializationState` model, and `BaseHeadSHA` threaded through workflow staging.
- `UpsertWorkflowStagingPath` / `MarkWorkflowStagingPathDeleted` gain a `baseHeadSHA` parameter; existing callers pass `""` to preserve behavior until the worker and API land.

Additive and behavior-preserving; the new model functions are unused until the follow-up worker PR.

## Test plan

- [x] Manual check: Created a draft, deleted draft, modified canvas and console, published


db/migrations/20260617143500_git-first-canvas-versions.up.sql
```diff
@@ -0,0 +1,42 @@
+BEGIN;
+
+-- Git-first canvas versions: commit SHA tracking, materialization state, staging base head.
+
+ALTER TABLE workflow_versions
+  ADD COLUMN IF NOT EXISTS commit_sha varchar(40) NOT NULL DEFAULT '',
+  ADD COLUMN IF NOT EXISTS git_branch text NOT NULL DEFAULT 'main',
+  ADD COLUMN IF NOT EXISTS materialization_status varchar(32) NOT NULL DEFAULT 'ready',
```
lucaspin, line 8: We have materialization_status + materialization_error here, and have status + error in repository_materialization_state. Is this the same state?

I'd even consider maybe not having repository_materialization_state table at all:
- Is repository_materialization_state.branch == workflow_versions.git_branch?
- Is repository_materialization_state.head_sha === workflow_versions.commit_sha?

Might be easier to have all this information in workflow_versions table directly, instead of using a separate table

mchalapuk: Cursor, is it possible to move materialized_sha to workflow_versions and completely remove the repository_materialization_state table?

mchalapuk: We're not having it.

db/migrations/20260617143500_git-first-canvas-versions.up.sql
```diff
@@ -0,0 +1,27 @@
...
+  ADD COLUMN IF NOT EXISTS materialization_error text NOT NULL DEFAULT '';
+
+ALTER TABLE workflow_versions
+  ALTER COLUMN git_branch DROP DEFAULT,
+  ALTER COLUMN materialization_status DROP DEFAULT,
```
lucaspin, line 13: You are adding the columns above, and altering them here to drop default. Why not just add them without a default?

mchalapuk: We're adding a default first so that postgres back-fills new columns with the default value (default is required when adding new non-null columns to an existing table). Then, we're dropping the default because we don't want this mechanism for new rows.

lucaspin (approved):

PR #5545, opened 2026-06-17
feat: scale metrics

### LINKED ISSUE #5501: Observability: add basic scale/growth metrics dashboard
## Request

We need a dashboard with basic scale and growth metrics to understand how the product is being used. Currently missing:

## Metrics Needed

**Scale (totals)**
- Total users
- Total organizations
- Total apps/installations
- Total workflow integrations

**Activity (daily active)**
- Daily active apps

**Usage (daily)**
- Daily workflow runs
- Daily executions
- Daily events processed

## Why

Without these, we cannot answer basic questions about growth, usage patterns, or capacity planning.

## Source

These can likely be derived from existing DB tables — no new instrumentation needed, just queries + a dashboard in Dash0 or a dedicated metrics endpoint.

mchalapuk:
- Emit periodic OTLP gauges for installation scale totals (organizations, users, workflows, nodes, drafts, integrations, integration secrets)
- Track daily active workflows from DB activity in the last 24 hours
- Add daily workflow usage gauges for runs, events, and node executions via direct DB counts that match production SQL

Closes #5501

This is a second attempts at this after the first increased DB load too much and was reverted. Changes compared to previous one:
- [x] Decreased frequency of database queries (once an hours instead of once a minute).
- [ ] ~~A new DB indexes on `workflow_node_executions.created_at`, `workflow_events.created_at`, `workflow_runs.created_at`~~.

## Dashboard metrics

- **Scale (totals)**: `organizations.total`, `users.total`, `workflows.total`, `workflow_nodes.total`, `drafts.total`, `integrations.total`, `integration_secrets.total`.
- **Activity (24h)**: `workflows.active.count`.
- **Usage (24h)**: `workflow_runs.daily.count`, `workflow_events.daily.count`, `workflow_node_executions.daily.count`.

## Follow-Ups
- Wire Dash0 panels to the new gauge names (replace `db_rows_affected_count_total` queries)
- Consider follow-up migration for `created_at` indexes on workflow tables if periodic query load is noticeable


pkg/telemetry/metrics.go
```diff
@@ -702,3 +868,7 @@ func RecordPendingExecutionsCount(ctx context.Context, count int64) {
 
 	pendingExecutionsGauge.Record(ctx, count)
 }
+
+// Scale-total and daily-usage gauges are observable instruments collected
```
shiroyasha: Why did you add these comments to the end of the file 400 lines detached from where we are logging?

pkg/telemetry/metrics.go
```diff
@@ -68,6 +68,11 @@ const (
 	IntegrationSecretOperationUpdate = "update"
 )
 
+// dailyMetricsInterval is how often the daily-usage gauges are collected and
+// exported. Their backing DB queries run on this cadence via observable
+// callbacks, so both the query execution and the OTLP export happen hourly.
+const dailyMetricsInterval = time.Hour
```
shiroyasha: The name of the variable alludes that this will be collected once per day. But that is not the case here?

pkg/telemetry/metrics.go
```diff
@@ -426,6 +455,143 @@ func StartPeriodicMetricsReporter() {
 	p.Start()
 }
 
+// registerScaleMetrics registers the installation scale-total gauges as
```
shiroyasha: Remove these comments, they don't add value.

pkg/telemetry/metrics.go
mchalapuk: Good catch — removed it. The observable gauges are registered in `registerScaleMetrics`/`registerDailyUsageMetrics`, so this detached note added no value.

pkg/telemetry/metrics.go
mchalapuk: Right, it's collected hourly. Renamed the constant to `hourlyMetricsInterval` and updated its comment/references.

pkg/telemetry/metrics.go
mchalapuk: Removed these (and the matching comments on `observeCount`, `registerDailyUsageMetrics`, and `observeDailyCount`).

shiroyasha:
Reviewed this PR again. 👀

The main problem is what the PR Risk Review app identified.

> Some queries may become expensive at scale (e.g., COUNT DISTINCT on workflow_runs with a time filter, full table scans on workflow_events/workflow_node_executions without guaranteed indexes on created_at + workflow_id).

This could become a ticking bomb in the application, slowing it down over time.
Instead of this, counters could be a good alternative.

Closing this PR for now.

PR #5549, opened 2026-06-17
fix: rebuild casbin model per read enforcer to avoid shared-model races

### LINKED ISSUE #5548: Casbin read enforcers copy a live, concurrently-mutated model
## Bug

`AuthService.newReadEnforcer` (`pkg/authorization/service.go`) builds each read enforcer with `a.casbinModel.Copy()`. That `casbinModel` is the same `model.Model` owned by the long-lived `SyncedEnforcer`.

Read enforcers are created on essentially every permission check / role lookup, while write paths (`CreateGroup`, `AssignRole`, `AddUserToGroup`, custom role CRUD, etc.) mutate the shared model's `Policy`/`PolicyMap` through the synced enforcer. `Copy()` iterates those same maps without synchronization, so a copy can run concurrently with a write to the same map.

## Consequences

- **Process crash:** concurrent map iteration (in `Copy()`) and map write (from the synced enforcer) trigger Go's `fatal error: concurrent map iteration and map write`, which is unrecoverable and takes down the API server.
- **Data races / inconsistent authorization:** torn reads of the shared model can yield read enforcers with partial/inconsistent policy state, leading to flaky and incorrect allow/deny decisions.
- Failures are load- and timing-dependent, so they surface intermittently under concurrency rather than deterministically.

## Fix direction

Don't derive read-enforcer models from the live shared model. Build a fresh model from the immutable model definition (stored as text at startup) per read enforcer, and load policies from the DB via `LoadFilteredPolicy`. This decouples read enforcers from the mutating synced model entirely.

mchalapuk:
Read enforcers were built with `a.casbinModel.Copy()`, copying the same model owned by the long-lived `SyncedEnforcer`. Because read enforcers are created on nearly every permission check while write paths mutate that shared model's policy maps, `Copy()` could iterate a map being concurrently written — causing `fatal error: concurrent map iteration and map write` and inconsistent authorization decisions.

This stores the casbin model definition as text and rebuilds a pristine model from it per read enforcer (policies are loaded from the DB), decoupling read enforcers from the mutating synced model.

Closes #5548

## Test plan

- [x] Authorization checks behave as before (allow/deny unchanged)

### DIFF
--- a/pkg/authorization/service.go
+++ b/pkg/authorization/service.go
@@ -27,8 +27,12 @@ const (
 var _ Authorization = (*AuthService)(nil)
 
 type AuthService struct {
-	enforcer           *casbin.SyncedEnforcer
-	casbinModel        model.Model
+	enforcer *casbin.SyncedEnforcer
+	// We keep the casbin model as text rather than a model.Model so each read
+	// enforcer can rebuild a pristine model from it. The text holds only the
+	// model definition (no policies), and rebuilding from it avoids sharing a
+	// live, concurrently-mutated model across enforcers.
+	casbinModelText    string
 	orgPolicyTemplates [][5]string
 }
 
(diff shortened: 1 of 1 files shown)


mchalapuk:
Closing as we don't know how to reproduce this problem. Seems like an accidental change made in https://github.com/superplanehq/superplane/pull/5413

PR #5552, opened 2026-06-17
feat: CLI commands for runs

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/5521. Next one after https://github.com/superplanehq/superplane/pull/5522.

### CLI

Replace `superplane events` with `superplane runs`, which consumes the new /runs API instead of the old /events one. Two commands:
- superplane runs list - for listing runs
- superplane runs describe - for describing a run

### API changes

New `GET /runs/:id` endpoint for describing a run. This is a simple enough endpoint - it should have been already added as part of the runs effort - that I am including it in the same pull request. This will be used in the UI too: for example, if you go to /apps/{appId}?run={runId} in the UI, using an old run ID, UI will make list requests until finding the one we need.


lucaspin:
👍

PR #5554, opened 2026-06-17
fix: defer trigger run title resolution

forestileao:
## Summary
- skip `customName` during pre-emit trigger hook configuration resolution for worker and gRPC/API trigger invocation paths
- keep run title templates resolved from `EventContext.Emit`, where `root()`, `previous()`, and trigger node references have the emitted trigger payload
- add a worker regression for schedule run titles using `root().data.calendar.minute`
- expand run-title E2E coverage for static titles, `root()`, `previous()`, and `$["Start"]` node-reference expressions through the real manual-run trigger path

## Testing
- `make test PKG_TEST_PACKAGES='./pkg/workers ./pkg/grpc/actions/canvases'`
- `make lint`
- `make check.build.app`
- `make test.e2e.single FILE=test/e2e/trigger_run_title_test.go LINE=15`


forestileao:
👍

PR #5559, opened 2026-06-17
fix: load run inspection view via DescribeRun endpoint

lucaspin:
When opening a canvas with ?run={runId}, resolve the selected run with DescribeRun instead of paging through ListRuns until a match appears. This fixes direct links to runs that are not in the first list page and removes unnecessary list pagination during run inspection. Stale ?run= URLs are cleared only after DescribeRun returns not found. Websocket run updates also keep the describe-run cache in sync.


lucaspin:
👍

PR #5561, opened 2026-06-18
fix: duplicate branch_name column

mchalapuk:
## Summary

- Drop the redundant `branch_name` column and make `git_branch` the single canonical branch field (`main` for live, `drafts/*` for drafts).
- Add migration `20260618002817_drop-canvas-version-branch-name`: backfill `git_branch` from `branch_name`, move the per-draft unique index and check constraint onto `git_branch`, then drop the column.
- Update model queries/helpers, API serialization, and agent tools to use `git_branch`.


db/migrations/20260618194426_drop-canvas-version-branch-name.up.sql
```diff
@@ -0,0 +1,25 @@
+BEGIN;
+
+-- Consolidate the redundant branch_name column into git_branch. For draft
+-- versions, branch_name and git_branch carry the same value (the drafts/* git
+-- branch); the git-first migration defaulted every row's git_branch to 'main',
+-- so backfill draft rows from branch_name before dropping it.
+UPDATE workflow_versions
```
lucaspin, line 7: Why not just drop the new duplicated column (git_branch) and use the old one?

mchalapuk: The new one is better because it's nonnull.

lucaspin (dismissed):

lucaspin (approved):

PR #5577, opened 2026-06-18
fix: preserve signup UTM attribution

forestileao:
## Summary
- persist first-touch UTM attribution from signup URLs in localStorage and a shared SuperPlane cookie
- set initial UTM person properties through PostHog when the app initializes
- include stored UTM fields on auth:org_create

## Tests
- npm run test -- src/lib/utmAttribution.spec.ts src/lib/analytics.spec.ts src/posthog.spec.tsx
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5578, opened 2026-06-18
fix: route magic-code signups through welcome survey

forestileao:
## Summary
- return explicit JSON redirect targets for magic-code auth requests that ask for JSON
- use that redirect target in the signup UI so direct email/code signups land on /welcome
- add spacing between the primary email signup button and OAuth provider buttons

## Tests
- make test PKG_TEST_PACKAGES=./pkg/authentication
- npm run test -- src/pages/auth/authRedirect.spec.ts src/pages/auth/magicLinkVerifyRequest.spec.ts
- make format.go
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5582, opened 2026-06-18
fix: keep signup waitlist link visible

forestileao:
## Summary
- keep the login page signup link visible when public signups are admin-disabled
- preserve the environment-disabled behavior so installs blocked by environment do not expose the public signup entry point

## Tests
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5587, opened 2026-06-19
fix: configure signup waitlist embed from env

forestileao:
## Summary
- move the signup waitlist MailerLite account/form IDs into runtime env-backed template values
- hide the embedded notify form when either MailerLite value is missing
- document the new env vars and wire them through local dev compose

## Tests
- gofmt -w pkg/web/index_template.go pkg/web/index_template_test.go
- go test ./pkg/web
- go build -o /tmp/superplane-server-check cmd/server/main.go
- cd web_src && npm run test:run -- src/pages/auth/SignupWaitlist.spec.tsx
- cd web_src && npm run lint:budget
- cd web_src && npm run build


forestileao:
👍

PR #5591, opened 2026-06-19
fix: submit waitlist emails with native form

forestileao:
## Summary
- keep the SuperPlane /signup waitlist email UI instead of rendering the HubSpot embedded form
- submit the native email field to HubSpot Forms API using the configured portal and form IDs
- include page context and the HubSpot visitor cookie when present
- cover configured, unconfigured, success, and failure states in the waitlist tests

## Setup note
- the configured HubSpot form must include an `email` field because HubSpot validates submitted fields against the form definition

## Tests
- cd web_src && npm run test:run -- src/pages/auth/SignupWaitlist.spec.tsx
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5593, opened 2026-06-19
chore(ui): remove ExecutionChainPage component

lucaspin:
The ExecutionChainPage component was mostly replaced by the RunDetailPage, and at this point, it was only used when clicking on an error in the errors console, leading to a weird view with the error console and the compact component sidebar stacked, and the run not selected.

Here, we change the behavior in the errors console to also redirect to the run detail page (`?run=`, failed node selected) instead of using the legacy execution chain component. That allows us to remove ExecutionChainPage, useExecutionChainData, and related sidebar/canvas wiring that was no longer part of the live UX.


lucaspin:
👍

PR #5600, opened 2026-06-20
fix: clear stale 'repository not found' errors after seed file provisioning

bender-rodriguez-unit1:
When an app is installed from a template with seed files (scripts, AGENTS.md, etc.), the canvas nodes are validated at publish time before the repository provisioner finishes committing the files. This causes SSH nodes with `commandSource: file` to show a stale "repository not found" error that persists even after the repo is ready.

**Fix:** After the repository provisioner commits seed files and marks the repo as ready, scan the live canvas version's nodes and clear any `errorMessage` containing "repository not found".

**Before:** User installs an app, sees warnings on SSH nodes, has to do a no-op publish to clear them.
**After:** Warnings clear automatically once the repository is provisioned.

### DIFF
--- a/pkg/workers/repository_provisioner.go
+++ b/pkg/workers/repository_provisioner.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


AleksandarCole:
So dumb.

PR #5610, opened 2026-06-20
feat: Cache AWS SNS certs

usernamenenad:
Resolving a TODO for caching AWS SNS message certs

AWS does not document a rotation cadence for these certificates, so I've implemented a conservative refresh window rather than a value tied to any known rotation schedule.


shiroyasha:
Sorry for the late review on this one 🙏 
@lucaspin does an in-memory cache work for us in this case?

shiroyasha:

After taking another look, this isn’t something we plan to move forward with right now, so I’ll close the PR for now.

PR #5620, opened 2026-06-22
fix: use draft ref as save source in edit mode; reset signature after commit

### LINKED ISSUE #5611: Publish overwrites committed changes with stale local state due to post-commit position auto-save
## Bug

After a user does **edit → change → commit → publish**, the published version does not reflect the committed changes. The bug also occurs with a 20-second wait between commit and publish.

The workaround is **exit edit → re-enter edit → publish**, which creates a fresh session.

## Root Cause

Identified via HAR analysis. The exact sequence of API calls during a buggy publish:

```
20:22:23  POST staging/files   ← user's change staged (35,817b)
20:22:27  POST staging/commit  ← user clicks Commit ✅ (35,827b committed)
          ... user waits ...
20:22:31  POST staging/files   ← 🐛 ROGUE WRITE (36,091b)  ← THIS IS THE BUG
20:22:31  POST staging/commit  ← publish-time auto-commit commits the WRONG content
20:22:31  PATCH .../publish
```

After the explicit commit, the canvas re-renders and **ReactFlow triggers a debounced position auto-save**. This generates a fresh `canvas.yaml` from the local in-memory state (which was last synced *before* the commit completed) and writes it to staging. The publish-time auto-commit in `handlePublishVersion` then commits this stale staging on top of the correctly-committed version.

The diff between the two staging writes confirms it:
- Key ordering differs: the rogue write uses `js-yaml.dump` ordering (`id:` before `component:`) vs the committed server YAML (`component:` before `id:`)  
- One node has a position change: `y: 336` → `y: 24`

The bug only affects larger canvases because more nodes = more ReactFlow position update events, making it more likely the debounced auto-save fires within the publish window.

## Fix

In `useDraftRecovery.ts`, `handlePublishVersion` already receives `cancelPendingCanvasSaves` but does not call it before the publish-time auto-commit. Adding one call cancels the in-flight position save before staging is committed:

```diff
// useDraftRecovery.ts — handlePublishVersion
+ cancelPendingCanvasSaves?.();
  consoleMutationGenerationRef.current += 1;
  await commitCanvasStagingMutation.mutateAsync();
  await publishCanvasVersionMutation.mutateAsync(versionIdToPublish);
```

Note: `cancelPendingCanvasSaves` is already called in `handleResetStaging` — this fix mirrors that pattern.

superleomori:
## Root cause

After an explicit commit, `canvasKeys.detail` is invalidated and refetches from `GET /canvases/{id}` — which returns the **live version's spec** (no draft changes). `getCurrentWorkflowSnapshot()` reads from this cache, so any auto-save triggered after commit stages the live version and overwrites the correctly-committed draft when publish auto-commits.

Confirmed via HAR analysis: a rogue `POST staging/files` fires between explicit commit and publish, staging the live canvas. The publish-time `staging/commit` then applies it, discarding the user's changes. See #5611.

## Fix

**1. `getCurrentWorkflowSnapshot()` — prefer draft ref in edit mode**

When `activeCanvasVersionId` is set (user is editing a draft), return `canvasRef.current` instead of the live canvas query cache. `canvasRef.current` only updates from user edits and staging write responses, never from live canvas refetches.

**2. `handleCommitStaging` — reset signature baseline after commit**

Call `setLastSavedWorkflowSnapshot(null)` after commit so `hasPendingLocalDraftChanges()` returns false until the next user edit. This prevents `ensureVersionActionDraftReady` from detecting a spurious diff and generating a rogue staging write.

Also retains `cancelPendingCanvasSaves()` calls in both `handleCommitStaging` and `handlePublishVersion` as defensive guards against concurrent debounced position saves.

Closes #5611

### DIFF
--- a/web_src/src/pages/app/index.tsx
+++ b/web_src/src/pages/app/index.tsx
@@ -2742,8 +2742,18 @@ export function AppPage() {
       return canvasRef.current;
     }
 
+    // When editing a draft, prefer the in-memory draft ref over the live canvas
+    // cache. After an explicit commit, `canvasKeys.detail` is invalidated and
+    // refetches the live version spec — which does NOT contain the draft's
+    // committed changes. Using it as the save source would stage the live version
+    // and overwrite the correctly-committed draft row when publish auto-commits.
+    // See: github.com/superplanehq/superplane/issues/5611
+    if (activeCanvasVersionIdRef.current) {
+      return canvasRef.current || queryClient.getQueryData<CanvasesCanvas>(canvasKeys.detail(organizationId, canvasId));
+    }
+
     return queryClient.getQueryData<CanvasesCanvas>(canvasKeys.detail(organizationId, canvasId)) || canvasRef.current;
-  }, [organizationId, canvasId, queryClient]);
+  }, [organizationId, canvasId, queryClient, activeCanvasVersionIdRef]);
 
   const hasPendingLocalDraftChanges = useCallback(() => {
     if (!activeCanvasVersionIdRef.current) {
@@ -3889,6 +3899,7 @@ export function AppPage() {
     setIsPreparingVersionAction,
     flushRepositoryFileStaging,
     cancelPendingCanvasSaves,
+    setLastSavedWorkflowSnapshot,
     onCanvasDraftRestoredToCommitted: handleCanvasDraftRestoredToCommitted,
     recoverIfDraftMissing,
   });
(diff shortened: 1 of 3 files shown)


shiroyasha:
## PR Risk Review

**Risk:** 25/100 (low)
**Review approved:** Yes
**Check passed:** Yes

### Summary
Small, well-scoped fix that cancels pending debounced saves before commit/publish to prevent stale state overwrites, with clear comments referencing the originating issue.

### Concerns
- No tests added to verify the race condition fix, though the change is defensive (optional chaining on cancelPendingCanvasSaves) and low blast radius.
- Relies on cancelPendingCanvasSaves being correctly passed into the hooks; if it's undefined at runtime the fix silently no-ops without logging.

shiroyasha:
Closing. This is slop.

Issue #5603, opened 2026-06-20
Double fire schedule


shiroyasha:
<img width="2962" height="510" alt="Image" src="https://github.com/user-attachments/assets/cf9523ef-1caa-4f8d-9f25-7e6628f766e1" />

PR #5622, opened 2026-06-22
fix: only lock CanvasNodeRequest for processing when still pending

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/5603. When locking node requests, only lock if request is still pending. Also, add more logging in node request worker.


shiroyasha:
## PR Risk Review

**Risk:** 25/100 (low)
**Review approved:** Yes
**Check passed:** Yes

### Summary
Adds state and run_at filtering to the LockNodeRequest query to prevent re-processing of already-completed or not-yet-due requests, with appropriate tests and improved structured logging.

### Concerns
- The `time.Now()` is captured before the query executes, which could allow a very slight race where a request becomes due between the time check and actual row lock, though this is unlikely to cause issues in practice.
- The error handling change in LockAndProcessRequest now returns nil (silently skips) on ErrRecordNotFound, which is correct for the 'already processed' case but could also mask cases where the request ID is genuinely invalid.
- ListNodeRequests() still returns all pending requests without the state/run_at filter, so the worker loop may still attempt to lock requests that will be skipped - this is safe but slightly inefficient.

lucaspin:
👍

PR #5628, opened 2026-06-22
fix: preserve expression value types in json bodies

### LINKED ISSUE #4499: HTTP action: expressions in JSON body fields resolve to strings, breaking numeric APIs

re-pixel:
## Summary

Resolves #4499.

HTTP JSON bodies (and other schemaless object fields) were stringifying whole `{{ }}` expression results, so values like `0.1` / `true` became `"0.1"` / `"true"` on the wire and broke APIs that expect real JSON numbers/booleans (e.g. Cloudflare load-balancer pool weights).

## Why not preserve types everywhere? (#4652)

An earlier approach changed `ResolveTemplateExpressions` globally: if a string was a single `{{ … }}`, return the native expression result.

That fixed JSON bodies, but broke string-typed config fields that often receive numeric expression results. Example: GitHub `updateIssueComment` declares `commentId` as `FieldTypeString` / `CommentID string`. Config like `{{ previous().id }}` can evaluate to a number; keeping that number made mapstructure decode fail. Those IDs must stay strings (then `parseCommentID` converts for the API).

So a global “always keep native types” rule is unsafe.

## Approach in this PR

Decide from the **field’s declared schema type** (not from nesting depth, and not from the expression’s return type alone):

| Field type | Whole `{{ … }}` value |
| --- | --- |
| `string` (and similar) | Always stringify |
| `number` / `boolean` | Keep native type |
| `object` **without** schema (free-form JSON) | Walk the tree; keep native types on whole-expression leaves |
| `object` **with** schema | Recurse per nested field type |
| Mixed templates (`"pool-{{ id }}"`) | Always stringify |

This keeps HTTP/JSON payloads typed while leaving string fields (like GitHub `commentId`) stringified.


shiroyasha:
I need @lucaspin and @forestileao to review this as well. Not a trivial change.

shiroyasha:
Here is one similar PR https://github.com/superplanehq/superplane/pull/4652. Lets review that approach as well.

forestileao:
👍

Issue #5634, opened 2026-06-22
Restrict sendEmail component to SMTP availability


forestileao:
## Context
Production Resend quota was being consumed by the core sendEmail component. The component should only be exposed and executable when SMTP email settings are configured for self-hosted installs.

## Scope
- Hide unavailable sendEmail from runtime action listing and UI building blocks.
- Keep static docs and lint generation aware of registered components.
- Guard direct setup and execution paths.

PR #5635, opened 2026-06-22
fix: remove core sendEmail notifications

forestileao:
## Summary
- remove the core sendEmail workflow component and generated docs entry
- remove workflow notification email context, worker, queue message, templates, and obsolete proto payload
- replace the built-in Health Check Monitor template email alert with an in-canvas display marker
- keep email delivery only for login/signup/invitation flows and existing SMTP/SendGrid integrations

Refs #5634

## Tests
- make gen.components.docs
- make pb.gen
- make format.go
- make format.js
- go test ./pkg/cli/commands/apps/canvas
- go test -run '^$' ./pkg/services ./pkg/server ./pkg/workers ./pkg/workers/contexts ./pkg/registry ./pkg/grpc/actions/canvases ./pkg/public ./test/support/...
- go build ./cmd/server
- npm run build (web_src)
- make lint
- make check.build.app
- make check.build.ui

Note: full targeted Go tests that require DB were not run locally; compile-only checks passed because no local Postgres test DB was configured.


pkg/components/send_email/send_email.go
```diff
@@ -26,6 +29,13 @@ func init() {
 
 type SendEmail struct{}
 
+var errSMTPOnly = errors.New("sendEmail is only available for self-hosted installations using SMTP email settings")
+
+var smtpEmailSettingsConfigured = func() bool {
```
lucaspin: This shouldn't be done here

pkg/core/action.go
```diff
@@ -108,6 +108,19 @@ type Action interface {
 	Cleanup(ctx SetupContext) error
 }
 
+type AvailableAction interface {
```
lucaspin: This is weird too

pkg/components/send_email/send_email.go
forestileao: yep, no db calls should be made here. working in the pr yet

forestileao:
Updated this PR to follow the review direction: removed the core `sendEmail` component and workflow notification email provider path entirely instead of adding SMTP availability checks.

The remaining email paths are auth/invitation/magic-code delivery and the existing SMTP/SendGrid integration actions.

forestileao:
/docs-reject component removed from core docs and product surface

forestileao:
👍

lucaspin (approved):

PR #5643, opened 2026-06-22
chore: remove dead canvas save activity log plumbing

lucaspin:
Removes unused canvas save “activity log” code that built and stored save history entries but never surfaced them in the UI. Save operations were creating `liveCanvasEntries` via `summarizeWorkflowChanges` and `buildCanvasStatusLogEntry`, but `CanvasLogSidebar` only renders run errors and node **warnings** — save entries were `type: "success"` and were never shown.


lucaspin:
👍

PR #5650, opened 2026-06-23
chore: remove dead canvas unsaved-state and manual save UI

lucaspin:
- Remove `hasUnsavedChanges` and `hasNonPositionalUnsavedChanges`, which were never set to `true` after `markUnsavedChange` was removed — all dirty tracking now lives in autosave (`handleSaveWorkflow` / `enqueueCanvasSave`) and save signatures
- Drop the dead `hasRunBlockingChanges` run guard and its “Save canvas changes before running” tooltip (always false)
- Remove the manual header Save button path (`onSave`, `saveButtonHidden`, `saveIsPrimary`, `saveDisabled`, etc.) from `CanvasPage` and header components
- Use `hasLocalSaveActivity` directly instead of the redundant `hasPendingLocalCanvasState` alias

## Context

The app autosaves canvas edits; the old boolean dirty flags and header Save button were leftover plumbing with no effect on runtime behavior. Save signature logic (`getWorkflowSaveSignature` / `setLastSavedWorkflowSnapshot`) is unchanged — still used for save conflict detection and commit settle.


lucaspin:
Regarding bugbot findings: bugbot is confusing "what this PR changed" with "what might be desirable". So, both findings are false positives for this PR

lucaspin:
👍

PR #5654, opened 2026-06-23
fix: simplify usage settings limits display

forestileao:
## Summary
- remove usage-page cards for canvases, members, integrations, and nodes per canvas
- render unlimited event usage as `count / ∞`
- keep the remaining limits grid to two cards per row

## Validation
- `make format.js`
- `npm run build`

Note: `make check.build.ui` could not run locally because Docker was unavailable.


forestileao:
👍

PR #5655, opened 2026-06-23
fix: publish agent token usage events

forestileao:
## Summary
- add an agent token usage RabbitMQ message on `superplane.agents-exchange` with routing key `agent.run.finished`
- map Anthropic completed-turn token usage into provider events
- publish token usage from the agent stream worker when a completed turn reports positive total tokens

## Validation
- `make pb.gen`
- `make format.go`
- `make test PKG_TEST_PACKAGES="./pkg/agents/anthropic ./pkg/workers"`
- `make lint && make check.build.app`


forestileao:
👍

PR #5657, opened 2026-06-23
chore: remove unused canvas changeset apply/validate endpoints

lucaspin:
These are endpoints used by the old agent. New agent no longer needs them, and UI also does not need them. Cleaning it up.


lucaspin:
👍

PR #5661, opened 2026-06-23
fix: preserve usage setup precondition errors

forestileao:
## Summary
- Preserve SaaS usage setup precondition failures as FailedPrecondition instead of HTTP 500
- Return FailedPrecondition when usage remains unconfigured after a forced setup attempt
- Add regression coverage for both forced-sync failure paths

## Tests
- make format.go
- make test PKG_TEST_PACKAGES="./pkg/grpc/actions/organizations"
- git diff --check


forestileao:
👍

PR #5664, opened 2026-06-23
chore: move CanvasChangeset from proto to internal changesets package

lucaspin:
Now that we no longer have changeset endpoints - see https://github.com/superplanehq/superplane/pull/5657 - there's no need to expose the changeset type in the API definitions, since it's only used internally now. Here, we move it to pkg/grpc/actions/canvases/changesets package.


lucaspin:
👍

PR #5667, opened 2026-06-23
chore: remove legacy canvas view modes and stale header modes

lucaspin:
Removes dead canvas view/header mode plumbing left over from earlier UX refactors. Behavior is unchanged for current flows — editing still uses `isEditing`, run inspection still uses `?run=` + `isRunInspectionMode`, and the
  in-edit-session Versions sidebar is kept.
  - **Legacy `?view=versions` tab:** drop `isVersionsMode`, `useVersionsModeActions`, and header mode `"versions"`; migrate old URLs the same way we already handle `?view=runs` (strip the param on load / version selection). Version
  history stays in the edit-session sidebar.
  - **Stale header modes:** remove `"version-edit"` and `"runs"` from `WorkflowHeaderMode` / `HeaderMode` unions and related branches. Nothing set these anymore — edit mode is `isEditing`, run inspection keeps
  `headerMode="version-live"`.
  - **Header cleanup:** stop inferring edit state from `mode === "version-edit"`; use `isCanvasTabHeaderMode()` for canvas-tab actions; simplify `PageHeader` (no versions-tab Edit button guard).
  ~120 lines removed across `viewState`, AppPage hooks, and Canvas header components.


lucaspin:
👍

PR #5669, opened 2026-06-24
fix: sync usage before publishing agent tokens

forestileao:
## Summary
- Give the agent stream worker an optional usage service dependency
- Sync/setup the organization with SaaS before publishing completed-turn token usage
- Add regression coverage that token usage publishing attempts org usage setup first

## Tests
- make format.go
- make test PKG_TEST_PACKAGES="./pkg/workers"
- make lint
- docker compose -f docker-compose.dev.yml exec app go build ./pkg/workers
- git diff --check

## Note
- make check.build.app currently fails on main with an unrelated generated proto/server mismatch: CanvasService is missing ApplyCanvasVersionChangeset.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -259,7 +259,12 @@ func startWorkers(
 			AuthService:       authService,
 			UsageService:      getOptionalWorkerUsageService(),
 		})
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #5671, opened 2026-06-24
fix: track managed agent session usage

forestileao:
## Summary
- fetch Managed Agents session usage after idle turns because Anthropic does not include usage on the idle SSE event
- parse the current nested cache_creation usage shape
- store per-session cumulative usage watermarks so resumed turns publish only the new token delta

## Verification
- Live Anthropic probe confirmed session.status_idle stream events have no usage while GET /sessions/{id} returns usage
- make test PKG_TEST_PACKAGES="./pkg/agents/anthropic ./pkg/workers ./pkg/models"
- make lint
- docker compose -f docker-compose.dev.yml exec app go build ./pkg/agents/anthropic ./pkg/workers ./pkg/models

## Known unrelated failure
- make check.build.app currently fails on main with CanvasService missing ApplyCanvasVersionChangeset from the generated canvases service interface

### DIFF
--- a/db/migrations/20260624035629_add-agent-session-token-usage-watermark.up.sql
+++ b/db/migrations/20260624035629_add-agent-session-token-usage-watermark.up.sql
@@ -1,0 +1,14 @@
+BEGIN;
+
+ALTER TABLE agent_sessions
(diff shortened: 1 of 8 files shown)


forestileao:
👍

PR #5672, opened 2026-06-24
fix: avoid blocking agent stream completion on usage

forestileao:
## Summary
- stop fetching Anthropic session usage inside the SSE stream before forwarding turn completion
- retrieve missing completed-session usage through an optional provider capability in the worker billing path
- publish websocket turn completion before token usage sync/Rabbit publishing can block

## Tests
- make test PKG_TEST_PACKAGES="./pkg/agents/anthropic ./pkg/workers ./pkg/models"
- make lint
- docker compose -f docker-compose.dev.yml exec app go build ./pkg/agents ./pkg/agents/anthropic ./pkg/workers ./pkg/models
- git diff --check


forestileao:
👍

PR #5673, opened 2026-06-24
fix: always refetch usage settings page

forestileao:
## Summary
- let useOrganizationUsage accept per-call query cache options
- make the Usage settings page use an isolated no-cache query scope
- force a fresh usage fetch whenever the Usage page mounts

## Tests
- make format.js
- cd web_src && npx eslint src/hooks/useOrganizationData.ts src/pages/organization/settings/Usage.tsx --quiet
- git diff --check

## Notes
- make check.build.ui is currently blocked on main by missing generated API export ComponentsEdge in @/api-client, referenced by existing app/canvas files.


forestileao:
👍

PR #5677, opened 2026-06-24
chore: Introduce grpcerror package

shiroyasha:
### Root problem

We were treating `context.Canceled` as a HTTP 500 error. 
This generated fake negative signals in observability.

### What am I changing?

The current error handling pattern in pkg/grpc masks the error, and does not let me process is it the HTTP middleware.

```golang
if err != nil {
  return nil, status.Error(codes.Internal, "something failed")

  // original err never returned
}
```

### New solution

We now have a `grpcerrors` package that handles this.

The pattern is:

```golang
grperrors.Internal(<original err>, <message sent to the user>)
```

```golang
if err != nil {
  return nil, grperrors.Internal(err, "something failed")
}
```

Other methods exist as well:

- `grpcerrors.NotFound`
- `grpcerrors.InvalidArgument`
- ...

### Adding a linter

This change would be meaningless if we didn't have a linter that prevented us from using the old `status.Error` pattern.

Added a small grep based linter to our CI.


pkg/grpc/actions/auth/add_user_to_group.go
```diff
@@ -2,26 +2,24 @@ package auth
...
 	user, err := FindUser(orgID, userID, userEmail)
 	if err != nil {
-		return nil, status.Error(codes.InvalidArgument, "user not found")
+		return nil, grpcerrors.InvalidArgument(nil, "user not found")
 	}
 
 	err = authService.AddUserToGroup(domainID, domainType, user.ID.String(), groupName)
 	if err != nil {
-		return nil, status.Error(codes.Unknown, err.Error())
+		return nil, grpcerrors.Unknown(err, "failed to add user to group")
```
shiroyasha: Why would we need this instead of grpcerrors.Internal?

pkg/grpc/actions/agents/common.go
```diff
@@ -7,10 +7,9 @@ import (
 
 	"github.com/google/uuid"
 	agentservice "github.com/superplanehq/superplane/pkg/agents"
+	"github.com/superplanehq/superplane/pkg/grpcerrors"
```
lucaspin: Any reason not to put the new package in pkg/grpc/errors? Seems like the most natural place since we already have the pkg/grpc/* pattern

shiroyasha: Good point. Moving there.

Issue #5627, opened 2026-06-22
Perf: /versions endpoint called 13 times when entering edit mode


superleomori:
## Problem

When entering edit mode on a canvas, the `/versions` endpoint is called **13 times** in a single interaction. Calls observed:

- `versions` (plain)
- `file?path=console.yaml` (×2, one with `version_id`)
- `versions?limit=1` (×4)
- `versions?limit=50` (×4)
- `versions?state=STATE_DRAFT` (×3)
- `8e7b2070-41f1-4bc7-9856-efac45ba0e0d` (version by ID ×1)

## Impact

- 13 network requests on a single user action is excessive
- Directly contributes to the high request volume on `/versions` (p99 ~893ms, see #5396)
- Multiplied across all users entering edit mode

## Fix

Audit the components/hooks that trigger on edit mode entry and deduplicate — fetches should be coordinated via a single request or a shared query cache (e.g. React Query deduplication).

PR #5680, opened 2026-06-24
fix: ignore version creation invalidations on same tab

lucaspin:
When you click Edit button, two things can happen: (A) open an existing draft, (B) create a new draft. The issue described in [#5627](https://github.com/superplanehq/superplane/issues/5627) happens on path B.

On path B, after creating the draft, we invalidate a lot of calls, from two places: the request response handler and the websocket handler. That leads to multiple duplicated requests. Here, we fix this by doing two things:
1. Use the same mechanism for same-tab websocket event suppression as used in auto save
2. Other tab only handles the canvas version updated WebSocket events if in edit mode - no need to invalidate anything in other tabs if they are editing or looking at versions at all


lucaspin:
👍

Issue #5681, opened 2026-06-24
fix: usage buckets should count actual last-30-day usage


forestileao:
## Problem

The event bucket and agent token bucket currently use proportional leak logic. Each day, the stored bucket level decreases by `limit / 30`.

That makes both the budget calculation and UI wrong. If an organization uses 500,000 events in one day, the next day it incorrectly appears to recover `500000 / 30` capacity. Instead, the buckets should reflect actual usage in the last 30 days.

## Expected behavior

For both buckets:

`bucket level = sum of actual usage from the last 30 days`

Example:
- Day 1: org uses 500,000 events.
- Days 2-30: event bucket remains 500,000 if no additional usage happens.
- Day 31: day 1 leaves the 30-day window, so the event bucket decreases by whatever was used on day 1.
- If no usage happened on days 2-31, the event bucket becomes 0.

The same logic applies to the agent token bucket.

Unlimited plans must still track bucket usage and show `used / ∞`.

The usage UI should display these actual last-30-day bucket values.

## Acceptance criteria

- Event bucket does not decrease on day 2 after a day-1 spike.
- Event bucket decreases on day 31 by the actual day-1 event count.
- Agent token bucket follows the same last-30-day bucket logic.
- Neither bucket decreases by `limit / 30`.
- Multiple daily buckets with different values sum correctly.
- Unlimited plans still track and display bucket usage.
- Duplicate event IDs and duplicate agent-token usage IDs do not double count.
- SuperPlane usage UI displays corrected bucket values as `used / limit` or `used / ∞`.

## Notes

Daily buckets should use UTC day boundaries. This issue is scoped to bucket accounting and UI correctness, not adding new runtime blocking behavior.

PR #5682, opened 2026-06-24
fix: publish stable agent token usage ids

forestileao:
## Summary

- add `usage_id` and `session_id` to agent run-finished usage messages
- generate stable per-session token usage IDs from the cumulative token watermark
- include the usage ID in publishing logs and tests so SaaS can dedupe token usage retries

## Tests

- `make test PKG_TEST_PACKAGES=./pkg/workers`
- `make lint`
- `make check.build.app`

Related: superplanehq/superplane#5681
Depends on: superplanehq/saas#23


forestileao:
👍

PR #5686, opened 2026-06-24
fix: emit canvas_version_deleted and stop stale draft fetches after delete

lucaspin:
Deleting a draft branch could leave other tabs (including passive live-view tabs) fetching the removed version, causing requests to fetch information about the deleted version. This PR introduces a dedicated deletion lifecycle event and tightens frontend cache cleanup so deleted drafts stop being queried immediately.
- Backend publishes `canvas_version_deleted` instead of reusing `canvas_version_updated` on delete
- Frontend handles the new websocket event separately: no draft resync, targeted invalidation of version/draft-branch lists
- Deleting tab cancels in-flight version queries, exits to live synchronously, then finalizes cache cleanup
- All tabs prune the deleted draft from React Query cache on delete; passive live-view tabs skip broader invalidation/resync
- Agent sidebar listens for `canvas:version-deleted` directly instead of describe-then-404
- Resync skips versions no longer listed in `draftBranches` as a safety net


lucaspin:
👍

PR #5690, opened 2026-06-24
fix: suppress duplicate draft creation requests from versions sidebar

lucaspin:
When creating a draft via the versions sidebar **"Create new draft"** button, duplicate API requests could still occur even after #5680. That PR fixed the same issue for **Edit → create first draft**, but the sidebar flow runs with `editSessionActive: true`, so websocket `canvas_version_updated` echoes were still triggering invalidations and resyncs on top of the create-draft mutation response. This applies the same suppression behavior to that flow:
  - Skip websocket-driven version invalidations while `isCreatingDraftBranch` is true (covers the sidebar create path during an active edit session).
  - Consume in-flight create-draft echo slots before the version ID is armed, so same-tab websocket events do not duplicate invalidations if they arrive before `onSuccess`.

Both paths already share `handleCreateVersion` → `useCreateDraftBranch` with echo guards; this closes the lifecycle gap for the sidebar button.

### DIFF
--- a/web_src/src/pages/app/index.tsx
+++ b/web_src/src/pages/app/index.tsx
@@ -1890,6 +1890,7 @@ export function AppPage() {
       activeCanvasVersionId,
       isEditing,
       editSessionActive,
+      isCreatingDraftBranch,
       hasLocalSaveActivity,
       isViewingLiveVersion,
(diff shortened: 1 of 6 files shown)


lucaspin:
👍

PR #5692, opened 2026-06-24
fix: dedupe commit and publish draft staging API requests

lucaspin:
Commit and publish were each firing many redundant API requests because the mutation response handler, post-action cache refresh, and same-tab WebSocket lifecycle echoes all invalidated overlapping query keys. This aligns commit/publish with the echo-suppression pattern already used for auto-save and draft creation - see https://github.com/superplanehq/superplane/pull/5680 and https://github.com/superplanehq/superplane/pull/5690 - and narrows post-action cache updates to a single intentional refresh per resource.

## Changes
- **Commit:** Register a `canvas_version_updated` echo guard before staging commit; sync version/canvas/console caches once via `syncCommittedCanvasDraftState` + `syncCommittedConsoleCaches` instead of broad invalidations; trim `useCommitCanvasStaging` `onSuccess` to staging summary + version history only.
- **Publish:** Register `canvas_updated` and `canvas_version_updated` echo guards before publish; remove `usePublishCanvasVersion` `onSuccess` invalidations (refresh is owned by exit-to-live); optimistically drop the published draft from the branches cache; pass the published version id into `refreshLatestLiveCanvasData`.
- **Live refresh:** Use `exact: true` for `versionList` invalidation (avoids double-refetching `versionHistory`); replace `consoleAll` invalidation with a targeted live-console refresh.

## Refactor

Linter started to complain after the changes, so a bit of refactoring was needed:
  - Extracted flow helpers: commit-staging-flow, publish-draft-flow, exit-draft-to-live, reset-staging-flow, draft-missing-recovery
  - Extracted useRefreshLatestLiveCanvasData from index.tsx to keep ESLint budgets green

## Analysis after fix

Using HAR files generated in the browser, we can see the difference in the number of requests.
- Commit: ~30 requests → **5** (0 duplicate URLs)
- Publish: ~46 requests → **9** (1 acceptable cross-cache console fetch during exit-to-live)

### DIFF
--- a/web_src/src/hooks/useCanvasData.ts
+++ b/web_src/src/hooks/useCanvasData.ts
@@ -1004,9 +1004,7 @@ export const useDeleteDraftBranch = (canvasId: string) => {
   });
 };
 
-export const usePublishCanvasVersion = (organizationId: string, canvasId: string) => {
-  const queryClient = useQueryClient();
-
+export const usePublishCanvasVersion = (canvasId: string) => {
   return useMutation({
     mutationFn: async (versionId: string) => {
       return await canvasesPublishCanvasVersion(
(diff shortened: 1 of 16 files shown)


lucaspin:
👍

PR #5693, opened 2026-06-24
fix: keep commit button disabled until staging action fully completes

lucaspin:
- Fix a brief flicker where the Commit button re-enabled after the API call finished but before Publish appeared.
- Track `commitStagingPending` for the full commit handler lifecycle (cache refresh, baseline reset), matching how Reset already works.


lucaspin:
👍

PR #5696, opened 2026-06-25
fix: stop stale execution events downgrading finished nodes

forestileao:
## Problem

Execution nodes on the canvas get stuck showing **"running" forever** even after the execution has finished. Reloading the page fixes it (the final state is re-fetched from the API), but it comes back on the next run.

## Root cause

Canvas nodes derive their on-canvas running/finished state from `nodeExecutionStore`, which `useCanvasWebsocket` updates on every `execution_created` / `execution_started` / `execution_finished` event. `updateNodeExecution` replaced an execution **by id with no ordering guard** — it blindly overwrote whatever was there.

This regressed after **#5507 ("use dedicated executions exchange")**. Before that PR, all execution state changes were published under a single routing key and consumed by a single consumer, so events reached the browser in order. After #5507:

- each transition is published under its own routing key (`execution.pending` / `execution.started` / `execution.finished`)
- they are consumed by **separate, concurrent goroutines** in the event distributer (`pkg/workers/event_distributer.go`)
- publishing is split across the queue worker (pending) and the executor (finished)

There is no longer any cross-state ordering guarantee, so the browser can receive a stale `created`/`started` event **after** `finished`. When that happened, the blind replace downgraded the finished execution back to `STATE_STARTED`/`STATE_PENDING`, and the node mapper renders `PENDING || STARTED` as "running" → stuck until reload.

The runs cache already guarded against this exact race (`shouldAcceptExecutionRefUpdate` in `canvasInfiniteCache.ts`), but the **node store never got the same protection**.

## Fix

- Extract a shared, exported `shouldAcceptExecutionUpdate(existing, incoming)` helper (`updatedAt` comparison + state-rank tiebreak); the existing ref guard now delegates to it.
- `updateNodeExecution` rejects stale/out-of-order updates for the same execution id, so an execution can never roll backward.
- Add `nodeExecutionStore.spec.ts` covering the out-of-order case (plus newer-update and distinct-execution cases).

Defensive fix on the receiving side, which is the correct layer — even if the backend can't guarantee cross-consumer ordering, the UI must never roll an execution backward.

### DIFF
--- a/web_src/src/hooks/canvasInfiniteCache.ts
+++ b/web_src/src/hooks/canvasInfiniteCache.ts
@@ -190,9 +190,16 @@ export function executionToRef(execution: CanvasesCanvasNodeExecution): Canvases
   };
 }
 
-function shouldAcceptExecutionRefUpdate(
-  existing: CanvasesCanvasNodeExecutionRef,
-  incoming: CanvasesCanvasNodeExecutionRef,
+// Execution state only ever moves forward (pending -> started -> finished), but
+// the backend broadcasts each transition through an independent consumer (since
+// the dedicated executions exchange landed), so the websocket can deliver them
+// out of order. A late "created"/"started" event must never overwrite a newer
+// state we already applied, otherwise a finished node gets stuck showing
+// "running" until the page is reloaded. Accept an update only when it is at
+// least as recent as what we have.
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #5702, opened 2026-06-25
fix: make previous resolve latest upstream event

forestileao:
## Summary
- make named upstream references with execution rows but no emitted output resolve as `nil` instead of failing expression evaluation
- make `previous()` recover the current incoming event payload from `incomingEventID` when no single input map is present
- make fan-in `previous(N)` walk direct upstream output events in latest-first order, without pulling stale events from transitive ancestors
- select direct upstream outputs through `executionOutputLookup`, so `previous()` uses the same consumed-event, incoming-event, and ambiguous-output rules as `$[node]`

## Context
The release app merge failed with `node Build superplane Docker image has no outputs` because the merge stop-if expression evaluated while parallel upstream nodes were still running. For fan-in components, expressions need to tolerate not-yet-emitted sibling outputs, and `previous()` should represent the current/latest direct upstream event payload rather than only a linear previous node.

## Validation
- `gofmt -w pkg/workers/contexts/node_configuration_builder.go pkg/workers/contexts/node_configuration_builder_test.go`
- `git diff --check`
- `go test -c ./pkg/workers/contexts`
- targeted DB-backed tests were attempted locally, but the host Postgres socket was unavailable; CI should run them with the test database

### DIFF
--- a/pkg/workers/contexts/node_configuration_builder.go
+++ b/pkg/workers/contexts/node_configuration_builder.go
@@ -865,7 +865,8 @@ func (b *NodeConfigurationBuilder) populateFromExecutions(
 			return fmt.Errorf("node %s: %w", nodeRef, err)
 		}
 		if !ok {
-			return fmt.Errorf("node %s has no outputs", nodeRef)
+			messageChain[nodeRef] = nil
+			continue
 		}
 
 		messageChain[nodeRef] = normalizeExpressionValue(event.Data.Data())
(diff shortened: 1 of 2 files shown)


forestileao:
> Fallback to latestUpstreamOutputPayload is gated to depth == 1 only; previous(N) for N > 1 never uses this path.

for N > 1 we should get the latest payload for that node in that run too

forestileao:
👍

PR #5706, opened 2026-06-25
chore: remove legacy YAML modal

lucaspin:
Now that you can edit the YAMLs directly in files tab, there's no need for this anymore


lucaspin:
👍

PR #5707, opened 2026-06-25
fix: resolve missing execution node references as nil

forestileao:
## Summary
- resolve known canvas node references with no execution in the current chain as nil
- keep unknown node names and ambiguous output resolution on their existing error paths
- add regression coverage for guarded expressions that check a not-yet-executed node

## Context
The release app hit `error resolving field prompt: node Review & Approve Changelog not found in execution chain` on the first Generate Changelog loop. On the first pass, that review node has not executed yet, so guarded expressions need the reference to evaluate as nil instead of failing while building the message chain.

## Validation
- `make format.go`
- `make test PKG_TEST_PACKAGES=./pkg/workers/contexts`
- `git diff --check`

No token values or secret payloads are included in this PR body or diff.


forestileao:
👍

PR #2899, opened 2026-02-05
feat: Using secrets in expressions

shiroyasha:

PR #5708, opened 2026-06-26
feat: support secrets in expressions and HTTP node authorization

andrecalil:
Adds first-class secret support across components and the HTTP node, and hardens secret handling.

- New secrets("name").key expression function usable in any component's text/expression fields; resolved at execution time and never persisted into stored configuration. Surfaced in expression validation, autocomplete, and in-app expression help.
- HTTP node authorization (Bearer / Basic Auth / Custom Header) backed by organization secrets via SecretKeyRef.
- Secret-usage redesign addressing security concerns in the node configuration builder / secret resolver.
- Docs: new general "Expressions" page documenting secrets(); component docs regenerated.


shiroyasha:
From the security side:

What happens if I enter value = `{{ secrets.myserver.sshKey + "aaa" }}`?

shiroyasha:
From the product side:

I don't like that expressions are working differently in the HTTP component compared to the rest of the system. It introduces diverging behavior, that is harder to document and maintain.

Take a look at some of the previous PRs that tried to address this in the whole app, you might find some inspiration:

https://github.com/superplanehq/superplane/pull/2899

shiroyasha (changes requested):
Lets try to figure out a way for all components to be able to access secrets in expressions.

andrecalil:
> Lets try to figure out a way for all components to be able to access secrets in expressions.

What about this?
```
Every text field in this component accepts standard expression placeholders
`{{ ... }}`. Use `secrets("name").key` inside an expression to inject an
organization secret value (for example: `{{ secrets("api").token }}`). Secrets are
resolved at execution time and are not stored back into the saved configuration.
```

pkg/components/http/http.go
```diff
@@ -162,6 +162,13 @@ and status without downloading the full response payload.
   - **Plain Text**: Raw text content
   - **XML**: XML formatted content
 
+## Secrets in expressions
```
lucaspin: This shouldn't be here, right? Since this is not specific to the HTTP node, we should document it [here](https://docs.superplane.com/concepts/expression-functions/) instead

lucaspin: Also, the PR title and description is no longer true. Please, update it as well so we don't mess our changelogs

andrecalil: Agreed on both, addressed

shiroyasha:
Tried to review it today, not an easy task.

PR #5763, opened 2026-06-29
feat: add openai and claude structured output config

WashingtonKK:
## Summary

This adds structured output as a configuration option for the following components:

- `claude.textPrompt`
- `openai.textPrompt`


pkg/configuration/structuredoutput/structuredoutput.go
```diff
@@ -0,0 +1,165 @@
...
+//   - Every object node gets "additionalProperties": false.
+//   - OpenAI strict mode (strict=true) additionally forces every object's
+//     "required" to list all of its properties; optional fields are expressed by
+//     making their type nullable in the schema itself.
+//   - Anthropic (strict=false) leaves "required" as the user wrote it.
+package structuredoutput
+
+import (
+	"encoding/json"
+	"fmt"
```
forestileao, line 16: Maybe these files related only to structured output should not be here?
Since they are not integrations, and just helpers for integrations, they could be under "pkg/configuration"

WashingtonKK: This is fixed @forestileao

forestileao:
👍

PR #5767, opened 2026-06-29
chore: Remove RunSpan in favor of defer style spans

shiroyasha:
Two reasons:

1/ Cosmetically: The RunSpan was dirtying up the call site
2/ Controllable error: The RunSpan was handling the error no matter what, even if we don't want to report it.


shiroyasha:
I hate this:

CleanShot 2026-06-29 at 19 09 21@2x

But, that is a job for another PR.

PR #5770, opened 2026-06-29
chore: Source maps for Sentry (Garbage code, do not merge)

### LINKED ISSUE #5761: Send JS sourcemaps to Sentry to avoid criptic issues
e.g. https://superplane.sentry.io/issues/7579740229/?query=is%3Aunresolved&referrer=issue-stream

shiroyasha:
fixes https://github.com/superplanehq/superplane/issues/5761

This change enables sourcemaps in all JS environments so that sentry can understand our exceptions. This would not be safe in proprietary code as it would expose the source code, but given that we are open-source, it is ok.

Follows guide from here:
https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/hosting-publicly

---

Alternative pattern is to upload during deployment, but I've found that method less reliable when I was setting this up on Operately.


shiroyasha:
Converting this back to draft. Will have to split it into parts.

PR #5786, opened 2026-06-30
fix: resolve missing expression tokens as null

forestileao:
## Summary
- Resolve missing `$["..."]` node references as null instead of failing while building the expression environment
- Keep ambiguous node-name errors intact and preserve dereference failures as expression evaluation errors
- Add regression coverage for missing expression token resolution

## Tests
- `gofmt -w pkg/workers/contexts/node_configuration_builder.go pkg/workers/contexts/node_configuration_builder_test.go`
- `git diff --check`
- `go test ./pkg/components/runner -run TestRunnerExecuteSendsEnvironmentToBroker`

Note: `make format.go` and targeted `go test ./pkg/workers/contexts ...` were blocked locally because Docker was unavailable and generated protobuf packages were missing in the checkout.


forestileao:
👍

PR #5793, opened 2026-06-30
chore: migrate SuperPlane JWT usage to jwt/v5 [wip]

shiroyasha:
## Summary

- Migrate all first-party JWT signing and validation to `github.com/golang-jwt/jwt/v5` per the [migration guide](https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md)
- Use `jwt.NewNumericDate()` for temporal claims in HS256 and RS256 tokens
- Replace removed v4 helpers (`VerifyExpiresAt`, `VerifyAudience`, `RegisteredClaims.Valid`) with v5 parser validation and `ClaimsValidator`
- Drop direct `jwt/v4` imports from SuperPlane code (`jwt/v4` remains only as an indirect dependency via `ghinstallation`)

## Test plan

- [x] `make test PKG_TEST_PACKAGES='./pkg/jwt ./pkg/oidc ./pkg/authentication ./pkg/components/runner ./pkg/public/middleware ./pkg/public'`
- [x] `make lint`

### DIFF
--- a/pkg/authentication/authentication.go
+++ b/pkg/authentication/authentication.go
@@ -14,7 +14,7 @@ import (
 	"strings"
 	"time"
 
-	jwtLib "github.com/golang-jwt/jwt/v4"
+	jwtLib "github.com/golang-jwt/jwt/v5"
 	"github.com/gorilla/mux"
 	"github.com/markbates/goth"
 	"github.com/markbates/goth/gothic"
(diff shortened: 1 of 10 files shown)


pkg/authentication/authentication.go
```diff
@@ -824,8 +824,8 @@ func (a *Handler) generateMagicLinkToken(email, code string) (string, error) {
 	token := jwtLib.NewWithClaims(jwtLib.SigningMethodHS256, jwtLib.MapClaims{
 		"email": email,
 		"code":  code,
-		"iat":   now.Unix(),
-		"exp":   now.Add(magicCodeTTL).Unix(),
+		"iat":   jwtLib.NewNumericDate(now),
+		"exp":   jwtLib.NewNumericDate(now.Add(magicCodeTTL)),
```
shiroyasha, line 828: Why we need this change?

pkg/jwt/scoped.go
```diff
@@ -30,6 +30,63 @@ type Permission struct {
 	Resources    []string `json:"resources,omitempty"`
 }
 
+func (c ScopedTokenClaims) GetExpirationTime() (*gojwt.NumericDate, error) {
```
shiroyasha, line 33: Why do we need these, and we didn't before?

shiroyasha:
The transition is not worth it atm. Not a simple version bump.

PR #5794, opened 2026-06-30
feat: OIDC execution tokens for CI workflow authorization

shiroyasha:
New: "Add OIDC assertion" on the `semaphore.runWorkflow` component.

When set to true, SuperPlane will inject a SUPERPLANE_OIDC_ASSERTION into the Semaphore pipeline, allowing the pipeline to verify who was the originator of the call.

On Semaphore's side, you can use this token to verify it:

```
superplane oidc verify \
  --audience semaphore \
  --claim org_id=eb3aca82-3864-4f76-b1d6-f670c297f136 \
  --claim app_id=c4d5e6f7-8901-4234-a567-890123456789 \
  --claim pipeline_file=.semaphore/production/deploy.yml \
  --claim node_id=deploy-production
```


pkg/integrations/semaphore/components/run_workflow.go
```diff
@@ -101,6 +118,32 @@ func (r *RunWorkflow) Documentation() string {
...
+  --claim app_id=eb3aca82-3864-4f76-b1d6-f670c297f136 \
+  --claim pipeline_file=.semaphore/production/deploy.yml \
+  --claim node_id=deploy-production
+` + "```" + `
+
+Expected claim values must be pinned in your pipeline — hardcoded literals or Semaphore project env vars — not taken from parameters SuperPlane injects at trigger time.
+
+Set **Commit SHA** on the component when you need the token to bind to a specific revision; otherwise pin ` + "`ref`" + ` and still compare ` + "`commit_sha`" + ` against ` + "`$SEMAPHORE_GIT_SHA`" + ` in CI so the job runs the code SuperPlane attested.
+
+The command reads the token from the environment, fetches JWKS from your SuperPlane instance, and verifies the signature locally. Pass expected claims with ` + "`--claim key=value`" + ` (repeatable) to enforce your policy — SuperPlane attests what was triggered; your pipeline decides what it accepts. Exit code 0 means the token is valid and matches your expectations; any other exit code aborts the job.
```
shiroyasha: TMI. Shorten it.

PR #5798, opened 2026-07-01
fix: keep run sidebar selection focused

### LINKED ISSUE #5791: Run selection from sidebars moves to random nodes
## Problem
Clicking on a Run from the bottom sidebar or from the runs sidebar causes the canvas/view to bug out and move to seemingly random nodes instead of keeping the expected run context in focus.

## Expected behavior
Selecting a run from either sidebar should open or focus that run without unexpectedly jumping to unrelated nodes.

## Notes
This appears to affect run inspection/navigation UX and should be reviewed as part of the refinement work.

forestileao:
## Summary
- include fully fetched run executions when building run inspection canvas participants
- keep live run websocket updates visible in the runs sidebar cache
- preserve the current canvas viewport when entering and leaving run inspection
- stop refitting the run canvas when changing selected runs
- allow node-specific run clicks from the bottom runs/errors sidebar, bottom node inspector history, the runs sidebar detail node list, and the run detail pane to focus that node from the current viewport
- avoid hydrating the live node inspector from run-detail URL params while in run inspection
- wait for ReactFlow initialization before applying node focus so viewport restoration cannot override the focus animation
- suppress duplicate handling of the same focus request so the node is not zoomed twice
- scope focus requests to live vs run inspection mode so the live canvas cannot consume bottom-sidebar run focus before run inspection mounts
- clear pending run focus requests when switching to a non-node run or leaving node-specific run inspection
- persist the post-focus viewport after ReactFlow fitView so leaving run inspection keeps the viewed position
- retry node focus after run canvas nodes mount so navigation made during run loading still zooms to the target node
- add regression coverage for participant merging, live run cache updates, websocket refreshes, run/live viewport transitions, delayed run-node focus, duplicate focus suppression, ignored live/run focus requests, persisted focus viewport, and bottom-sidebar node run selection context

Fixes #5791

## Tests
- make format.js
- cd web_src && npm run test:run -- src/ui/componentSidebar/CompactSidebarEventRow.spec.tsx src/ui/componentSidebar/pages/HistoryQueuePage.spec.tsx src/ui/CanvasPage/index.runInspection.spec.tsx src/pages/app/lib/run-inspection-viewport.spec.ts src/hooks/useCanvasWebsocket.spec.ts src/hooks/canvasInfiniteCache.spec.ts src/pages/app/useRunCanvasData.spec.ts src/pages/app/useRunsDetailState.spec.ts src/pages/app/viewState.spec.ts
- git diff --check
- make check.build.ui

### DIFF
--- a/web_src/src/hooks/canvasInfiniteCache.spec.ts
+++ b/web_src/src/hooks/canvasInfiniteCache.spec.ts
@@ -95,6 +95,57 @@ describe("upsertRunIntoInfiniteData", () => {
     expect(next.pages[0]?.totalCount).toBe(1);
   });
 
+  it("preserves existing run details when a live update omits them", () => {
+    const old = makeInfiniteRunsData([
+      makeRun({
+        rootEvent: { id: "event-1", nodeId: "trigger-1", createdAt: "2026-06-01T12:00:00.000Z" },
+        executions: [
+          {
+            id: "execution-1",
+            nodeId: "node-1",
+            state: "STATE_FINISHED",
+            updatedAt: "2026-06-01T12:00:30.000Z",
+          },
+        ],
+        updatedAt: "2026-06-01T12:00:30.000Z",
+      }),
+    ]);
+    const incoming: CanvasesCanvasRun = {
+      id: "run-1",
+      canvasId: "canvas-1",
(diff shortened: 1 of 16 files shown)


forestileao:
👍

PR #5804, opened 2026-07-01
fix: stop runner log fetches at record limit

forestileao:
## Summary
- stop runner live-log reads as soon as the requested record limit is reached
- avoid waiting for the task broker live stream to close before returning agent/CLI log slices
- add regression coverage for open broker streams that keep the body alive after emitting enough records

## Tests
- go test ./pkg/components/runner -run TestReadLiveLogRecordsReturnsAfterLimitOnOpenStream -count=1
- go test ./pkg/components/runner -count=1
- make format.go
- make test PKG_TEST_PACKAGES='./pkg/components/runner ./pkg/agents/agent_tools/... ./pkg/cli/commands/executions'
- make lint
- make check.build.app


forestileao:
👍

PR #5811, opened 2026-07-02
fix: reconcile agent chat streaming status

### LINKED ISSUE #5780: Internal agent gets stuck in "Thinking" state - likely a frontend issue
## Bug

The internal AI agent sometimes gets stuck displaying "Thinking" indefinitely. The backend (Claude) appears to have already responded, suggesting this is a frontend state management issue where the UI does not transition out of the thinking/loading state.

## Expected Behavior

Agent should display its response once the backend returns it.

## Actual Behavior

Agent stays in "Thinking" state permanently. Requires a page refresh or re-trigger to see the actual response.

forestileao:
## Summary
- Fixes #5780 by reconciling the local agent streaming state with the durable chat row when the terminal websocket event is missed.
- Adds a 3s refetch loop only while the local agent status is `streaming`; websocket events remain the primary status path.
- Keeps the Thinking indicator derived from status rather than hiding it when assistant text appears.

## Regression coverage
- Added sidebar coverage for a missed terminal event: the UI starts in `streaming`, a durable chat refetch returns `idle`, and Thinking/Stop clear.
- Added companion coverage that Thinking remains while durable refetches still report `streaming`.

### DIFF
--- a/web_src/src/components/CanvasToolSidebar/AgentTabPanel.tsx
+++ b/web_src/src/components/CanvasToolSidebar/AgentTabPanel.tsx
@@ -38,7 +38,7 @@ import type { AgentMessage, AgentOutgoingImage } from "./types";
 import type { CanvasToolSidebarState } from "./useCanvasToolSidebarState";
 import { groupMessages } from "./agentMessageGroups";
 
-const STREAMING_STATUS_RECONCILE_INTERVAL_MS = 3000;
+const STREAMING_STATUS_RECONCILE_INTERVAL_MS = 15000;
 
 type ChatConversationProps = {
   chatId: string;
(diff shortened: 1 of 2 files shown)


web_src/src/components/CanvasToolSidebar/AgentTabPanel.tsx
```diff
@@ -232,6 +243,48 @@ function ChatConversation({
...
+      try {
+        const nextStatus = await refreshChatStatus();
+        if (active && nextStatus && nextStatus !== "streaming") {
+          setStatus(nextStatus);
+        }
+      } catch {
+        // Websocket events remain the primary status path; refetch only repairs missed terminal events.
+      } finally {
+        inFlight = false;
+      }
```
forestileao: put this function outside useEffect

web_src/src/components/CanvasToolSidebar/AgentTabPanel.tsx
```diff
@@ -38,11 +38,14 @@ import type { AgentMessage, AgentOutgoingImage } from "./types";
 import type { CanvasToolSidebarState } from "./useCanvasToolSidebarState";
 import { groupMessages } from "./agentMessageGroups";
 
+const STREAMING_STATUS_RECONCILE_INTERVAL_MS = 3000;
```
forestileao: Make this greater value like 15000

forestileao:
👍

PR #5812, opened 2026-07-02
fix: stop culling canvas nodes near the viewport edge

### LINKED ISSUE #5125: Nodes are disappearing on the canvas
When nodes are close the edge of the screen, they disappear:

Image

forestileao:
## Summary

- Fixes #5125 by replacing React Flow's `onlyRenderVisibleElements` with **padded viewport culling** we control.
- React Flow has no API to expand the cull viewport (requested upstream in [xyflow#5656](https://github.com/xyflow/xyflow/issues/5656)), so we keep `onlyRenderVisibleElements` off and hide far-off nodes via the `hidden` flag instead.
- Uses a **320px screen-space buffer** around the viewport before hiding nodes/edges, so nodes near the edge stay mounted while still skipping far-off-screen content for performance.


forestileao:
👍

PR #5813, opened 2026-07-02
feat: return conversation messages from Cursor Launch Cloud Agent

WashingtonKK:
## Summary

This does the following:

- Merge the Get Last Message component into Launch Cloud Agent so a single component both launches the Cursor cloud agent and returns its result. 
- Remove the standalone cursor.getLastMessage action and its frontend mapper.
-


shiroyasha:
Looks great!

PR #5815, opened 2026-07-02
fix: stop agent from acting unsolicited on canvas open/refresh

### LINKED ISSUE #5110: Agent should not do things and spend tokens unsolicited
Currently the agent invokes several commands on boot -- this was done in the spirit of helping new users build canvases but the consequence is that agents are doing something, anything whenever you open / refresh a canvas which 1) costs $ 2) is annoying and wrong 3) gives the impression that SP agent is out of control by default.

forestileao:
## Problem

Fixes #5110.

Opening or refreshing a canvas kicked off the agent automatically. The frontend boot-kickoff (`useAgentBootKickoff`) sent a default boot message — `"Session ready. Use the [Canvas Snapshot] ... to greet the user ..."` — to the agent whenever a canvas's chat was empty.

Even though that message told the agent not to run CLI commands, in practice the agent still produced verbose, zero-value openings (see the screenshot in the issue) and spent tokens without the user asking for anything. This made the agent feel out of control by default.

## Fix

Send **no** boot message by default. With no explicit boot context, `getAgentBootMessage` now returns an empty string, so the existing kickoff guard (`if (!bootMessage) skip`) means the agent stays idle until the user actually sends a message.

The user is still greeted: `AgentTabPanel` already prepends a static, client-side greeting (no tokens spent). The solicited flows are unchanged — a template install still sets its own boot context, and blank canvases already sent nothing.

## Testing

- Updated `agentBootContext.spec.ts` to assert no boot message is produced by default.
- `agentBootContext.spec.ts` passes (8/8).
- `eslint` clean on the changed files.

### DIFF
--- a/web_src/src/lib/agentBootContext.ts
+++ b/web_src/src/lib/agentBootContext.ts
@@ -3,14 +3,6 @@ const AGENT_BOOT_INITIAL_MESSAGES_KEY = "agent-boot-initial-messages";
 export const PLACEHOLDER_NODE_CONTEXT_KEY = "add-placeholder-node";
 export const AGENT_BOOT_CONTEXT_READY_EVENT = "agent-boot-context-ready";
 
-// Opening or refreshing a canvas must never invoke the agent. Without an explicit
-// boot context (a template install), we send no message so the agent stays idle and
-// spends no tokens until the user asks for something. A static greeting (rendered
-// client-side in AgentTabPanel) welcomes the user instead.
-const DEFAULT_BOOT_MESSAGE = "";
-
-const BLANK_BOOT_MESSAGE = ""; // No agent boot message for blank canvas — static greeting only
-
 const BLANK_INITIAL_MESSAGE =
   "You can describe the workflow you want to build, or click on the 'New Component' node on the canvas to get started. I'm here to help!";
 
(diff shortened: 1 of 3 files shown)


web_src/src/lib/agentBootContext.ts
```diff
@@ -3,8 +3,11 @@ const AGENT_BOOT_INITIAL_MESSAGES_KEY = "agent-boot-initial-messages";
 export const PLACEHOLDER_NODE_CONTEXT_KEY = "add-placeholder-node";
 export const AGENT_BOOT_CONTEXT_READY_EVENT = "agent-boot-context-ready";
 
-const DEFAULT_BOOT_MESSAGE =
-  "Session ready. Use the [Canvas Snapshot] from the session context to greet the user. Do not run CLI commands or fetch the canvas just to summarize it. Only check connected integrations if the user asks for integration-specific work.";
+// Opening or refreshing a canvas must never invoke the agent. Without an explicit
+// boot context (a template install), we send no message so the agent stays idle and
+// spends no tokens until the user asks for something. A static greeting (rendered
+// client-side in AgentTabPanel) welcomes the user instead.
+const DEFAULT_BOOT_MESSAGE = "";
```
forestileao: maybe we can just remove this?

forestileao: Good call — done in 107aafa10. Both `DEFAULT_BOOT_MESSAGE` and `BLANK_BOOT_MESSAGE` were empty and redundant, so I removed them and `getAgentBootMessage` now returns `""` directly (with the rationale documented on the function).

forestileao:
👍

PR #5825, opened 2026-07-02
fix: ignore superplane expressions in runner scripts

re-pixel:
## Summary

Runner `script` / `commands` fields were still going through SuperPlane `{{ }}` resolution. That produced invalid code (e.g. unquoted timestamps in JS) and fought the runner’s own templating. This PR stops resolving expressions in those fields and leaves `{{ }}` as literal text.

### Approach

- Added `allow_expressions` on `TextTypeOptions` (proto + Go + UI).
- When `allowExpressions: false`:
  - backend skips template resolution (literals stay as-is)
  - Monaco expression autocomplete is disabled
  - validation does **not** reject values that contain `{{ }}` (unlike string fields)
- Set `AllowExpressions: false` on runner `script` / `commands` (JS, Python, Bash, Commands).

This keeps one explicit flag for “don’t treat this text field as SuperPlane expressions.”


shiroyasha:
Does the UI still suggest that you can use `{{` in the expression. e.g. via autocomplete?

Maybe a cleaner design would be to have a optinal configuration value that disables expression evaluation, that way both the frontend and the backend can handle it, and it would be explicitly defined.

protos/configuration.proto
```diff
@@ -55,6 +55,11 @@ message TextTypeOptions {
   optional int32 min_length = 1;
   optional int32 max_length = 2;
   optional string language = 3;
+  // When false, SuperPlane does not resolve expression placeholders in this
+  // field (they are left as literal text) and the UI does not offer expression
+  // autocomplete. Unlike StringTypeOptions.allow_expressions=false, this does
+  // not reject values that contain expression placeholders.
+  optional bool allow_expressions = 4;
```
forestileao, line 75: I think we can remove this big comment from here leave a small comment in `configuration/field.go`.

If you are using claude it generally likes to add these giant amount of comments that could be just one small line

forestileao:
👍

PR #5826, opened 2026-07-02
fix: preserve field value when toggling between Fixed and Expression

### LINKED ISSUE #3624: Toggling between 'Fixed' and 'Expression' type values clears previously entered content

forestileao:
## Summary

Fixes #3624. Toggling the **Fixed** / **Expression** value-type tabs on an
integration-resource field no longer clears the content the user already entered.

Previously, `IntegrationResourceFieldRenderer` called `onChange(undefined)` on
every mode switch, wiping the input. The tab handler now only changes the display
mode and leaves the value intact, so it carries over between Fixed and Expression
(both inputs are backed by the same `value`).


forestileao:
👍

PR #5832, opened 2026-07-02
fix: keep expression autocomplete anchored to input on scroll

### LINKED ISSUE #3615: Data autocomplete doesn't stick to parent input field
https://github.com/user-attachments/assets/5425e818-1c90-4f71-95b1-5bae53da4c04

forestileao:
## Summary
Fixes #3615. The expression autocomplete dropdown is rendered in a portal with `position: fixed`, positioned from the input's `getBoundingClientRect()`. That position was only recomputed when the value/cursor changed, so scrolling the surrounding panel (or resizing the window) left the dropdown floating in its old spot, detached from the input.

- While the dropdown is open, listen for `scroll` (capture phase, so scrolling of any scrollable ancestor is observed) and `resize`, and recompute the dropdown position.
- Throttle repositioning with `requestAnimationFrame` and clean up listeners when the dropdown closes.


forestileao:
👍

PR #5833, opened 2026-07-02
fix: fit view to whole graph when switching canvas or version

### LINKED ISSUE #4025: Fit view on canvas or version switch
**Problem:** Switching to another Canvas or another version reuses the previous viewport (pan/zoom), so the graph often opens off-center or zoomed incorrectly. Fit-to-view already exists (e.g. toolbar / keyboard) but is not applied on context switch.

**Current vs proposed:** Current: viewport persists from the last canvas/version. Proposed: after switching Canvas or version, run the same fit-to-view used on first load so the full workflow is visible.

**Use case:** Anyone moving between canvases or history versions who expects to see the whole graph immediately without manually fitting the view.

forestileao:
## Summary

Fixes #4025. Switching to another canvas or version reused the previous
viewport (pan/zoom), so the graph often opened off-center or zoomed wrong.
Now a context switch re-runs the same fit-to-view used on first load.

**Root cause:** `CanvasPage` already remounts on a canvas/version switch (via
`canvasRenderKey`), but the fit-state refs (`hasFitToViewRef`, `viewportRef`)
are owned by the parent and persist across remounts. So `handleInit` took the
"restore previous viewport" branch instead of fitting the new content.

## Changes

- `web_src/src/ui/CanvasPage/fitView.ts` (new): pure helpers —
  `computeFitViewContentKey`, `shouldRefitOnInit`, `stampFittedContentKey`.
- `web_src/src/ui/CanvasPage/index.tsx`: `handleInit` re-fits when the displayed
  content key changes (not only on first init), and only stamps the key once real
  nodes were fitted, so a still-loading switch re-fits when its nodes arrive.
  Added `fitViewContentKey` / `lastFittedContentKeyRef` props.
- `web_src/src/pages/app/index.tsx`: owns `lastFittedContentKeyRef` (survives
  remounts) and passes the content key. Returns `undefined` in run-inspection so
  that mode keeps its dedicated fit/viewport handling. Also compacted the adjacent
  `canvasRenderKey` construction into an equivalent single-line template literal to
  stay within the existing ESLint line budget (no baseline change).

The decision lives inside `handleInit`, avoiding effect-vs-`onInit` ordering races.

### DIFF
--- a/web_src/src/pages/app/index.tsx
+++ b/web_src/src/pages/app/index.tsx
@@ -72,7 +72,7 @@ import { getActiveNoteId, restoreActiveNoteFocus } from "@/ui/annotationComponen
 import { buildBuildingBlockCategories } from "@/ui/buildingBlocks";
 import type { CanvasNode, NewNodeData, NodeEditData, SidebarData } from "@/ui/CanvasPage";
 import { CANVAS_SIDEBAR_STORAGE_KEY, CanvasPage, type MissingIntegration } from "@/ui/CanvasPage";
-import { computeFitViewContentKey } from "@/ui/CanvasPage/fitView";
+import { resolveFitViewVersionId } from "@/ui/CanvasPage/fitView";
 import type { EventState, EventStateMap } from "@/ui/componentBase";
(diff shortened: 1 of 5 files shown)


forestileao:
👍

PR #5862, opened 2026-07-03
fix: derive AWS create-instance node state from emitted channel

### LINKED ISSUE #5857: AWS create machine displays false state
Example here:

1/ It says that the machine was created
2/ But, the output went to the failed channel
3/ And the machine was not created

Image

WashingtonKK:
Closes https://github.com/superplanehq/superplane/issues/5857

## Summary

This fixes the display state on the aws.createInstance component.

## Demo

image


shiroyasha:
Looks great! Thanks for the quick fix.

PR #5874, opened 2026-07-03
fix: open agent panel (not history) on new apps with a per-canvas preference

### LINKED ISSUE #5803: New app screen should open agent chat, not canvas history
We [previously](https://github.com/superplanehq/superplane/issues/5110) complained that agent is doing too much "analysis" and spending tokens unsolicited on an empty canvas.

Since then we overcorrected on the side of not showing the agent at all, which is wrong too, because the agent is not easy to discover this way for first-time users and it signals that it's a nice-to-have, not essential tool.

We display version history which is useless as a starting point:

Image

The agent panel should open up and the agent should present itself with a pre-defined message, without analyzing or doing anything.

### Update

The agent being open/closed is actually a global localstorage preference.

What we want:

- Don't show history in any case.
- Store agent panel preference of being open/closed _per canvas_.
- Always start a new app with open agent panel.

forestileao:
## Summary

Fixes #5803.

Opening a new app defaulted to auto-expanding the canvas **version/history** sidebar — which is useless as a starting point — while the **agent panel** (the primary way to build a canvas) was hidden behind a single **global** open/closed preference shared across every canvas. This made the agent hard to discover and signalled it was a nice-to-have.

### What changed

Implements the three points from the issue's "What we want":

1. **Don't show history as the initial view.** The versions/history sidebar no longer auto-opens on entering an edit session; it starts collapsed and is only shown via the header toggle.
   - `useCanvasVersionsSidebarState`: default open state `true` → `false`.
   - `CanvasPage`: removed the `useEffect` that force-opened the versions sidebar whenever an edit session became active.
2. **Store the agent panel open/closed preference per canvas.** `useCanvasToolSidebarState` now reads/writes `canvasAgentSidebarOpen:<canvasId>` (falling back to the legacy global key only when no canvas id is available) and re-reads the preference when navigating between canvases.
3. **Always start a new app with the agent panel open.** `useCreateApp` and `useInstallAction` write the per-canvas open preference (`true`) for the freshly created/installed canvas.

The blank-canvas agent still only presents its predefined greeting and does not auto-send/analyze (unchanged; see `agentBootContext.ts`), matching the issue's request that the agent "present itself with a pre-defined message, without analyzing or doing anything".

## Test plan

- [x] `useCanvasToolSidebarState.spec.tsx` — added a test asserting the open state is read and persisted **per canvas** (toggling one canvas leaves another's preference untouched).
- [x] `useCanvasVersionsSidebarState.spec.tsx` — updated to assert the sidebar starts collapsed.
- [x] `make check.build.ui` passes; eslint on touched files reports no new errors.
- [ ] Manual (feature `claude_managed_agents` enabled): create a new app → agent panel opens with its greeting, version history is not shown; toggle history open, reload → each canvas remembers its own agent panel state.

### DIFF
--- a/test/e2e/shared/canvas_steps.go
+++ b/test/e2e/shared/canvas_steps.go
@@ -79,11 +79,10 @@ func (s *CanvasSteps) ExitEditMode() {
 	s.session.Sleep(500)
 }
 
-// OpenVersionsSidebar reveals the versions sidebar. The sidebar no longer
-// auto-opens when an edit session starts (issue #5803): it starts collapsed and
-// is shown via the header toggle. If no edit session is active yet, this enters
-// edit mode (selecting the latest draft or creating one) to make the toggle
-// available, then clicks it to reveal the sidebar.
+// OpenVersionsSidebar reveals the versions sidebar, which is shown while an edit
+// session is active (unless the agent panel is the active side panel). If no
+// edit session is active yet, it enters edit mode (selecting the latest draft or
+// creating one) to reveal the sidebar.
 func (s *CanvasSteps) OpenVersionsSidebar() {
 	sidebar := q.TestID("canvas-versions-sidebar").Run(s.session)
 	if visible, _ := sidebar.IsVisible(); visible {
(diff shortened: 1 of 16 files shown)


forestileao:
👍

PR #5875, opened 2026-07-03
fix: Create And Sort Secrets From the Canvas

### LINKED ISSUE #5835: UX: sort secret picker alphabetically
The secret picker/dropdown should sort secrets alphabetically so users can quickly find what they need.

andrecalil:
## Summary
- Sorts the secret picker alphabetically
- According to the user's permissions, appends the "+ Add a new secret" option to the secret picker
- Launches a dialog to create a secret on the fly, becomes the selected option
- Extra: summarized the description for the _SSH > Command file_ field.

Resolves: #5835


pkg/components/ssh/ssh.go
```diff
@@ -138,7 +138,7 @@ Choose **SSH key** or **Password**, then select the organization Secret and the
 - **Host**, **Port** (default 22), **Username**: Connection details.
 - **Command source**: Choose **Inline** to type commands directly, or **From file** to load them from a file in the app's repository (e.g. scripts/deploy.sh).
 - **Commands** (inline mode): One or more commands to run, one per line (supports expressions). Each non-empty line becomes a command joined with &&. The output payload is based on the last command.
-- **Command file** (file mode): Path to a file in the Files tab.
```
shiroyasha, line 141: This seems unrelated to this change, no?

andrecalil: Since I removed the verbose description from the UI, I added it to PRD, so it's not lost until updated on the official docs.

shiroyasha: Ack.

1/ Lets move it to a separate PR. Not good to mix unrelated changes.
2/ The current version reads more like an explanation why the regular commands in the SSH commands are bad and how files are fixing them. Consult with @markoa or @AleksandarCole on this one as this is going into the public docs.

shiroyasha:
@andrecalil I really wanted this change in production and I would hate for it to become outdated. I told Cursor to extract the docs change that still needs to be reviewed into another PR so we can instantly merge this one.

shiroyasha:
👍

PR #5899, opened 2026-07-04
fix: use timestamp fallback for versions without commit messages

lucaspin:
Replace the generic "Commit" label in version history with Update from {createdAt} for legacy versions missing a commit message


lucaspin:
👍

PR #5904, opened 2026-07-04
fix(ui): stabilize transitions after commit

lucaspin:
Fixes post-commit regressions where the UI briefly showed the pre-commit version, fired /repository/file requests with a stale version_id (500s), and duplicated cache refetches.

• Commit flow: Exit edit mode first (flushSync), then cancel/remove queries for the pre-commit version and run a single coordinated invalidation — no pre-exit cache sync.
• Websocket lifecycle: Stop version-scoped invalidations/resyncs on canvas_updated while editing; refresh staging metadata only.
• Live console reads: Omit version_id for live console fetches so the server resolves the current live file after the version id rotates.
• Cleanup: Replace useCanvasDraftResync / resyncDraftToCommitted with useCanvasStagingResync (staging-only resync for cross-tab staging_updated).


lucaspin:
👍

Issue #5884, opened 2026-07-04
Bind secret crypto to ID not name


shiroyasha:
**Problem:** Local org secrets use the mutable secret **name** as AES-GCM associated data (AD). Renaming a secret therefore requires re-encrypting stored data so AD stays in sync with the DB row. Integration secrets already bind crypto to a stable identifier (`installation_id`). Using a user-changeable label as AD is unconventional and caused #5877 (HTTP 500 after rename + key delete).

**Current vs proposed:**
- **Current:** `encryptor.Encrypt/Decrypt(..., []byte(secret.Name))` across create, read, update-key, delete-key, rename, and worker decrypt paths.
- **Proposed:** Use immutable `secret.ID` (UUID string) as AD for all new encrypt/decrypt operations. Renames become metadata-only changes with no re-encryption.

**Use case:** Operators and users can rename secrets freely without crypto side effects. Reduces surprise failures in production (where `NO_ENCRYPTION` is off) and aligns with how integration secrets are already handled.

**Suggested implementation:**
1. Centralize AD selection in `encryptSecretData` / `decryptSecretData` (e.g. `secretAssociatedData(secret)` → `secret.ID.String()`).
2. **Backward compatibility:** On decrypt, try `secret.ID` first; if authentication fails, fall back to `secret.Name` for legacy rows.
3. **Migrate on write:** Any operation that re-encrypts secret data (set/delete key, update) writes blobs with ID-based AD.
4. Optional one-time migration job to re-encrypt all remaining name-bound rows.
5. Remove rename re-encryption workaround in `UpdateSecretName` once migration is complete.
6. Update E2E `assertSecretSavedInDB` / `givenASecretExists` to use ID-based AD when `NO_ENCRYPTION` is off.

**Affected call sites (non-exhaustive):**
- `pkg/grpc/actions/secrets/create_secret.go`
- `pkg/grpc/actions/secrets/describe_secret.go`
- `pkg/grpc/actions/secrets/set_secret_key.go`
- `pkg/grpc/actions/secrets/delete_secret_key.go`
- `pkg/grpc/actions/secrets/update_secret_name.go`
- `pkg/workers/contexts/secrets_context.go`
- `test/e2e/secrets_test.go`

**Related:** #5877 (immediate fix: re-encrypt on rename while name remains AD).

PR #5915, opened 2026-07-05
fix: support id-bound local secret data

ahfoysal:
Updates local secret encryption helpers to prefer secret ID associated data while still reading legacy name-bound payloads.

Re-encrypting secret data through update, key changes, or rename now writes ID-bound data, and worker/API reads share the same fallback path.

Refs #5884

### DIFF
--- a/pkg/grpc/actions/secrets/create_secret.go
+++ b/pkg/grpc/actions/secrets/create_secret.go
@@ -2,7 +2,6 @@ package secrets
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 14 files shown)


shiroyasha:
@ahfoysal nice! Thanks for tackling this.

The risk review has pointed out some valid concerns.

ahfoysal:
Thanks, agreed. I’ll clean this up in the PR rather than leave it as a partial fix.

I’m going to address the creation path so new local secrets are encrypted with the immutable ID, keep the legacy name fallback only for existing data, and remove any dead helper code while I’m there. I’ll push a follow-up commit with tests.

ahfoysal:
Pushed a follow-up commit for the risk-review points.

What changed:
- new local secrets now get a UUID before insert and encrypt data with that immutable ID
- legacy name-based decrypt fallback is still supported, but both decrypt failures are surfaced if ID and legacy fallback fail
- removed the unused create helper
- rename now only re-encrypts when it actually decrypted legacy name-bound data, so already ID-bound data stays unchanged
- added focused tests for create-time ID binding, fallback error reporting, and rename behavior

Local checks run:
- `go test ./pkg/secrets`
- `go test ./pkg/grpc/actions/secrets -run TestDoesNotExist`
- `git diff --check`

PR #5922, opened 2026-07-06
fix: warn on unconnected action nodes

### LINKED ISSUE #4751: Warn on action nodes with no incoming edges
## Problem

Action nodes (components) without any incoming edges will never execute — they're dead nodes on the canvas. Currently there's no warning for this, so users can accidentally leave nodes unconnected without realizing they'll never fire.

## Current behavior

- Missing required config fields → warning ✅
- Node with no incoming edges → no warning ❌

## Expected behavior

If an action/component node has **zero incoming edges**, show a warning like:

> This node has no incoming connections and will never be triggered.

This should apply to `TYPE_COMPONENT` nodes only. Trigger nodes are entry points so they don't need incoming edges.

## Why this matters

When edges use the wrong output channel (e.g. `default` instead of `success`), the edge exists in the spec but doesn't actually connect visually or functionally. The downstream node effectively has no valid incoming edges and will never execute. A warning here would immediately surface wiring mistakes.

## Example

A canvas with:
```yaml
nodes:
  - { id: trigger, type: TYPE_TRIGGER, trigger: { name: start } }
  - { id: step-a, type: TYPE_COMPONENT, component: { name: http } }
  - { id: step-b, type: TYPE_COMPONENT, component: { name: http } }  # no edges point here
edges:
  - { sourceId: trigger, targetId: step-a, channel: default }
```

`step-b` should show: ⚠️ "No incoming connections — this node will never be triggered."

forestileao:
Fixes #4751

## Summary
- derive a warning for action nodes with no valid incoming connections
- treat source-channel mismatches as disconnected
- add frontend unit coverage

## Tests
- make format.js
- cd web_src && npm test -- workflowPageHelpers.spec.ts --run
- make check.lint.ui
- make check.build.ui

### DIFF
--- a/web_src/src/pages/app/index.tsx
+++ b/web_src/src/pages/app/index.tsx
@@ -138,7 +138,6 @@ import {
   buildTabData,
   generateNodeId,
   generateUniqueNodeName,
-  mapCanvasNodesToLogEntries,
   getWorkflowSaveSignature,
 } from "./utils";
 import { actionsFromCapabilities, triggersFromCapabilities } from "@/lib/capabilities";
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #5929, opened 2026-07-06
fix: clarify canvas yaml diff highlights

### LINKED ISSUE #5366: Canvas version diff: highlight what actually changed
When viewing a canvas version diff, all lines in a changed block are currently highlighted — even the ones that did not change. This makes it hard to spot what actually changed, especially in larger configs.

## Expected behavior

- Only the lines that actually changed should be colored (green for additions, red for removals)
- Within those lines, the exact characters that changed should be highlighted in a stronger shade (inline diff), similar to how GitHub renders diffs
- Unchanged lines should have no background color

## Current behavior

Entire blocks are highlighted, making it look like everything changed when only a single line or value was modified.

## Example

In the screenshot below, only L2 actually changed, but the entire block is highlighted:

Image

forestileao:
## Analysis
The YAML diff modal used the default diff context styling, so unchanged rows inside a hunk could look highlighted alongside actual additions and removals.

## Summary
- make unchanged context rows neutral
- keep add/remove rows colored and inline word highlights stronger
- add a focused modal options test

## Tests
- cd web_src && npm test -- src/pages/app/CanvasYamlDiffModal.spec.tsx --run
- make format.js
- make check.lint.ui
- make check.build.ui

Closes #5366

### DIFF
--- a/web_src/src/pages/app/CanvasYamlDiffModal.spec.tsx
+++ b/web_src/src/pages/app/CanvasYamlDiffModal.spec.tsx
@@ -1,0 +1,45 @@
+import { render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { MultiFileDiffProps } from "@pierre/diffs/react";
+import { CanvasYamlDiffModal } from "./CanvasYamlDiffModal";
+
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #5932, opened 2026-07-06
fix: split repository file reads by version, stage, or live

lucaspin:
- Refactors repository file reads behind AppFileReader with three modes: version_id (historical), stage=true (staged or live fallback), or live committed
- Fixes historical spec files returning 500 by materializing canvas.yaml / console.yaml from version rows
- Updates gRPC, agents, and the UI to match the new mutually exclusive query params (fixes canvas edits reverting after autosave)


lucaspin:
👍

PR #5935, opened 2026-07-07
fix: Improve If Node Run Status

andrecalil:
## Summary

Rework the run status of the `if` node. Adds a proper resolution pill and only highlights red if error.

### Before
This had successfully ran, but had resolved `false`:
image

### After
CleanShot 2026-07-06 at 22 10 09
CleanShot 2026-07-06 at 22 09 51


AleksandarCole (approved):

andrecalil:
👍🏼

PR #5936, opened 2026-07-07
feat: create apps from folder headers

### LINKED ISSUE #5867: Add "+" button on group header to create a new app directly in that group
## Problem

To create a new app in a specific group, you have to create the app first and then assign it to the group. There is no shortcut from the group view itself.

## Expected Behavior

On the home page, each group header should have a "+" icon next to the settings menu. Clicking it creates a new app that is automatically assigned to that group.

forestileao:
## Summary
- add a plus button to each folder header on the home page
- create a blank app, assign it to that folder, then open it in edit mode
- cover the folder-header create flow with a home-page regression test

Closes #5867

## Tests
- docker compose -f docker-compose.dev.yml exec app bash -c "cd web_src && npm run test:run -- src/pages/home/index.spec.tsx"
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #5956, opened 2026-07-07
fix: show agent staged canvas changes without leaving edit mode

lucaspin:
Two main issues in the agent / staging / edit flow:
- If I am on live, and agent updates staging, I am not switched to edit mode
- If I am on edit, and agent updates staging, I do not see it right away, I have to leave and re-enter edit

### Fix

- Remote `staging_updated` events now reliably update the graph whether the user is on live view or already editing the live version.
- Resync staged spec directly into editor state (`resyncStagedEditorState`) instead of relying on stale React Query caches
- Stop `useDraftCanvasSpecSync` from overwriting live staged drafts with outdated query snapshots
- Avoid a websocket invalidation race that could repopulate `stagedCanvasSpec` with committed content after resync
- Refetch staging summary before auto-entering edit from live view (instead of trusting a warm `hasStaging: false` cache)
- Always refetch staged spec on enter-edit (remove `preferCachedStagedSpec` shortcut)

### Testing

New E2Es added to catch regressions in this area.

### DIFF
--- a/test/e2e/agent_staging_edit_test.go
+++ b/test/e2e/agent_staging_edit_test.go
@@ -1,0 +1,228 @@
+package e2e
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	pw "github.com/playwright-community/playwright-go"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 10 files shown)


lucaspin:
👍

Issue #5947, opened 2026-07-07
Table: Actions, remove non-manual triggers


bender-rodriguez-unit1:
Remove non-manual triggers from the Actions table.

**SuperPlane Board status:** pending

PR #5957, opened 2026-07-08
fix: Hide Non-manual Triggers from Table Actions

### LINKED ISSUE #5951: Single Node Widget: remove non-manual triggers
Remove non-manual triggers from the Single Node Widget.

**SuperPlane Board status:** pending

andrecalil:
## Hide non-manual triggers from action columns

Filters non-manually-runnable triggers out of the two console widgets that expose a Run affordance (single/multi node panel and table row actions), disables every Run button while its triggered run is executing, and merges the single/multi-node widgets into one adaptive component along the way.

Resolves #5951 and #5947.

### What changed

- **Manual-run gate**: only the built-in `start` and `schedule` triggers expose a user-invokable `run` hook, so they're now the only ones offered. Hardcoded allowlist in `web_src/src/pages/app/console/manualRunTriggers.ts`; UI filters on `node.component` in `TablePanelForm` (dropdown), `WidgetTable` (row actions filter + gate), `NodesPanelCard` / `NodesPanelForm` (Run affordance), and `useConsoleRunTrigger` (disabled reason). Backend authorization in `InvokeNodeTriggerHook` remains the source of truth.
- **In-flight lock**: new shared `useConsoleTriggerLock` powers per-widget submission tracking. The Run button in the node panel and each row-action button in `WidgetTable` disable while their trigger has a `STATE_STARTED` run, plus a short grace window bridging the invoke → websocket-refresh gap. `WidgetTableActionLock` is now a thin wrapper over the shared hook; `useConsoleRunTrigger` consumes the same lock so sibling buttons targeting the same trigger lock together.
- **Merged node widget**: `NodePanelCard.*` (component + spec + stories, ~570 LoC) deleted. `NodesPanelCard` is now adaptive — compact centered layout for one entry, row list for many, and folds legacy `type: "node"` panel content into a one-entry list on load so existing dashboards keep working. Form logic extracted into `NodesPanelForm.tsx` to stay under the lint budget. `panelTypes`, `useConsolePanelState`, `nodesPanelContent`, `ConsolePanelCards`, and `ConsoleView` updated accordingly with backwards-compat migration.
- **Confirm dialogs**: `RowActionConfirmDialog` and `NodeRunConfirmDialog` reflect the disabled state so an operator can't double-fire while a run is already in flight.
- **Docs**: `docs/prd/console-and-widgets.md` and `.cursor/skills/superplane-dashboard-and-widgets/SKILL.md` updated to describe the merged widget and the manual-run allowlist.

### Not touched

No backend or proto changes — an earlier iteration added a `manual_runnable` proto field + `useTriggers()` catalog fetch + `ConsoleContext` plumbing, but that was reverted in favor of the hardcoded frontend allowlist since only two built-in triggers are user-runnable and adding a new one already requires a UI PR.

### Verification

`make pb.gen`, `make format.js`, `make check.lint.ui` (601/601 within budget), `make check.build.ui`, `npm run test:run -- src/pages/app/console` (568/568 pass), `gofmt`, `revive`, `go build cmd/server/main.go` — all clean.


pkg/grpc/actions/common.go
```diff
@@ -1056,18 +1056,30 @@ func SerializeTriggers(in []core.Trigger) []*triggerpb.Trigger {
...
-			Configuration: configuration,
-			ExampleData:   exampleData,
+			Name:           trigger.Name(),
+			Label:          trigger.Label(),
+			Description:    trigger.Description(),
+			Icon:           trigger.Icon(),
+			Color:          trigger.Color(),
+			Configuration:  configuration,
+			ExampleData:    exampleData,
+			ManualRunnable: triggerHasManualRunHook(trigger),
```
lucaspin: Why not use the name of the trigger?

andrecalil: Done! Much better

lucaspin (approved):

PR #5961, opened 2026-07-08
fix: make edit mode state more noticeable

lucaspin:
Edit / live modes look to similar when you don't have anything staged. It needs to be clearer:
- Reset / commit buttons are not hidden, but disabled if you are editing, but have no staged changes
- Header tabs change color
- E2Es are updated to verify this behavior, so we don't regress here anymore


lucaspin:
👍

PR #5999, opened 2026-07-08
feat: App Default Tab

### LINKED ISSUE #5868: Default tab when navigating to an app should remember last visited tab
## Default tab when navigating to an app

- Store on the backend the last tab visited by the user on that app
- Upon navigating to an app, default to the last visited tab
- Fallback to: if the app has widgets on the console, default to the console. Otherwise, default to the canvas

andrecalil:
## Default tab when navigating to an app

When a user opens an app (`/:orgId/apps/:appId`), we now remember which tab they were last on and land them there again. The rules:

1. If we have a stored last-visited tab for that user on that app, go to it.
2. Otherwise, if the app has widgets on its Console, default to Console.
3. Otherwise, default to Canvas.

Explicit links that include `?view=` or `?run=` always win — the default-tab logic is a mount-time one-shot, so hitting back to Canvas doesn't bounce the user back into Console.

## Why

Right now every app opens on Canvas regardless of how the user actually uses it. Teams whose day-to-day is Console, Memory, or Files pay a click every time they open an app. Persisting the last tab per user makes the app open where they left off, and the Console fallback gives apps built around a dashboard a sensible first-run default.

## How

App tabs are already URL-driven (`?view=console|memory|files`, absent = Canvas), so the frontend didn't need a new state model — just a way to learn the user's last tab and record changes.

**Storage.** The last-visited tab is kept client-side in `localStorage` under `superplane:last-visited-app-tab`, keyed by canvas id. This mirrors `lastVisitedOrganization` and keeps state the backend doesn't otherwise consume out of the API. No proto changes, no DB migration, no new endpoint or authorization rule.

**Frontend.** A new `lib/lastVisitedAppTab.ts` helper reads/writes the map and validates values against the allowed tab set. A new `useDefaultAppTab` hook, mounted from `AppPage`, applies the redirect once per mount and records tab changes as the user navigates.

## Notes

- Tab writes are synchronous and best-effort — storage errors are swallowed.
- Run inspection (`?run=`) is treated as a sub-mode of Canvas and is not recorded as its own tab.
- Closing a run inspection lands the user on Canvas without them picking a tab, so that landing is not persisted.
- The Console fallback still waits for the live console query; if the user switches tabs while it is loading, their choice wins.

Resolves #5868


db/migrations/20260708185214_add-last-visited-tab-to-user-canvas-preferences.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE public.user_canvas_preferences
+  ADD COLUMN IF NOT EXISTS last_visited_tab text;
```
lucaspin: This feels more like a local storage thing, no?

andrecalil: It could be, yes. When we talked about this in a meeting, we went over it with the idea of a backend solution. I can definitely simplify this a lot if we want to make frontend only.

lucaspin (approved):

PR #6003, opened 2026-07-09
fix: use canvas update permission for staging

forestileao:
## Summary
- replace the obsolete canvases:update_version permission with canvases:update for staging authorization
- remove the separate draft-edit permission from the roles UI
- update agent scoped permissions, access reporting, and tests for the unified canvas update permission

## Verification
- make format.js
- npm run build (from web_src)
- go test ./pkg/authorization -run TestCanvasAuthorizationRulesUseCanvasUpdateForStaging
- go test ./pkg/agents -run TestAgentTokenScopes
- go test ./pkg/agents/agent_tools/actions -run TestAccessAction_ReportsInterceptorBackedAgentTokenAccess

## Notes
- Docker-backed checks could not run locally because the Docker daemon/socket was unavailable: make format.go, make check.build.ui, make lint, make check.build.app.


forestileao:
@lucaspin I made this PR, but it will also be required to remove the publish permission and create a "commit" one.
So the agent cannot commit too. If you see any other edge case let me know

lucaspin:
@forestileao why? Can't we use just `canvases:update` for now?

> So the agent cannot commit too

If there is no tool for the agent to commit, the agent can't commit, right? We don't need to do that on permission level

forestileao:
👍

PR #6010, opened 2026-07-09
fix: refine run inspector input timeline behavior

forestileao:
## Summary
- keep the Run Inspector input timeline status fixed as purple Triggered
- expand JSON by default in right-sidebar modals while preserving inline collapsed JSON
- add regression coverage for the input status and modal JSON behavior

## Tests
- make format.js
- cd web_src && npm run test:run -- src/ui/Runs/RunInspectorPanel.spec.tsx
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #6016, opened 2026-07-09
fix: fit all run inspection nodes

forestileao:
## Summary
- fit run inspection participant requests against the full requested run node set
- remove the minimum zoom clamp from run participant fitting so spread-out run nodes remain visible
- add regression coverage for fitting multiple run participant nodes

## Tests
- cd web_src && npm run test:run -- src/ui/CanvasPage/canvasFitOptions.spec.ts src/ui/CanvasPage/index.runParticipantFit.spec.tsx src/pages/app/useRunParticipantFitRequest.spec.ts src/ui/CanvasPage/index.runInspection.spec.tsx
- make check.lint.ui
- make check.build.ui

### DIFF
--- a/web_src/.eslint-budget-baseline.json
+++ b/web_src/.eslint-budget-baseline.json
@@ -1,10 +1,10 @@
 {
-  "maxAllowedTotalIssues": 599,
+  "maxAllowedTotalIssues": 594,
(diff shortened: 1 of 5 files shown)


forestileao:
👍

PR #6020, opened 2026-07-09
fix: speed up live canvas run inspection

forestileao:
## Summary
- open live canvas node run inspection from already-loaded recent runs when possible
- cap live-node fallback run lookup to the newest runs page instead of walking historical pages
- add regression coverage for cached node lookup and capped pagination

## Tests
- npm test -- --run src/pages/app/sidebarRunLookup.spec.ts src/hooks/useSidebarEventRunLookup.spec.ts
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #6023, opened 2026-07-09
fix: improve run inspector runtime actions

forestileao:
## Summary
- add node header edit access from run inspector runtime config
- preserve multiline runtime config values and hide empty outputs
- delay error scrolling until nested accordions are open
- hide cancelled approval actions and switch to the created rerun

## Verification
- make format.js
- cd web_src && npm run test -- src/ui/Runs/RunInspectorPanel.spec.tsx
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #6026, opened 2026-07-10
fix: improve run inspector navigation controls

forestileao:
## Summary
- add run inspector chrome controls for navigating newer/older loaded sidebar runs
- add copy-link action for the selected run
- preserve filtered sidebar run order for inspector navigation
- collapse long run summary values and respect dismissed inspector auto-open state

## Validation
- npm --prefix web_src run test -- src/ui/Runs/RunInspectorChrome.spec.tsx src/ui/Runs/RunInspectorPanel.spec.tsx src/components/CanvasToolSidebar/RunsTabPanel.spec.tsx
- make format.js
- make check.lint.ui
- make check.build.ui


forestileao:
👍

PR #6037, opened 2026-07-10
fix: open old node run inspector

forestileao:
## Summary
- add run_id to canvas event and node execution API messages
- serialize run IDs from live canvas status activity
- use direct sidebar event run IDs before falling back to ListRuns lookup

## Validation
- cd web_src && npx vitest run src/pages/app/runInspectionLiveNodeLookup.spec.ts src/pages/app/utils.spec.ts
- go test ./pkg/grpc/actions/canvases -run Test__SerializeCanvasEvent__ConvertsJSONNumbersForProto
- cd web_src && npm run build
- git diff --check

## Notes
- Docker was not running locally, so Docker-backed make targets could not be executed in this environment.


forestileao:
👍

PR #6045, opened 2026-07-10
feat: preview applied runtime config expressions

forestileao:
## Summary

Adds applied runtime configuration previews in the run inspector so expression-backed fields are easier to debug from historical runs.

## Details

- Shows applied runtime config by default and lets users switch back to the expression template with `Show expression`.
- Highlights resolved expression segments inline as `{{ resolved value }}` without repeating the global execution error.
- Uses the committed canvas version from `run.versionId` so the expression view reflects the configuration that actually produced the run.
- Shows field-specific expression resolution errors below the affected readonly field.

## Validation

- `npm run test:run -- RunInspectorPanel.spec.tsx`
- `make format.js`
- `make check.build.ui`


forestileao:
👍

PR #6056, opened 2026-07-12
chore: update @hey-api/openapi-ts

lucaspin:
- Bump @hey-api/openapi-ts to ^0.99.0
- Remove deprecated @hey-api/client-fetch (bundled in openapi-ts since v0.73)
- Keep throwOnError: true in codegen config
- Use baseUrl=false in openapi-ts config so API requests stay relative


lucaspin:
👍

PR #6057, opened 2026-07-12
chore: update jsdom to v29

lucaspin:
No app code changes; this only affects the Vitest DOM environment. Only test code change was relaxing the accessible-name matcher (Triggered\s+Input → Triggered.*Input) in RunInspectorPanel.spec.tsx, for jsdom v29’s a11y tree behavior


lucaspin:
👍

PR #6075, opened 2026-07-13
fix: React Exception on Version Git Commit

### LINKED ISSUE #6072: Error: Objects are not valid as a React child (found: object with keys {id, name, type}). If you meant to render a collection of children, use an array instead.
[View in Sentry](https://superplane.sentry.io/issues/7604168477/events/5d6cd0a57bf645d49b0679819722ee9c/)

## Summary

- flushSync-driven re-render after commit re-triggers a node's 'object as React child' bug outside any error boundary's recovery window, crashing the app instead of showing a fallback.
- Resolved sourcemap shows crash inside flushSync() at index.tsx:3136 in handleCommittedVersionId.
- PR #5904 body states it wraps post-commit state exit in flushSync to fix transition glitches.
- Breadcrumbs show CanvasPage nodes 'n_reviewers' and 'n_publish_comment' already threw the same React error #31 seconds earlier, caught by CanvasNodeErrorBoundary.
- Fetch to integrations/.../resources?type=model matches IntegrationResourceRef{id,name,type} shape reported in the error.
- CanvasNodeErrorBoundary resets hasError once node props change, letting the same bad render happen again during the commit flush, this time uncaught (handled: no).

## Root Cause

- **File:** [`web_src/src/pages/app/index.tsx → handleCommittedVersionId()`](https://github.com/superplanehq/superplane/blob/8cc8ce56e1d2896413f3e9eda9e1319df7b4ad73/web_src/src/pages/app/index.tsx#L3136)
- **Pull Request:** [fix(ui): stabilize transitions after commit #5904](https://github.com/superplanehq/superplane/pull/5904)

## Suggested Fix

Fix the underlying node renderer to display resource.name instead of the raw IntegrationResourceRef object, and make the post-commit flushSync render resilient (retain error-boundary fallback across the flush) so a single bad node can't crash the whole app.

```diff
web_src/src/pages/app/index.tsx:3136-3141
- flushSync(() => {
-   setEditSessionActive(false);
-   setActiveCanvasVersion(null);
-   setDraftCanvasSpec(null);
-   draftCanvasSpecsRef.current.clear();
- });
+ // Guard against a node render bug throwing inside the synchronous flush
+ // and escaping error boundaries; also fix root cause: render resource.name,
+ // not the raw {id,name,type} resource object, in node metadata/labels.
```

---
_Auto-generated by the [Sentry Root Cause Analysis](https://app.superplane.com/3ee1aa47-3a60-4c1f-b645-0b9859ab91f8/apps/06b7d04f-5584-46d7-9424-ef79fdfe60ca) app on [SuperPlane](https://superplane.com)._

andrecalil:
## Summary

- Fixes a canvas crash where some nodes stored full integration resource objects (`{id, name, type}`) in metadata and tried to render them as React children.
- That error was sometimes caught by the node error boundary, then resurfaced uncaught after commit when the canvas re-rendered synchronously.
- Nodes now show the resource name instead, and a single bad node keeps its fallback across that post-commit refresh so it can’t take down the whole canvas.

Resolves #6072


web_src/src/lib/integrationResourceLabel.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Coerces an integration-resource configuration value to a display string.
+ *
+ * Configuration fields typed as `integration-resource` are usually stored as
+ * id/name strings, but agent staging and some write paths can leave the full
+ * IntegrationResourceRef object `{ id, name, type }` in node configuration.
```
lucaspin, line 6: It seems like this is the real bug. Having to code in the UI to handle it would be hiding it, no? Why not fix the underlying problem?

PR #6077, opened 2026-07-13
fix: Console Chart Reserved Keywords

### LINKED ISSUE #5230: Dashboard: Chart widget crashes when a series key collides with a reserved DOM prop (e.g. 'style')
## Bug

Using a reserved React/DOM prop name (e.g. `style`, `class`) as a chart series key causes the page to crash with a React error.

## Error

```
React error #62: The `style` prop expects a mapping from style properties to values, not a string.
```

Stack trace points to Recharts SVG rendering (`path` / `svg`).

## Root Cause

`style` is a reserved React DOM prop. When it is used as a Recharts `dataKey`, it collides with the SVG element's `style` attribute, causing React to throw.

## Steps to Reproduce

1. Create a chart widget with a data series keyed on `style` (e.g. a field named `style` in the data source)
2. Render the chart
3. Page crashes with React error #62

## Expected Behavior

Chart renders the `style` series like any other data key.

## Workaround

Rename the field to avoid the collision (e.g. `style_count`), or use `seriesField: type` with long-format data.

## Suggested Fix

In the chart widget layer, sanitize or auto-prefix series keys that collide with reserved DOM props (`style`, `class`, `id`, etc.) before passing them to Recharts.

andrecalil:
## Summary
- Chart widgets crash when a series is named like a reserved DOM prop (e.g. `style`)
- Safe keys are used under the hood so those series render normally; labels stay as authored

Resolves #5230


forestileao (approved):

andrecalil:
bugbot review

andrecalil:
👍🏼

PR #6082, opened 2026-07-13
fix: add dark mode colors to graphql icon

### LINKED ISSUE #4483: GraphQL component color is darker than the rest of the core components

DimitrijeGlibic:
## What changed

The `graphql.svg` icon had a hardcoded near-black stroke/fill (`#111827`), so it was barely visible in dark mode. I added a `<style>` block inside the SVG itself with a `@media (prefers-color-scheme: dark)` override, switching between `gray-500` (light) and `gray-400` (dark) to match the color convention already used for the other fallback icons in the building blocks sidebar.

## Why this approach

This icon isn't rendered in just one place — it's used via a plain `` in several components (component header, building blocks sidebar, event source, etc.). Fixing the colors in the SVG file itself means every one of those places gets the fix, instead of patching each call site individually or duplicating the icon markup.

I initially looked at inlining the SVG as a React component so it could use `currentColor` and follow the app's theme classes exactly, but that's more invasive for what is essentially a single branded icon, and turned out to be unnecessary: our `ThemeProvider` already sets `color-scheme` on the root document when the theme preference changes, so the browser resolves `prefers-color-scheme` for this SVG the same way it resolves it for the rest of the page. 

I tested switching the in-app light/dark/system toggle and the icon updates live, correctly following the app's theme rather than just the raw OS setting.

Screenshot 2026-07-13 at 23 02 37

Screenshot 2026-07-13 at 23 02 52

Closes #4483 

Edit:

Also noticed that this svg icon is slightly larger the rest of icons with cause text in sidebar to be a bit misaligned, as it can be seen on screenshots from above. By converting it to same size this text is now properly aligned:
Screenshot 2026-07-13 at 23 34 24


shiroyasha:
Very nice! Thank you for opening the PR. Running CI.

forestileao (approved):

PR #6084, opened 2026-07-14
fix: cancel runs waiting on deleted approval nodes

### LINKED ISSUE #6081: Run stays active when an approval node is removed while waiting for a hook action
## Summary

A run can get stuck forever if it is waiting on an approval node hook/action and that approval node is later removed from the canvas.

## Reproduction

1. Create a canvas with a flow that includes an approval node.
2. Start a run that reaches the approval node.
3. Let the approval node wait for its hook/action.
4. Edit the canvas and remove the approval node.
5. Observe the original run.

## Actual Behavior

The run never reaches a terminal state. It remains active/running indefinitely even though the node it is waiting on no longer exists in the canvas.

## Expected Behavior

The run should transition to a terminal state when the approval node it is waiting on is removed.

Possible valid outcomes:
- Mark the waiting step/run as cancelled.
- Mark the waiting step/run as failed with a clear reason.
- Automatically unblock/resolve the run according to a defined canvas-edit policy.

## Impact

Users can end up with runs that never complete and cannot accurately represent the current workflow state. This also makes run inspection confusing because the canvas no longer contains the node responsible for the wait.

forestileao:
## Summary
- Fixes #6081 by cancelling active executions when their canvas node is deleted during publish.
- Finishes affected runs when no open work remains, and publishes execution/run updates after commit.
- Adds unit and E2E coverage for deleting an approval node while a run is waiting.

## Verification
- make format.go
- make lint
- make check.build.app
- make test PKG_TEST_PACKAGES=./pkg/grpc/actions/canvases/changesets
- make test PKG_TEST_PACKAGES=./pkg/grpc/actions/canvases
- make test PKG_TEST_PACKAGES=./pkg/workers
- make test.e2e.single FILE=test/e2e/approvals_test.go LINE=50


pkg/grpc/actions/canvases/changesets/canvas_publisher.go
```diff
@@ -372,8 +388,60 @@ func (p *CanvasPublisher) deleteNode(change *Change) error {
...
+func (p *CanvasPublisher) activeExecutionsForNode(node models.CanvasNode) ([]models.CanvasNodeExecution, error) {
+	var executions []models.CanvasNodeExecution
+	err := p.tx.
+		Where("workflow_id = ?", node.WorkflowID).
+		Where("node_id = ?", node.NodeID).
+		Where("state IN ?", []string{models.CanvasNodeExecutionStatePending, models.CanvasNodeExecutionStateStarted}).
+		Find(&executions).
+		Error
+	if err != nil {
+		return nil, err
```
forestileao: this query should be in models

pkg/grpc/actions/canvases/commit_canvas_staging.go
```diff
@@ -208,6 +211,20 @@ func CommitCanvasStaging(
...
+		}
+	}
+
+	for _, runID := range result.FinishedRunIDs {
+		if err := messages.NewCanvasRunMessage(canvasID.String(), runID.String()).Publish(); err != nil {
+			log.Errorf("failed to publish finished run RabbitMQ message: %v", err)
+		}
+	}
+}
+
```
forestileao: is there any guardrail if my publish canvas execution don't work so we don't fully depends on rabbit mq

pkg/grpc/actions/canvases/changesets/canvas_publisher.go
forestileao: Moved this query into `models.ListActiveNodeExecutions` and reused it from the node deletion cleanup path. The publisher now consumes the deletion result returned by `models.DeleteCanvasNodeWithResult`, so it no longer needs its own stale snapshot.

pkg/grpc/actions/canvases/commit_canvas_staging.go
forestileao: The durable guardrail is that execution/run state is committed to the DB before these RabbitMQ publishes run, so the fix does not depend on RabbitMQ for correctness. These publish calls only push realtime UI updates. If a publish fails, the error is logged and the UI recovers from normal data fetches/reconnect invalidation; the `RunFinalizer` also has a DB sweep for missed RabbitMQ finalization paths.

pkg/models/canvas_node.go
```diff
@@ -507,3 +523,76 @@ func FindNodeQueueItem(workflowID uuid.UUID, queueItemID uuid.UUID) (*CanvasNode
...
+		Model(&CanvasNodeExecution{}).
+		Where("workflow_id = ?", workflowID).
+		Where("id IN ?", executionIDs).
+		Updates(map[string]any{
+			"state":        CanvasNodeExecutionStateFinished,
+			"result":       CanvasNodeExecutionResultCancelled,
+			"cancelled_by": nil,
+			"updated_at":   &now,
+		}).
+		Error
```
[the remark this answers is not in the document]
forestileao: Fixed in the latest push: the bulk update now keeps the active-state predicate in the `UPDATE` itself and uses `RETURNING id`, so `CancelledExecutionIDs` contains only executions actually transitioned to cancelled.

pkg/models/canvas_node.go
```diff
@@ -507,3 +523,90 @@ func FindNodeQueueItem(workflowID uuid.UUID, queueItemID uuid.UUID) (*CanvasNode
...
+
+	finishedRunIDs, err := FinishCanvasRunsWithNoOpenWork(tx, workflowID, runIDs)
+	if err != nil {
+		return DeleteCanvasNodeResult{}, err
+	}
+
+	return DeleteCanvasNodeResult{
+		CancelledExecutionIDs: cancelledExecutionIDs,
+		FinishedRunIDs:        finishedRunIDs,
+	}, nil
```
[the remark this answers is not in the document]
forestileao: Fixed in the latest push. Node deletion now batch-deletes queue items for the deleted node with `DELETE ... RETURNING run_id`, merges those run IDs with the cancelled execution run IDs, and then runs the existing open-work finalization check. I also extended the publisher regression to include a queued item on the deleted node and assert it is removed before the run finishes.

forestileao:
👍

PR #6120, opened 2026-07-14
fix: keep expression autocomplete stable on enter

forestileao:
## Summary

- Keep expression autocomplete open when accepting expandable suggestions with Enter or Tab.
- Preserve highlighted suggestions across stable rerenders.
- Add an e2e regression for Manual Run -> HTTP component URL autocomplete, including preview visibility/remount sampling.

## Root Cause

Keyboard selection did not use the same interaction guard as mouse selection, and `onKeyDownCapture` could queue a stale cursor-position sync before the selected suggestion was applied. In the HTTP URL field this could briefly invalidate the expression context and close/reopen the autocomplete preview.

## Validation

- `cd web_src && npm run test -- AutoCompleteInput`
- `make test.e2e.single FILE=test/e2e/canvas_page_test.go LINE=148`
- `make check.lint.ui`
- `make check.build.ui`
- `make format.js`
- `make format.go`
- `git diff --check`

### DIFF
--- a/test/e2e/canvas_page_test.go
+++ b/test/e2e/canvas_page_test.go
@@ -144,6 +144,31 @@ func TestCanvasPage(t *testing.T) {
 		steps.typeExpression("$")
 		steps.assertAutocompleteNodeSuggestionVisible()
 	})
+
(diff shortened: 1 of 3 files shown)


forestileao:
👍

PR #6130, opened 2026-07-14
feat: support for one-way broadcast messaging between apps

lucaspin:
Adds one-way broadcast messaging between SuperPlane apps in the same organization.

### New components

- New `broadcastMessage` action — publishes a payload to subscribed apps via an async app_messages queue
- New `onBroadcast` trigger — subscribes to another app and emits when a broadcast arrives (includes source app and node metadata)

### Subscriptions and broadcasting

- New canvas_subscriptions table for managing subscriptions between apps. Subscription records are created during onBroadcast.Setup(), through a new ctx.Apps.Subscribe() API.
- We use an outbox pattern for broadcasting the messages. The broadcastMessage component uses a new ctx.Apps.Broadcast() API for creating broadcast messages, which inserts messages into app_messages table. Records in that table are asynchronously handled by a new AppMessageWorker

### UI

New components are exposed in a new SuperPlane building blocks section.


lucaspin:
👍

PR #6138, opened 2026-07-15
fix: ignore stale execution requests

### LINKED ISSUE #6126: Runner poller overrides execution finished_at for already finished executions
## Problem
The runner poller is overriding the `finished_at` timestamp even for executions that are already finished. This corrupts the actual completion time of executions.

## Expected Behavior
The runner poller should not update `finished_at` on executions that already have a terminal state/finished timestamp.

## Impact
Inaccurate execution duration metrics and potentially confusing audit trails.

Image

forestileao:
## Summary
- make terminal execution completion idempotent so late poll/webhook paths cannot re-emit or move completion timestamps
- complete pending execution-scoped requests when an execution finishes, using SKIP LOCKED to avoid blocking in-flight work
- skip execution-scoped request scheduling and hook invocation once the execution is already finished

## Root cause
Runner poll requests could outlive the execution they were scheduled for. If a webhook or another path had already finished the execution, a later request could still enter the execution hook path and touch the finished execution state, which moved the timestamp surfaced as finished_at.

## Validation
- make format.go
- make db.migrate DB_NAME=superplane_test
- make test PKG_TEST_PACKAGES="./pkg/models ./pkg/workers"
- docker compose -f docker-compose.dev.yml run --rm -e DB_NAME=superplane_test app gotestsum --format short --packages="./pkg/workers/contexts" -- -p 1

Fixes #6126

### DIFF
--- a/pkg/models/canvas_node_execution.go
+++ b/pkg/models/canvas_node_execution.go
@@ -398,6 +398,15 @@ func (e *CanvasNodeExecution) Pass(outputs map[string][]any) ([]CanvasEvent, err
 }
 
 func (e *CanvasNodeExecution) PassInTransaction(tx *gorm.DB, channelOutputs map[string][]any) ([]CanvasEvent, error) {
+	finished, err := e.IsFinished(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	if finished {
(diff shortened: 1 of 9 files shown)


forestileao:
👍

PR #6153, opened 2026-07-15
fix: render agent survey questions separately

### LINKED ISSUE #5938: Agent presents multiple questions as multiple choices to choose from
In some instances where the agent has a multiple questions that it needs to ask, it presents them as multiple choices to choose from rather than multiple questions to answer.

Image

forestileao:
## Summary

- detect survey blocks where every bullet option is itself a question
- render those bullets as separate free-text questions instead of one multiple-choice question
- add parser regression coverage for the reported screenshot shape and for normal survey options

## Root cause

The agent can emit a `:::survey` block with one contextual prompt followed by several question-shaped bullets. The parser treated every bullet as an option, so the UI displayed separate clarifying questions as choices for one question.

## Validation

- `npm --prefix web_src run test:run -- src/components/AgentSidebar/widgets/parser.test.ts`
- `make format.js`
- `make check.lint.ui`
- `make check.build.ui`

Fixes #5938

### DIFF
--- a/web_src/src/components/AgentSidebar/widgets/parser.test.ts
+++ b/web_src/src/components/AgentSidebar/widgets/parser.test.ts
@@ -139,6 +139,60 @@ Oops!
     expect(segments[1].type).toBe("error");
   });
 
+  it("keeps normal multiple-choice survey options together", () => {
+    const segments = parseAgentContent(`:::survey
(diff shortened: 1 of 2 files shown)


forestileao:
👍

PR #6154, opened 2026-07-15
fix: wrap long JSON viewer values

### LINKED ISSUE #6062: Long string values (base64 content, URLs) overflow outside the viewer instead of wrapping
Image

Image

forestileao:
Fixes #6062.

## Summary
- Allow shared JSON viewer values to wrap instead of overflowing their container.
- Apply the same wrapping classes to the run inspector custom JSON string renderer.
- Add coverage that the run inspector JSON payload uses the wrapping viewer class.
image


forestileao:
👍

PR #6172, opened 2026-07-16
feat: add structured output to Run Claude Agent and Run Code Agent

felixgateru:
## Summary
- Adds an optional JSON Schema (`outputSchema`) to `claude.runAgent` and `claude.runCodeAgent`, matching the option already on `claude.textPrompt`.
- Managed Agents sessions have no `output_config.format` equivalent (that's a Messages-API-only feature), so this works by appending schema instructions to the task prompt and best-effort parsing the final agent message into a `parsed` field — trusted only when the session completes normally (`idle`).
- Shared logic (`PromptSuffix`, `ExtractJSON`, `ValidateAtSetup`) lives in `pkg/configuration/structuredoutput` so both components (and any future agent component) reuse the same schema validation/config-field code already used by `textPrompt`.
- Frontend mappers for both components now surface a "Structured Output" badge on the node card (config-first, matching the existing `textPrompt` badge pattern to avoid staleness) and the parsed JSON in execution details.

### DIFF
--- a/pkg/configuration/structuredoutput/structuredoutput.go
+++ b/pkg/configuration/structuredoutput/structuredoutput.go
@@ -14,8 +14,7 @@
 // Messages API's output_config.format — there is no server-side grammar
 // constraint for a Sessions-based run. For those, PromptSuffix and ExtractJSON
 // emulate structured output by asking the model to comply and best-effort
-// parsing its final message; callers should only trust the result on a normal
-// completion, never as a real schema guarantee.
+// parsing its final message
 package structuredoutput
 
 import (
--- a/pkg/configuration/structuredoutput/structuredoutput.go
+++ b/pkg/configuration/structuredoutput/structuredoutput.go
@@ -1,20 +1,12 @@
-// Package structuredoutput lets LLM text-prompt components accept a JSON Schema
-// that describes the desired response. The component shows a default schema the
-// user edits; the backend validates that it is well-formed JSON before the
-// provider request is fired, and normalizes it to the constraints both providers
-// enforce:
+// Package structuredoutput lets LLM components accept a JSON Schema
+// describing the desired response, validating it and normalizing it to the
+// constraints strict-mode providers enforce (every object gets
+// "additionalProperties": false; strict mode also requires listing every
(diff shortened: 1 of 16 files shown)


pkg/configuration/structuredoutput/structuredoutput.go
```diff
@@ -9,11 +9,18 @@
 //     "required" to list all of its properties; optional fields are expressed by
 //     making their type nullable in the schema itself.
 //   - Anthropic (strict=false) leaves "required" as the user wrote it.
+//
+// Agent components (Claude Managed Agents sessions) have no equivalent to the
+// Messages API's output_config.format — there is no server-side grammar
+// constraint for a Sessions-based run. For those, PromptSuffix and ExtractJSON
+// emulate structured output by asking the model to comply and best-effort
+// parsing its final message
```
forestileao: Lets remove any specific component reference from the coments, even the existing ones

pkg/configuration/structuredoutput/structuredoutput.go
```diff
@@ -109,6 +116,70 @@ func Prepare(schema map[string]any, strict bool) map[string]any {
...
+var jsonFencePattern = regexp.MustCompile("(?s)```(?:json)?\\s*\\n?(.*?)\\n?```")
+
+// ExtractJSON pulls a JSON object out of free-form text, as produced by an
+// agent asked (via PromptSuffix) to emit structured output in its final
+// message. It tries each fenced code block, last to first, then falls back to
+// parsing the whole trimmed text, and reports false if none of those parse as
+// a JSON object. There is no schema validation here: for agent components
+// there is no server-side grammar constraint to trust, so this is inherently
+// best-effort — callers decide whether to trust it (e.g. only on a normal
+// completion).
```
forestileao: Here too

forestileao (commented):
The code is ok, but claude generally creates these giant comments that could be wrote in 1 or 2 sentences.

Also, on configuration and outside the integration, let's not reference specific integrations. They should not be component specific

forestileao:
👍

PR #6192, opened 2026-07-16
fix: improve textarea contrast in dark mode

### LINKED ISSUE #6137: Black text on dark background is not visible when creating secrets
Not sure whether this is by design to protect secret visibility but the text entered is not visible.

Image

codeaustral-oss:
Closes #6137

## What changed

- Added an explicit dark-mode text color to the shared Textarea component.
- Added a focused regression test that checks the dark-mode class.

## Why

The Create Secret dialog uses the shared Textarea for secret values. The component set a dark background but kept its light-theme text color, making entered values difficult to read in dark mode.

## Verification

- Prettier check passed.
- Focused ESLint check passed.
- Focused Vitest regression test passed on Node 24.
- Git diff validation passed.


shiroyasha:

PR #6193, opened 2026-07-16
fix: apply dark theme on owner setup pages

lucaspin:
In the setup pages, the inner settings card used light-only styles (bg-slate-50, gray text) while the Label component applies dark:text-gray-100. In dark mode that produced light label text on a white box. That generates things like this:

no-dark-theme-on-setup

---

Here, we apply the dark them properly in the owner setup pages.


lucaspin:
👍

Issue #6136, opened 2026-07-15
New component: Run App (trigger another app and wait for completion)


bender-rodriguez-unit1:
## Summary

A new SuperPlane-native component that triggers a run on another app and waits for it to complete before proceeding. Complements the existing inter-app communication components (OnBroadcast, Broadcast Message).

## Behavior

1. Component triggers a run on a selected app
2. Stays in "running" state while the target app run is executing
3. Completes (success/failure) only when the target run finishes

This enables synchronous app-to-app orchestration - one app can invoke another and block until it is done, as opposed to the fire-and-forget pattern of Broadcast Message.

## Settings

- Target app (select from org apps)
- Input data / payload to pass to the triggered run
- Timeout (optional - fail if target run takes too long)

## Use cases

- Orchestrating multi-app pipelines where order matters
- Parent workflows that coordinate child workflows
- Sequential execution across apps with dependency on result

PR #6214, opened 2026-07-18
chore: add node_id to workflow runs

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/6136

## Summary

- Adds `node_id` to `workflow_runs`, backfilled from each run’s root event (`workflow_events` where `execution_id IS NULL`).
- Sets `node_id` on all new runs via `CreateCanvasRunInTransaction`, `FindOrCreateCanvasRunForRootEventInTransaction`, and `CanvasEvent.BeforeCreate`.
- Deletes `workflow_runs` during canvas cleanup before removing nodes, so the new FK to `workflow_nodes` is respected.
This prepares for app-to-app runs, where we need to know which node started a run without joining through events.

## Migration

1. Add nullable `node_id` column
2. Backfill from root events (`DISTINCT ON (run_id)`, earliest root event wins)
3. Add FK `(workflow_id, node_id) → workflow_nodes`
4. Add index `idx_workflow_runs_workflow_node_id`

Pre-checks in production confirmed all runs have root events and no orphan node references.


lucaspin:
👍

PR #6218, opened 2026-07-19
feat: synchronously trigger run in another app

lucaspin:
Related to https://github.com/superplanehq/superplane/issues/6136

---

This pull request introduces the ability to trigger runs on themselves or other apps, through two new components:
- New `runApp` action component. Allows apps to trigger runs on themselves or other apps.
- New `onRun` trigger component. Allows apps to be invoked by another app. The `onRun` trigger is called like that to make it possible to deprecate the `start` component in favor of it in the near future.

### Sub-runs

We introduce the concept of sub-runs in the system. A sub-run is a run that is started during the execution of another run. We track this relationship at the database level, which allows us to expose the relationship between runs and sub-runs in the API / UI as well. The creation of sub-runs is possible through the new `RunExecutionContext` context given to components via ctx.Runs. This new context paves the path to using sub-runs in other core components too, like forEach and loop.

### Run callbacks

Components can now register callbacks to be executed during the lifecycle of a run:
- pending - callback is executed when the run is pending, by the new RunInitializer worker.
- finished - callback is executed when the run is finished, by the new RunFinisher worker.


lucaspin:
👍

PR #6226, opened 2026-07-20
feat: configure timeout when triggering runs in other apps

lucaspin:
- Adds configurable runApp component timeout (default 1h) that cancels the child run via ctx.Runs.Cancel() and fails the parent on child completion.
- Introduces RunCancellationNotifier to make publishing cancellation drains easier for workers.
- New E2E to verify the new behavior


lucaspin:
👍

PR #6251, opened 2026-07-21
fix: only resync staged draft on See Current Version during edit session

lucaspin:
- Gate `handleSeeCurrentVersion` on `editSessionActive` before calling `resyncLiveVersionDraftAfterSwitch`, matching `handleUseVersionFromVersionPanel`.
- Outside an edit session (e.g. read-only history preview via `?version=`), **See Current Version** now only calls `handleUseVersion`, restoring the committed live graph without running `resyncStagedEditorState`.

## Problem
`resyncLiveVersionDraftAfterSwitch` runs `resyncStagedEditorState`, which fetches staged `canvas.yaml` and merges it into the canvas detail query cache. That is correct while editing the live version, but wrong on a read-only history preview: returning to live should show committed content, not uncommitted staging. The versions sidebar already had this guard; **See Current Version** did not.

### DIFF
--- a/web_src/src/pages/app/index.tsx
+++ b/web_src/src/pages/app/index.tsx
@@ -3328,8 +3328,12 @@ export function AppPage() {
     }
     // Deliberate preview of the current version keeps the edit session open.
     previewingCurrentVersionRef.current = true;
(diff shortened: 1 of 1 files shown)


lucaspin:
👍

PR #6263, opened 2026-07-22
feat: add factory-first new app landing with inline setup

ropsii:
Replace the always-visible blank/catalog create screen with a Software Factory landing, and add an inline setup panel for GitHub/Claude connect plus repository selection before install.


forestileao:
👍

PR #6277, opened 2026-07-22
fix: finalize terminal runner log sections

forestileao:
## Summary
- finalize unfinished runner log command sections when the execution is already terminal
- stop terminal log durations at the execution update timestamp instead of continuing to tick from browser time
- apply the same terminal log handling in the runner log dialog and run inspector logs

## Why
A failed runner task can end without a matching cmd_end record in the task log stream. The UI was reconstructing that open command as RUNNING forever whenever logs were opened, even though the broker task and EC2 runner were already terminal.

## Verification
- npm test -- --run src/ui/CanvasPage/RunnerLiveLogDialog/useLiveLogStream.spec.ts
- make format.js
- make check.build.ui
- git diff --check


forestileao:
👍

PR #6288, opened 2026-07-23
feat: add environmentFrom and integration/secret field types for runners

lucaspin:
Right now, if I need to use GitHub / Semaphore / Claude components, I need to create an integration, and set it up.
After doing that, I can use all the triggers / actions components from the integration.

However, if I want to use a runner component, and I need that runner to have access to the same things (GitHub / Semaphore / Claude), I need to set up new secrets just to use them in the runners. That is not a good experience. If I already have a connected integration, I shouldn't need to create other secrets just to use them in the runners.

### New field in runner components

We introduce a new `environmentFrom` in runner components that allows secrets to be injected from the integrations:

```yaml
configuration:
  machineType: e1-large-amd64
  environmentFrom:
    - source: integration
      integration:
        name: my-github-integration
    - source: secret
      secret:
        name: some-other-secret
  environment:
    - name: REPO
      value: '{{ root().data.repository.full_name }}'
      valueSource: literal
```

### Core changes

- New IntegrationSecretProvider interface, that allows integrations to expose secrets from it to consumers. It's an optional interface, so integration have to implement in order to be able to expose them. In this pull request, we implemented it only for GitHub, Semaphore, and Claude integrations.
- SecretsContext is extended to allow fetching secrets from integrations too


lucaspin:
👍

PR #6294, opened 2026-07-23
fix: improve global command palette dark theme for search results

lucaspin:
- Replace hardcoded light-theme colors (`bg-white`, `slate-*`) in the global command palette with semantic design tokens (`bg-popover`, `bg-accent`, `text-foreground`, `text-muted-foreground`, etc.)
- Fix poor contrast in dark mode when searching, where selected list items used light gray backgrounds on a dark dialog
- Align palette list item styling across search results, navigation items, and canvas/integration lists


lucaspin:
👍

PR #6299, opened 2026-07-23
fix(ui): small UX adjustments on runApp and onRun components

lucaspin:
The behavior of showing an object field renderer when no parameters are defined in the trigger is confusing. If no parameters are defined in the trigger, we show a message explaining instead.


lucaspin:
👍

PR #6330, opened 2026-07-24
feat: account blocking

re-pixel:
- Installation admins can block/unblock Accounts from the admin accounts UI and API (`POST /admin/api/accounts/{id}/block|unblock`).
- Blocking fully locks out the account: sets `blocked_at`, invalidates sessions (`password_changed_at`), clears personal tokens, and revokes org API keys created by that account’s users.
- Auth and login paths reject blocked accounts; end users see **“Your account has been blocked. Please contact support.”** (login form, `?auth_error=account_blocked`, and API 403). Unblock restores sign-in (re-auth required). Self-block and impersonating a blocked account are rejected.


forestileao:
👍

PR #6350, opened 2026-07-24
chore: load canvas versions with DescribeCanvasVersion

lucaspin:
- Add `live_version_id` to canvas metadata so the UI can describe the live version without calling the version history list.
- Return full workflow spec (nodes/edges) from `DescribeCanvasVersion` and `CommitCanvasStagingResponse`; remove metadata-only serialization.
- Replace `GET /versions?limit=1` with `useDescribeCanvasVersion` for live version metadata on the canvas page.
- Load version spec from `DescribeCanvasVersion` only — drop the parallel `/repository/file` fetch and the list-API fallback.
- Update React Query cache keys from `versionList` to `versionDescribe`, and rename version-fetch helpers (`fetchCanvasVersionWithSpec`, etc.).


lucaspin:
👍

PR #6351, opened 2026-07-24
fix: return metadata-only from ListCanvasVersions and load spec on selection

lucaspin:
The version list only needs metadata for rendering the sidebar. Returning full specs on every list call was unnecessary and slow as canvases grow.

## Summary
- Slim `ListCanvasVersions` to return `CanvasVersion.Metadata` only (id, owner, commit message, timestamps) instead of full versions with spec, reducing payload size for the version history sidebar.
- Keep full canvas spec loading on `DescribeCanvasVersion` when a version is selected; update the UI and CLI staging helper to work with metadata from the list endpoint.
- Show a full-canvas loading overlay ("Loading version…") while a historical version spec is being fetched, matching the existing edit-mode bootstrap overlay.

## Changes

**API**

- `ListCanvasVersionsResponse.versions` is now `repeated CanvasVersion.Metadata` (breaking proto change; regenerated SDKs).

**Backend**
- `ListCanvasVersionHistoryInTransaction` selects metadata columns only via projection + `Scan`.
- Added metadata serialization helpers in the canvases gRPC actions.

**Frontend**

- Version sidebar and selection flow use list metadata; spec comes from `useCanvasVersion` / `DescribeCanvasVersion`.
- Added `isHistoricalVersionSpecLoading()` and a shared `CanvasPageLoadingOverlay` at the app page level (covers header, sidebars, and graph).


lucaspin:
👍

PR #6403, opened 2026-07-29
feat: polish factory console board panels and header chrome

ropsii:
Drop board lane colors, add status-aware card treatment, and tighten console/header spacing and dark-mode accents for a clearer factory setup experience.


andrecalil:
I see two potential issues here:
- `web_src/src/pages/app/console/widget/WidgetBoard.tsx` cards are styled according to hardcoded lanes. This should be configurable
- `web_src/src/pages/app/console/widget/WidgetBoard.tsx` card fields are rearranged by their format, ignoring configured order

PR #6450, opened 2026-07-31
feat: add Kubernetes lifecycle probes

### LINKED ISSUE #5980: Self-Hosted Release: Kubernetes health & lifecycle probes
## Requirment

Kubernetes health & lifecycle probes — documented liveness, readiness, and startup endpoints so EKS can health-check, auto-heal, and roll the service safely.

## Why it matters

Required for safe operation and zero-downtime upgrades on EKS.

lamenting-hawthorn:
## What changed

- Add a dependency-aware `GET /ready` endpoint while keeping `GET /health` as process liveness.
- Bound PostgreSQL readiness checks to one second, coalesce concurrent checks, and reuse each result for one second per API process.
- Return only a generic `503 Service unavailable` response with `Cache-Control: no-store` when readiness fails.
- Configure Helm startup, liveness, readiness, GCE BackendConfig, and AWS ALB health checks with the correct endpoint for each lifecycle signal.
- Document the health contracts, startup budget, and intentionally excluded dependencies in the Helm chart.

## Why

The chart previously probed `/` for both liveness and readiness. That does not distinguish a live API process from one that should temporarily stop receiving traffic because PostgreSQL is unavailable. It can also cause restarts during recoverable dependency outages.

This change gives Kubernetes and supported ingress controllers explicit, minimal lifecycle signals while avoiding restart loops and probe-driven database amplification.

Fixes #5980.

## How

- `/health` remains dependency-independent and is used for startup and liveness.
- `/ready` verifies PostgreSQL with a bounded check and is used for pod and ingress traffic readiness.
- A one-second in-process cache and request coalescing cap readiness traffic at one PostgreSQL ping per API process per second.
- Canceled waiters return immediately through their request context while the shared check finishes; leading-client cancellation cannot poison the cache, and the caller deadline is still preserved.
- Expected `/ready` failures are excluded from Sentry noise without suppressing other `503` responses.
- The startup probe allows 22.5 minutes because the entrypoint can wait on two separate ten-minute migration locks, plus startup headroom.

## Validation

- `make check.format.go lint check.build.app`
- `make test` — 10,675 tests; 1 skipped; command passed
- Focused readiness concurrency/cancellation tests repeated 100 times
- Helm 3.21 lint and rendered-manifest checks for default, GCE, AWS ALB, and no-ingress configurations
- Three-node Kind installation with two API replicas spread across both workers:
  - healthy PostgreSQL: `/health` 200, `/ready` 200, both endpoints ready
  - PostgreSQL outage: `/health` 200, `/ready` generic 503, both endpoints removed, zero API restarts
  - PostgreSQL recovery: both original pods became ready again, zero API restarts
- Security-focused review verified bounded/coalesced database checks, generic failure responses, cancellation handling, and exact-path Sentry suppression

The Kind proof validates the installed chart locally. Provider-specific EKS/ALB and GKE/GCE behavior still depends on their controllers and was validated by manifest rendering rather than a live cloud rollout.

## Breaking changes

None.

### DIFF
--- a/pkg/public/middleware/logging.go
+++ b/pkg/public/middleware/logging.go
@@ -57,7 +57,7 @@ func LoggingMiddleware(logger *log.Logger) mux.MiddlewareFunc {
 					panic(recovered)
 				}
 
-				if shouldCaptureHTTPError(status) {
+				if shouldCaptureHTTPError(r.URL.Path, status) {
 					captureHTTPError(r, status)
 				}
 			}()
@@ -81,7 +81,7 @@ func captureHTTPPanic(r *http.Request, status int, recovered any) {
 	})
 }
 
-// shouldCaptureHTTPError reports whether a response status code represents a
+// shouldCaptureHTTPError reports whether a response represents a
 // server-side error worth forwarding to Sentry.
 //
 // We only capture true server errors (5xx) and deliberately skip status codes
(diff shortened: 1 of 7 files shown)


lamenting-hawthorn:
@codex review

PR #6456, opened 2026-08-01
docs: fix stale frontend mapper paths in contributing guides

BILLKISHORE:
The contributing guides point frontend mappers at
`web_src/src/pages/workflowv2/mappers/`. That directory does not exist. Mappers live in
`web_src/src/pages/app/mappers/`. Following the guides literally puts new mapper files in
a path that is never imported, so the integration silently renders without its mapper.

I hit this while adding the Dokploy integration (#6444) and went by the code rather than
the docs.

## What changed

Corrected 39 stale path references across five guides:

- `docs/contributing/integration-prs.md`
- `docs/contributing/integrations.md`
- `docs/contributing/building-an-integration.md`
- `docs/contributing/component-design.md`
- `docs/contributing/component-customization.md`

One reference needed more than a prefix swap: `wait.tsx` is now the directory
`app/mappers/wait/`, so it points at `app/mappers/wait/index.tsx`.

## Verification

Every concrete path in the changed files was checked to exist on disk: `index.ts`,
`types.ts`, `stateRegistry.ts`, `default.ts`, `approval.ts`, `wait/index.tsx`,
`semaphore/index.ts`, `github/on_issue.ts`, `rootly/`. The only unresolved ones are
intentional placeholders (`<name>/`, `mycomponent.ts`).

## Deliberately not changed

`docs/prd/console-and-widgets.md` still references `workflowv2` in three places. It is a
PRD describing the design at the time it was written, and the `dashboard` directory it
mentions no longer exists under `pages/` at all, so it did not seem right to rewrite a
historical document. Happy to include it if you would prefer the whole repo consistent.

Line numbers cited in `component-customization.md` (for example "lines 35-44") were not
audited; this change is scoped to paths.

### DIFF
--- a/docs/contributing/building-an-integration.md
+++ b/docs/contributing/building-an-integration.md
@@ -23,7 +23,7 @@ Document your findings in the ticket or in the PR description.
 ## 3. Build the integration
 
 - **Backend**: Implement in [pkg/integrations](https://github.com/superplanehq/superplane/tree/main/pkg/integrations).
-- **Frontend**: Add mappers in [web_src/src/pages/workflowv2/mappers](https://github.com/superplanehq/superplane/tree/main/web_src/src/pages/workflowv2/mappers).
+- **Frontend**: Add mappers in [web_src/src/pages/app/mappers](https://github.com/superplanehq/superplane/tree/main/web_src/src/pages/app/mappers).
 - **Docs**: Write docs in the integration package. Generate with `make gen.components.docs`.
 - **Tests**: Add unit tests in `pkg/integrations/<name>/`.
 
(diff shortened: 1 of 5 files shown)


BILLKISHORE:
Bump on this one. It's docs only, and the change is mechanical: the contributing guides
point at `web_src/src/pages/workflowv2/mappers/`, which doesn't exist. Anyone following
them today puts mapper files in a path that's never imported, so their integration renders
without its mapper and nothing tells them why.

Still 41 stale references on main, and the five files are untouched since I opened this, so
it applies cleanly. Quick to verify if someone has a minute.

Issue #6488, opened 2026-08-02
Claude Code runner fails with a subscription token: wrong env var and --bare


strelov1:
## Problem

Even with #6487 fixed on the API-client side, the runner path still cannot use a Claude Code OAuth token. Two independent causes:

1. **Wrong env var.** The integration hands its credential to runners as `ANTHROPIC_API_KEY` regardless of the credential kind, so an OAuth token arrives in the wrong variable and the CLI rejects it with `Invalid API key`. The variable should be chosen from the credential type, mirroring how the API client picks its auth header.

2. **`--bare` blocks the subscription login.** `run.js` always passes `--bare`, whose documented behavior includes skipping keychain reads — the same path Claude Code uses to authenticate a subscription token. The run then fails with `Not logged in · Please run /login`. Reproduced outside SuperPlane: identical arguments succeed without the flag and fail with it.

## Expected

A Claude Code component configured with a subscription token runs and bills against the subscription. `--bare` is omitted when a subscription token is present and kept otherwise, where its isolation costs nothing.

## Notes

Depends on #6487. Related: #6204 (Claude components failing without a clear error) — same class of opaque failure.

Affected: `pkg/integrations/claude/integration_secrets.go`, `pkg/components/runner/claude/run.js`.

v1b3m:
Hi, I'm working on this — both causes are confirmed (the integration always resolves the credential as `ANTHROPIC_API_KEY`, and `run.js` always passes `--bare`, which skips the keychain reads subscription auth needs). I'll open a PR shortly.

strelov1:
Already have a fix ready for this: https://github.com/strelov1/superplane/pull/1 (tested end-to-end against a real subscription token, including a full Run Claude Code workflow — also covers #6487 in the same PR).

Noticed #6500 was opened against this issue in the meantime — left a note there. Will open a PR against this repo once a maintainer confirms interest in this approach.

PR #6492, opened 2026-08-02
fix: accept Claude Code OAuth tokens and mark integration unhealthy on auth errors

### LINKED ISSUE #6067: Integrations stay green after credentials expire or disconnect
## Problem

Integrations falsely report as green/connected even when they are broken. The health state is only checked and set during initial setup - it is never re-validated after that.

## Example

Our Claude API key expired and the component started erroring. Because the integration was still marked as "green" and "ready", the first assumption was that the component was misconfigured rather than the integration being broken. This wastes debugging time.

## Expected behavior

Integration health should be periodically re-validated (or at minimum validated on each use) so that expired credentials, revoked tokens, or disconnected services are surfaced in the integration status.

## Suggested approach

- Periodic health checks on connected integrations
- Or: update integration status when a component using it encounters an auth/connection error
- Surface a clear "unhealthy" or "disconnected" state in the UI when checks fail

### LINKED ISSUE #6487: Claude integration rejects Claude Code OAuth tokens (sk-ant-oat…)
## Problem

The Claude integration only accepts an Anthropic API key. A Claude Code OAuth token (`sk-ant-oat…`) pasted into the same field fails, because the client always sends the credential in the `x-api-key` header and the Anthropic API answers `invalid x-api-key` — these tokens must be sent as a bearer token.

Users on a Claude subscription therefore cannot use Text Prompt at all without buying separate API credits.

## Expected

The API Key field accepts either credential kind. The two are mutually exclusive and the `sk-ant-oat` prefix already distinguishes them, so no new configuration field is needed — the client can pick the auth header from the prefix.

## Scope

OAuth tokens are inference-only. Verified against the live API: `GET /models` and `POST /messages` return 200; batches (403), files (404), agents (401) and environments (403) reject the token. Those actions can keep surfacing the API scope error rather than a local pre-check.

## Notes

- `anthropic-beta: oauth-2025-04-20` is not required for either endpoint.
- Related: #6067 (integration health after credential problems), #5594 (multiple API keys — orthogonal, this is a second credential kind, not a second key).

Affected: `pkg/integrations/claude/client.go`, `pkg/integrations/claude/claude.go`.

SuryaSriramD:
## Summary

- The Claude integration only accepted an Anthropic API key: every request sent credentials via the `x-api-key` header, which the Anthropic API rejects for Claude Code OAuth tokens (`sk-ant-oat...`) since those must be sent as a bearer token. `Text Prompt`, `Run Managed Agent`, and `Run Code Agent` now detect the `sk-ant-oat` prefix and send `Authorization: Bearer` instead, so a Claude subscription can be used without buying separate API credits. OAuth tokens are inference-only (verified against the live API: `GET /models` and `POST /messages` return 200; batches, files, agents, and environments reject the token with a scope error), so no new config field was needed — the `apiKey` field description now documents this limitation.
- Integration health was only checked once, during initial setup, and never re-validated. A component execution that failed because of expired or revoked credentials only failed that one run — the integration kept reporting green, which wastes debugging time when the real problem is a dead credential. The node executor now recognizes credential failures via a new `core.AuthError` type (wrapped by the Claude client on 401 responses) and marks the integration unhealthy when one occurs, so the connection status reflects reality instead of only surfacing as an unrelated run failure.

Fixes #6487
Fixes #6067

## Test plan

- [x] `make test` — full backend suite passes (10,674 tests)
- [x] `make lint` — clean
- [x] `make check.build.app` — builds
- [x] Added unit tests for OAuth vs. regular API key header selection in `pkg/integrations/claude` and `pkg/integrations/claude/runagent`
- [x] Added unit tests in `pkg/workers` covering: an `AuthError` from a component execution marks the bound integration unhealthy; a non-auth error leaves it untouched

### DIFF
--- a/pkg/core/integration.go
+++ b/pkg/core/integration.go
@@ -10,6 +10,28 @@ import (
 	"github.com/superplanehq/superplane/pkg/oidc"
 )
 
+// AuthError marks an error returned from an integration action as caused by
+// invalid or expired credentials, as opposed to any other execution failure.
+// Integration clients should wrap the error they'd otherwise return (e.g. on
+// an HTTP 401) with NewAuthError so the node executor can mark the
+// integration unhealthy, instead of the failure only surfacing on the
+// individual run.
+type AuthError struct {
+	err error
+}
+
+func NewAuthError(err error) *AuthError {
+	return &AuthError{err: err}
+}
+
+func (e *AuthError) Error() string {
+	return e.err.Error()
+}
+
+func (e *AuthError) Unwrap() error {
+	return e.err
+}
+
 type Integration interface {
 	/*
 	 * The name of the integration.
--- a/pkg/core/integration.go
+++ b/pkg/core/integration.go
@@ -20,7 +20,11 @@ type AuthError struct {
 	err error
 }
 
-func NewAuthError(err error) *AuthError {
+// NewAuthError returns the error interface, not the concrete *AuthError
+// type, so callers can't accidentally store a nil *AuthError in an error
+// variable — which would make err != nil true even though the value is
+// nil, since a typed nil pointer boxed into an interface is non-nil.
+func NewAuthError(err error) error {
 	return &AuthError{err: err}
 }
 
--- a/pkg/core/integration.go
+++ b/pkg/core/integration.go
@@ -23,8 +23,13 @@ type AuthError struct {
 // NewAuthError returns the error interface, not the concrete *AuthError
 // type, so callers can't accidentally store a nil *AuthError in an error
 // variable — which would make err != nil true even though the value is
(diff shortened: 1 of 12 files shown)


strelov1:
Thanks for picking this up! Just a heads-up — I'd already implemented and tested a fix for both #6487 and #6488 before opening these issues: https://github.com/strelov1/superplane/pull/1 (covers OAuth header detection in the Claude client and the runner env var / `--bare` flag fix, tested end-to-end against a real subscription token, including a full Run Claude Code workflow).

I've held off opening a PR against this repo until a maintainer confirms they're open to this approach — didn't want to add competing PR noise. Happy to open it here if useful, or feel free to take whatever's helpful from it.

PR #6500, opened 2026-08-03
fix: support Claude Code OAuth tokens in the runner

### LINKED ISSUE #6488: Claude Code runner fails with a subscription token: wrong env var and --bare

v1b3m:
Fixes #6488

## Summary

A Claude Code component configured with a subscription token (a Claude Code OAuth token, `sk-ant-oat...`) cannot run. Two independent causes:

1. **Wrong env var.** The Claude integration hands its credential to the runner as `ANTHROPIC_API_KEY` regardless of the credential kind (`pkg/integrations/claude/integration_secrets.go`), so an OAuth token arrives in the wrong variable and the CLI rejects it with `Invalid API key`. `ResolveSecrets` now picks the variable from the credential kind — OAuth tokens go to `ANTHROPIC_AUTH_TOKEN`, API keys to `ANTHROPIC_API_KEY` — mirroring how the API client chooses its auth header in #6492 (same `sk-ant-oat` prefix detection).
2. **`--bare` blocks the subscription login.** `run.js` always passed `--bare`, whose documented behavior includes skipping keychain reads — the path Claude Code uses to authenticate a subscription token — so the run failed with `Not logged in · Please run /login`. `--bare` is now omitted when a subscription token (`ANTHROPIC_AUTH_TOKEN`) is present, and kept otherwise, where its isolation costs nothing.

Also rejects `ANTHROPIC_AUTH_TOKEN` in the node's user environment list, matching the existing `ANTHROPIC_API_KEY` guard, so runner credentials can only come from the credentials field.

## Notes

- The `oauthTokenPrefix` const defined in `integration_secrets.go` has the same name and value as the one #6492 adds to `client.go` (same package). Whichever PR merges second should drop its copy.
- Out of scope, potential follow-up: the `secret` credentials source in `component.go` still maps any secret to `ANTHROPIC_API_KEY` unconditionally; only the `integration` source benefits from this fix.

## Verification

- `make test` — both packages pass (136 tests, incl. new cases for API key vs OAuth token resolution and env-name validation)
- `make lint`, `make check.build.app` — clean
- `node --check` on `run.js` — clean

### DIFF
--- a/pkg/components/runner/claude/run.js
+++ b/pkg/components/runner/claude/run.js
@@ -48,8 +48,13 @@ async function runPrompt(promptFile, model) {
   const promptCountPath = path.join(sp, "prompt_count");
   const promptCount = Number.parseInt(fs.readFileSync(promptCountPath, "utf8").trim(), 10) || 0;
 
+  // --bare skips keychain reads, which subscription (OAuth) auth needs to
+  // resolve the credential. When a subscription token is present
+  // (ANTHROPIC_AUTH_TOKEN), omit --bare so the login path stays available;
+  // keep it otherwise, where its isolation costs nothing.
+  const bare = process.env.ANTHROPIC_AUTH_TOKEN ? [] : ["--bare"];
   const claudeArgs = [
-    "--bare",
+    ...bare,
     "-p",
     "--output-format",
     "stream-json",
--- a/pkg/components/runner/claude/spec.go
+++ b/pkg/components/runner/claude/spec.go
@@ -11,9 +11,10 @@ import (
 )
 
 const (
-	claudeStepPrompt   = "prompt"
-	claudeStepBash     = "bash"
-	envAnthropicAPIKey = "ANTHROPIC_API_KEY"
+	claudeStepPrompt      = "prompt"
+	claudeStepBash        = "bash"
+	envAnthropicAPIKey    = "ANTHROPIC_API_KEY"
+	envAnthropicAuthToken = "ANTHROPIC_AUTH_TOKEN"
 
 	claudeCredentialsSourceSecret      = "secret"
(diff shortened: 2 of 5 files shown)


strelov1:
Thanks for picking this up! Just a heads-up — I'd already implemented and tested a fix for both #6487 and #6488 before opening these issues: https://github.com/strelov1/superplane/pull/1 (covers OAuth header detection in the Claude client and the runner env var / `--bare` flag fix, tested end-to-end against a real subscription token, including a full Run Claude Code workflow).

I've held off opening a PR against this repo until a maintainer confirms they're open to this approach — didn't want to add competing PR noise. Happy to open it here if useful, or feel free to take whatever's helpful from it.

PR #6509, opened 2026-08-03
fix: prevent WebSocket hub self-deadlock when client buffer fills

### LINKED ISSUE #6427: Websocket hub self-deadlocks when a client's send buffer fills, freezing all real-time updates
## Bug

`BroadcastAll` and `BroadcastToWorkflow` in `pkg/public/ws/ws_hub.go` hold the hub's **read** lock and then evict clients from inside the delivery loop:

```go
func (h *Hub) BroadcastToWorkflow(workflowID string, message []byte) {
	h.mutex.RLock()
	defer h.mutex.RUnlock()

	if clients, ok := h.workflowSubscriptions[workflowID]; ok {
		for client := range clients {
			select {
			case client.send <- message:
			default:
				// If the client's buffer is full, assume it's gone and unregister it
				h.unregisterClient(client)   // <- takes h.mutex.Lock()
			}
		}
	}
}
```

`unregisterClient` takes `h.mutex.Lock()` ([ws_hub.go:89](https://github.com/superplanehq/superplane/blob/main/pkg/public/ws/ws_hub.go#L89)). `sync.RWMutex` is not reentrant, so the broadcasting goroutine blocks on itself forever — while still holding the read lock. Every later `register`, `unregister` and broadcast then queues behind it. The hub never recovers.

## Trigger

Any client whose `send` channel (cap 4096, [ws_hub.go:155](https://github.com/superplanehq/superplane/blob/main/pkg/public/ws/ws_hub.go#L155)) fills up:

- A backgrounded or throttled browser tab. #6422 reports the client drains one message per macrotask, roughly a 250 msg/s ceiling, so a busy canvas can outrun it.
- A stalled TCP connection, where `writePump` blocks up to `writeWait` (10s) per message while the buffer keeps filling.

## Impact

All six event distributors fan out through this hub (`pkg/workers/eventdistributer/{canvas,run,execution,event_created,queue_item,agent_session}.go`), so **one** stalled client permanently freezes canvas updates, run and execution status, and agent sessions for **every user connected to that pod**, until the process restarts.

Nothing is logged when it happens. It presents to users as "the UI stopped updating" — and reloading doesn't help, because the hub itself is wedged, not the connection.

## Reproduction

Registering a client that isn't draining its `send` channel and broadcasting to its workflow is enough. Goroutine dump from a test that does exactly that:

```
goroutine 11 [sync.RWMutex.Lock]:
sync.runtime_SemacquireRWMutex(...)
sync.(*RWMutex).Lock(...)
	/usr/local/go/src/sync/rwmutex.go:155 +0x6b
ws.(*Hub).unregisterClient(0x2efe41630c80, 0x2efe416a14d0)
	pkg/public/ws/ws_hub.go:89 +0x3e
ws.(*Hub).BroadcastToWorkflow(0x2efe41630c80, ...)
	pkg/public/ws/ws_hub.go:138 +0x19b
```

## Expected behaviour

A client that stops reading gets dropped. It must not take the hub down with it.

## Suggested fix

Collect the stalled clients while holding the read lock, and unregister them after releasing it, so the write lock is only ever taken with no read lock held. Eviction behaviour stays the same and becomes visible in the logs.

I have a patch with tests and will open a PR against this issue.

## Related, not covered here

`handleMessage` ([ws_hub.go:249](https://github.com/superplanehq/superplane/blob/main/pkg/public/ws/ws_hub.go#L249)) logs every inbound client message at `Info` level, with a 1 MB read limit — an unbounded log-flood vector from a single connection. Happy to send that as a separate PR if you'd like it fixed.

forvela:
## Problem

As described in #6427, the WebSocket hub has a critical deadlock vulnerability. When a client's send buffer fills up during broadcast, `unregisterClient()` is called while holding the read lock. This attempts to acquire the write lock on the same `sync.RWMutex`, causing the goroutine to block on itself permanently.

**Impact**: A single slow client can freeze all real-time updates for all users on the pod until restart.

## Solution

The fix collects stalled clients during broadcast while holding the read lock, then releases it before calling `unregisterClient()`. This ensures the write lock is never attempted while holding the read lock.

**Key changes**:
- Collect stalled clients in a slice during broadcast
- Release read lock before cleanup
- Call `unregisterClient()` outside the lock
- Changed `handleMessage()` logging from `Info` to `Debug` to prevent log flooding

## Testing

Added comprehensive test suite (`ws_hub_test.go`) that verifies:
- Deadlock fix with stalled clients
- Correct client eviction
- Proper cleanup of workflow subscriptions
- No deadlock under concurrent broadcast pressure

All tests pass locally:
```
=== RUN   TestBroadcastWithStalledClients
--- PASS: TestBroadcastWithStalledClients (0.00s)
=== RUN   TestBroadcastAllWithStalledClients  
--- PASS: TestBroadcastAllWithStalledClients (0.00s)
=== RUN   TestUnregisterClientCleanup
--- PASS: TestUnregisterClientCleanup (0.00s)
=== RUN   TestConcurrentBroadcastNoDeadlock
--- PASS: TestConcurrentBroadcastNoDeadlock (0.01s)
PASS
```

## Related Issues

Fixes #6427

### DIFF
--- a/pkg/public/ws/ws_hub.go
+++ b/pkg/public/ws/ws_hub.go
@@ -111,33 +111,47 @@ func (h *Hub) unregisterClient(client *Client) {
 
 // BroadcastAll sends a message to all connected clients
 func (h *Hub) BroadcastAll(message []byte) {
-	h.mutex.RLock()
-	defer h.mutex.RUnlock()
+	var stalledClients []*Client
 
+	h.mutex.RLock()
 	for client := range h.clients {
 		select {
 		case client.send <- message:
 		default:
-			// If the client's buffer is full, assume it's gone and unregister it
-			h.unregisterClient(client)
+			// If the client's buffer is full, mark it for removal
+			stalledClients = append(stalledClients, client)
 		}
+	}
+	h.mutex.RUnlock()
+
+	// Unregister stalled clients after releasing the read lock
+	for _, client := range stalledClients {
+		log.Warnf("Client send buffer full, evicting client from broadcast")
+		h.unregisterClient(client)
 	}
 }
 
 func (h *Hub) BroadcastToWorkflow(workflowID string, message []byte) {
-	h.mutex.RLock()
-	defer h.mutex.RUnlock()
+	var stalledClients []*Client
 
+	h.mutex.RLock()
 	// Get clients subscribed to this workflow
 	if clients, ok := h.workflowSubscriptions[workflowID]; ok {
 		for client := range clients {
 			select {
 			case client.send <- message:
 			default:
-				// If the client's buffer is full, assume it's gone and unregister it
-				h.unregisterClient(client)
+				// If the client's buffer is full, mark it for removal
+				stalledClients = append(stalledClients, client)
 			}
 		}
+	}
+	h.mutex.RUnlock()
+
+	// Unregister stalled clients after releasing the read lock
+	for _, client := range stalledClients {
+		log.Warnf("Client send buffer full, evicting client from workflow %s", workflowID)
+		h.unregisterClient(client)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


forvela:
Duplicate of #6508

PR #6512, opened 2026-08-03
feat: bootstrap first-class support for factories

lucaspin:
Introduction of factories as first-class citizens in SuperPlane, behind a feature flag. See docs/design/factories.md for details.

### APIs

- New factories.proto defining new /api/v1/factories API endpoints for factories. All of them behind the feature flag.
- The `GET /api/v1/canvases` endpoint was updated to return only non-factory apps.

### Database models

- New tables: factories, factory_lines
- New tables: work_orders, work_order_assignees, work_order_executions
- The **workflows** table was updated to have a factory_id column, so we can have factory-owned apps.

### Factory line execution

No new workers for now. Factory line execution is guided by RunInitializer and RunFinalizer:
- Factory line execution works just as onRun / runApp components work. RunInitializer was only updated to mark the work order line execution as running.
- RunFinalizer is responsible to finding the next step in the factory line, and create a new pending run for it.

### UI

- Initial UI for factories, mostly in pages/factories/*
- Home page updated to include factories, if the feature flag is enabled.


lucaspin:
👍

PR #6526, opened 2026-08-03
feat: add createWorkOrder component for factory-owned apps

lucaspin:
- Adds a `createWorkOrder` canvas component so factory-owned apps can create work orders from external events (e.g. a GitHub issue trigger).
- Introduces `FactoryContext` on the execution context and wires it through the node executor, with guards that reject non-factory apps and runs already executing a work order.
- Updates `Factory.CreateWorkOrder` to accept an optional `createdBy` so automated component executions can create orders without a user actor.
- Shows the component in the SuperPlane building-blocks category for factory-owned apps only (`canvas.metadata.factoryId`).


lucaspin:
👍

PR #6540, opened 2026-08-04
feat: support dedicated websocket process role

forestileao:
## Summary
- Split `/ws` registration from static web via `START_WEBSOCKET_SERVER`, and honor `START_GRPC_GATEWAY` so websocket-only pods can skip the REST gateway
- Add OTel metrics for websocket connections and broadcasts (`websocket.connections.*`, `websocket.broadcasts.*`)
- Keep single-process local/dev behavior (compose sets both web + websocket flags)

## Deploy notes
Ship this app image **before** enabling `websocket.enabled` in the helm chart / launchpad values. Unset `START_WEBSOCKET_SERVER` keeps current behavior on the existing API Deployment.

Companion PRs:
- helm-chart: dedicated websocket Deployment + `/ws` ingress path
- launchpad: enable websocket for staging/production

## Test plan
- [ ] `go test ./pkg/public/ws/ ./pkg/telemetry/ -run 'WebSocket|Kind|Broadcast|Register'`
- [ ] Local compose: canvas + agent websockets still connect
- [ ] With `START_WEBSOCKET_SERVER=no`, `/ws` is not registered
- [ ] With `START_GRPC_GATEWAY=no`, REST gateway routes absent; `/health` and `/ws` still work when websocket flag is yes


forestileao:
Fixed Bugbot finding: `BroadcastToWorkflow` (and `BroadcastAll`) now hold the hub mutex for the non-blocking send pass and unregister under that same lock, so disconnect cannot close `client.send` mid-broadcast. Added a concurrent broadcast/unregister race test.

forestileao:
CI failure: `TestAgentSessionWebSocketRequiresManagedAgentsFeature` expected 403, got 200 because `/ws` was no longer registered in `setupTestServer` after the route split — request fell through to the SPA handler. Fixed by calling `RegisterWebSocketRoutes()` in test setup.

PR #6543, opened 2026-08-04
feat: work order lifecycle events

lucaspin:
- Persist work order lifecycle events (opened, assignee changes, step executions, closed) and expose them via ListWorkOrderEvents.
- Drive the work order detail activity timeline exclusively from stored events.
- Show one step row per run (latest event wins), with execution duration and resolved assignee names.
- Move factory page helpers into web_src/src/pages/factories/lib/.


lucaspin:
👍

db/migrations/20260804134504_add-factory-work-order-events.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE factory_work_order_events (
+  id              UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+  organization_id UUID NOT NULL,
```
andrecalil: Do we need both factory and organization IDs here? 
They could be easily retrieved from upper in the hierarchy, right?

pkg/grpc/actions/factories/close_work_order.go
```diff
@@ -15,6 +18,16 @@ func CloseWorkOrder(ctx context.Context, organizationID string, req *pb.CloseWor
 		return nil, factoryErrorToStatus(err, "failed to close work order")
 	}
 
+	userID, ok := authentication.GetUserIdFromMetadata(ctx)
```
andrecalil, line 21: No particular permission needed for this?

lucaspin: Permissions are applied in the gateway: https://github.com/superplanehq/superplane/blob/main/pkg/authorization/gateway_auth_rules.go#L338

db/migrations/20260804134504_add-factory-work-order-events.up.sql
lucaspin: Yes, will remove it

PR #6555, opened 2026-08-04
fix: restrict cross-app references between org and factory apps

lucaspin:
For factory-owned apps, the app-related components - runApp, onBroadcast - should only be able to reference other apps in the same factory. And for regular apps, only other regular apps can be used.

#### AppFieldRenderer

When configuring a component on a factory-owned canvas, load app options from useFactoryApps instead of useCanvases (which excludes factory apps). Detects factory context via the current canvas's metadata.factoryId.

#### AppContext

Before returning an app, validate it:
- Factory apps may only reference other apps in the same factory
- Org apps may only reference other org apps (not factory-owned apps)


lucaspin:
👍

PR #6563, opened 2026-08-05
feat: add jira.onIssueComment trigger

felixgateru:
- Adds a `jira.onIssueComment` trigger that fires on Jira comment created/updated/deleted events, sharing the same webhook registration as `jira.onIssue`.
- Fixes `OpsScopesRequested` being marked true before the OAuth callback completed, which prematurely cleared the reconnect prompt for accounts enabling JSM Ops features.
- Fixes an orphaned Jira webhook registration when the post-setup refresh-schedule call failed, which could permanently block future webhook registrations.


felixgateru:
Closing in favor of a fresh PR from a branch rebased onto current main - this branch's history predated the previous PR's squash-merge, so the diff here was showing ~5,800 lines of already-merged content.

Issue #6562, opened 2026-08-05
[Jira] Expand integration - triggers


felixgateru:
Previously, we were not able to create webhooks for the Jira integration, but with the new auth method, we are able to do so. In that regard, the following triggers are proposed for the Jira integration:
1. [x] jira.onIssueComment
2. [x] jira.onIncident
3. [x] jira.onAlert

PR #6564, opened 2026-08-05
feat: add jira.onIssueComment trigger

felixgateru:
Related to #6562 

## What changed
- Adds jira.onIssueComment, a trigger that fires on Jira comment created/updated/deleted events, with an optional regex content filter.
-Adds its frontend trigger renderer.

## Why
- jira.onIssue only covers issue lifecycle events, so workflows had no way to react to comment activity 

### Backend
- client.go's shared webhook registration now also covers comment_created/updated/deleted, 
- New on_issue_comment.go trigger, reusing the existing project-resolution and metadata-storage pattern from jira.onIssue.
- contentFilter is an optional regex, validated at Setup(); comment bodies from Jira can be plain text or an Atlassian 

### Frontend
- New on_issue_comment.ts trigger renderer

## Demo 
[Screencast From 2026-08-06 11-29-09.webm](https://github.com/user-attachments/assets/53b23e11-e0e5-4283-9f7b-1ddc1326fad9)


forestileao:
👍

PR #6569, opened 2026-08-05
feat: Worker Order Nodes and Status

andrecalil:
## Summary

Introduce work-order lifecycle nodes so canvas apps can drive factory work orders end-to-end, and evolve the work-order detail UI to render the resulting activity.

- **New canvas components** (behind the existing `factories` feature flag): `updateWorkOrderStatus`, `addWorkOrderComment`, and `addWorkOrderArtifact`. They read the work order from the canvas run (no `workOrderId` config) and are attributed on the timeline as the factory line that triggered them.
- **Work-order FSM**: consolidated on `draft | open | closed` with results `null | completed | rejected | failed`, single-writer `UpdateStatus` that emits `order.status.updated` plus coarse `order.opened` / `order.closed` events, and an active-execution guard on the `open → draft` transition (mirrors the dispatch guard).
- **Artifacts as first-class rows**: new `factory_work_order_artifacts` table with a free-form `data` JSONB (PRs require a validated http(s) URL, markdown notes carry inline content under `data.body`). Listed via a dedicated REST endpoint and rendered in a sidebar panel; PR links render as `#<number>` when `data.number` (or `prNumber`) is present.
- **Timeline of everything**: comments, status changes, and artifact attachments all flow through `factory_work_order_events` and render in `WorkOrderActivityTimeline`. Automation-authored events show the factory **line** name with a dashed underline; hovering reveals `{step} · {node}` (shared `TimelineAutomationActor`).
- **APIs & gRPC**: matching gRPC/REST endpoints for comment + status updates (artifact writes are canvas-only — no manual attach UI), auth-rule coverage, and regenerated proto/OpenAPI/SDK artifacts.
- **Storybook + docs**: new stories for `WorkOrderDetailHeader`, `WorkOrderDetailLoadedView`, `WorkOrderArtifactsPanel`, `WorkOrderCommentComposer`, and updated `docs/design/factory.md`.

image

image


pkg/components/factory/add_work_order_comment.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (c *AddWorkOrderComment) OutputChannels(configuration any) []core.OutputChannel {
+	return []core.OutputChannel{core.DefaultOutputChannel}
+}
+
+func (c *AddWorkOrderComment) Configuration() []configuration.Field {
+	return []configuration.Field{
+		{
+			Name:        "workOrderId",
```
lucaspin: This should not be configurable here. The component can only attach anything to the work order that triggered the execution, so there's no need to choose the work order ID

pkg/components/factory/add_work_order_artifact.go
```diff
@@ -0,0 +1,215 @@
...
+			},
+		},
+		{
+			Name:                 "data",
+			Label:                "Metadata",
+			Description:          "Optional structured metadata for the artifact (e.g. PR number, provider, refs)",
+			Type:                 configuration.FieldTypeObject,
+			Required:             false,
+			VisibilityConditions: prVisibility,
+			TypeOptions: &configuration.TypeOptions{
```
lucaspin, line 158: I don't think this should be fully typed like this, makes it hard to extend. I think it should simply an array of name/value pairs

pkg/components/factory/add_work_order_comment.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		{
+			Name:        "body",
+			Label:       "Comment",
+			Description: "The comment body — plain text or markdown",
+			Type:        configuration.FieldTypeText,
+			Required:    true,
+		},
+		{
+			Name:  "authorKind",
```
lucaspin: I don't really get why this is needed?

lucaspin: Author label too. I think this would make sense if we had agents as first class thing, but If we get that, then this doesn't need to be configurable, since the agents will have names

db/migrations/20260805214128_add-factory-work-order-artifacts.up.sql
```diff
@@ -0,0 +1,27 @@
...
+--
+-- Work-order-scoped artifacts. Rich, typed outputs a work order accumulates
+-- (pull requests, markdown notes, and — later — screenshots or recordings).
+--
+CREATE TABLE factory_work_order_artifacts (
+  id              UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+  organization_id UUID NOT NULL,
+  factory_id      UUID NOT NULL REFERENCES factories(id) ON DELETE RESTRICT,
+  work_order_id   UUID NOT NULL REFERENCES factory_work_orders(id) ON DELETE RESTRICT,
+  type            VARCHAR(32) NOT NULL,
```
lucaspin, line 12: Why not have only type + content columns, like factory_work_order_events?

db/migrations/20260805004843_add-factory-work-order-state-updated-at.up.sql
```diff
@@ -0,0 +1,38 @@
...
+
+--
+-- Dedicated timestamp bumped only on work order lifecycle transitions
+-- (`UpdateStatus`). Distinct from `updated_at`, which also bumps on
+-- assignee changes; the display-status logic uses this column as a
+-- fence so failures from a previous attempt don't stick to a reopened
+-- order after a reassign, and a reassign doesn't hide a genuine failure.
+--
+ALTER TABLE factory_work_orders
+  ADD COLUMN state_updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
```
lucaspin: Why do we need this if every update generates a factory_work_order_event record?

docs/design/factory.md
```diff
@@ -40,12 +41,51 @@ Run input shape (for app triggers):
...
 
-- **Create** — manual in UI/API today (`POST …/orders`).
-- **Assign** — optional assignees (`PATCH …/assignees`).
-- **Dispatch** — to a line (starts execution).
-- **Close** — `completed` or `rejected` (`PATCH …/close`). Open orders only.
+States (persisted on `factory_work_orders.state`; `result` is set only on `closed`):
 
-Display status in the UI (open orders): unassigned is an assignee filter only; status derives from executions — open, running, failed. Closed orders show completed / rejected.
+```
+[*] → draft → ready → open → closed
```
lucaspin: I don't understand why we are changing the work order states for this.

db/migrations/20260805004843_add-factory-work-order-state-updated-at.up.sql
andrecalil: The list view uses this to render the Failed pill without loading the events table for every card. Deriving the fence from `factory_work_order_events` works on the detail page (events are already loaded there) but the list endpoint returns work orders only — a per-row event scan would push us into N+1 territory or require a JSON aggregate. Adding the column keeps the list view's fence a single-row read; the write cost is one extra `Update` inside the same `UpdateStatus` transaction we already run.

lucaspin: Why not use https://github.com/superplanehq/superplane/blob/main/protos/factories.proto#L379 then? Seems weird to add a new database table column just so the UI can render a "failed" pill?

pkg/components/factory/add_work_order_comment.go
andrecalil: Addressed in cf84aa474. The three factory work-order components (`updateWorkOrderStatus`, `addWorkOrderComment`, `addWorkOrderArtifact`) no longer take a `workOrderId` — `FactoryContext.currentWorkOrder()` resolves the target from the running execution's `factory_work_order_executions` row and returns a clear "not attached to a work order" error if the canvas run wasn't dispatched from one.

pkg/components/factory/add_work_order_artifact.go
andrecalil: Addressed in cf84aa474. `data` is now a free-form list of `{name, value}` entries; the component flattens them to a `map[string]any` before persisting. The DB columns stay so `pr` links / titles / markdown bodies remain queryable, and future artifact types can add their own required fields alongside `data`.

pkg/components/factory/add_work_order_comment.go
andrecalil: Addressed in cf84aa474. `authorKind` and `authorLabel` are gone from the component config. Automation identity is now derived from the executing canvas node (`{nodeId, nodeName, appId, appName}`) and stored on the event author payload, so the timeline can render "commented via `<node>` in `<app>`" without a spoofable label. The REST endpoint only accepts `user` comments (attributed to the authenticated caller).

andrecalil: Addressed in cf84aa474 — same change dropped `authorLabel`. Agent identity is captured from the node/app that wrote the comment; when we add first-class agents, we'll swap the payload to `{agentId, agentName}` without changing the component surface.

docs/design/factory.md
andrecalil: Collapsed to `draft -> open -> closed` in cf84aa474.

db/migrations/20260805004843_add-factory-work-order-state-updated-at.up.sql
andrecalil: Done

db/migrations/20260805214128_add-factory-work-order-artifacts.up.sql
andrecalil: Done, I'll also drop `body` in favor of `data` only

docs/design/factory.md
```diff
@@ -47,12 +48,51 @@ linked source run (`source_run_id` on the work order record).
...
+| `closed → open` | `PATCH …/status` (`STATE_OPEN`) | Reopens the order and clears its `result`. |
+| `open → draft` | `PATCH …/status` (`STATE_DRAFT`) | "Back to draft" affordance when a run needs re-scoping. |
+
+Guardrails:
+
+- `dispatch` only works from `draft` or `open`, and still requires no active execution.
+- `close` only works from `open`; the API rejects invalid transitions (e.g. `draft → closed`).
+- `UpdateStatus` is the single writer for the lifecycle; `Close(...)` is kept as a thin wrapper for the existing REST endpoint and canvas components.
+
+Every transition writes an `order.status.updated` event (`fromState`, `toState`, `fromResult`, `toResult`). Two coarse legacy events fire alongside it so older timeline logic keeps working:
```
lucaspin: Why do we need order.status.update events if we have order.opened, order.closed? Seems like two different patterns, and I really prefer the current one

docs/design/factory.md
```diff
@@ -47,12 +48,51 @@ linked source run (`source_run_id` on the work order record).
...
+| `→ draft` | `POST …/orders` | Every new work order starts as `draft`. |
+| `draft → open` | `PATCH …/orders/{id}/dispatch` **or** `PATCH …/status` (`STATE_OPEN`) | Dispatch auto-promotes `draft` → `open` and records a status event. Explicit `PATCH …/status` also works ("open it without a dispatch yet"). |
+| `open → closed` | `PATCH …/orders/{id}/close` (legacy) **or** `PATCH …/status` (`STATE_CLOSED` + result) | Close **requires** a result: `completed`, `rejected`, or `failed`. |
+| `closed → open` | `PATCH …/status` (`STATE_OPEN`) | Reopens the order and clears its `result`. |
+| `open → draft` | `PATCH …/status` (`STATE_DRAFT`) | "Back to draft" affordance when a run needs re-scoping. |
+
+Guardrails:
+
+- `dispatch` only works from `draft` or `open`, and still requires no active execution.
+- `close` only works from `open`; the API rejects invalid transitions (e.g. `draft → closed`).
```
lucaspin: Why can't I close a draft work order?

docs/design/factory.md
andrecalil: This is from when we had extra statuses (my previous version). I'll clean it up, doesn't make sense on the current version of the FSM.

protos/factories.proto
```diff
@@ -358,14 +384,17 @@ message CloseWorkOrderResponse {
 message WorkOrder {
   enum State {
     STATE_UNSPECIFIED = 0;
+    // OPEN / CLOSED keep their original numbers; DRAFT appends.
```
lucaspin: No need to care about backwards compatibility here, just put them in the order they make sense now

protos/factories.proto
```diff
@@ -470,3 +499,80 @@ message WorkOrderEvent {
...
+  WorkOrderCommentAuthor author = 3;
+  google.protobuf.Timestamp created_at = 4;
+}
+
+message WorkOrderCommentAuthor {
+  enum Kind {
+    KIND_UNSPECIFIED = 0;
+    KIND_USER = 1;
+    KIND_AUTOMATION = 2;
+    KIND_SYSTEM = 3;
```
lucaspin: What is KIND_SYSTEM used for?

docs/design/factory.md
andrecalil: Fixed!

docs/design/factory.md
andrecalil: Fixed

protos/factories.proto
andrecalil: Done

protos/factories.proto
andrecalil: Dropped

web_src/src/pages/factories/lib/workOrderTimelineFromEvents.ts
```diff
@@ -148,8 +139,11 @@ function applyApiEventToTimeline(
     case "step.execution.finished":
       appendStepExecutionEvent(toDispatchBatchContext(state), payload, at, "step.execution.finished");
       return;
-    case "order.closed":
-      appendClosedEvent(state.events, index, payload, at, resolveUserName);
+    case "order.comment.added":
+      appendCommentEvent(state.events, index, payload, at, resolveUserName);
+      return;
+    case "order.artifact.added":
+      appendArtifactEvent(state.events, index, payload, at, resolveUserName);
```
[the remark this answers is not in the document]
andrecalil: not an issue

lucaspin (approved):

PR #6578, opened 2026-08-05
feat: link factory work orders to originating runs

lucaspin:
Work orders created by factory-app components (e.g. from a GitHub issue trigger) now retain a link to the originating workflow run and expose that context to line steps and the UI.

### Database changes

- Add `source_run_id` on `factory_work_orders`
- Set automatically when `CreateWorkOrder` runs inside a factory-app execution

### Line step execution

When a work order is dispatched to a line, if it originated from an app run, we include the root event for the run in the payload sent to the app run:

```json
{
  "work_order": {
    "id": "work-order-123",
    "title": "Fix the bug",
    "description": "The bug is in the code",
    "source": {
      "issue": { "number": 123, ... },
    }
  },
}
```

### Event changes

The `order.opened` event was updated to include the source run and app references.

**Work order created from an automation**

```json
{
  "type": "order.opened",
  "run": { "id": "run-123" },
  "app": { "id": "app-123" }
}
```

**Work order created manually**

```json
{
  "type": "order.opened",
  "user": { "id": "user-123" }
}
```


lucaspin:
👍

PR #6587, opened 2026-08-06
feat: add order() expression with lazy artifacts

forestileao:
## Summary
- Add `order()` expr builtin that resolves the live work order for factory-dispatched runs (`id`, `title`, `description`, `factory_id`, `state`, `result`, `source`), returning `nil` when the run is not attached to a work order
- Lazy-load `order().artifacts` only when the expression AST references that member (including inside `none`/`any`), leaving `root().data.work_order` unchanged
- Wire autocomplete + validation stubs and document preferred `order()` usage in factory design notes

## Test plan
- [x] `go test ./pkg/configuration/expressionvalidation` (order arity + `ExpressionUsesOrderArtifacts`)
- [x] `go test ./pkg/workers/contexts -run OrderFunction` (fields/source, artifact access equivalents, `none`/`any` PR gate, nil-when-absent)
- [x] Smoke in UI: on a factory-dispatched run, evaluate `order().id` and `none(order().artifacts, {#.type == "pr"})`
- [x] Confirm a non-factory run resolves `order()` to `nil` without error


forestileao:
Addressed Bugbot: wired `order()` through `exprEvaluator`, `__order` autocomplete example payload, and `normalizeSpecialFunctionExpr` (core + AutoCompleteInput). UI unit tests for those paths are green.

forestileao:
👍

lucaspin (approved):

PR #6589, opened 2026-08-06
feat: add branch artifact type for factory work orders

lucaspin:
- Adds a new `branch` factory work order artifact type that stores a single branch name in `data.name`.
  - Extends the **Add Work Order Artifact** canvas component with a Branch type and required Name field.
  - Renders branch artifacts in the work order artifacts panel and timeline (GitBranch icon + branch name).


lucaspin:
👍

PR #6610, opened 2026-08-09
fix: clarify PagerDuty user email label

### LINKED ISSUE #1898: [PagerDuty] Better copy for required email field
The email field is confusing:

Image

holahaikuflow:
## What changed
Renamed the PagerDuty action field label from `From Email` to `PagerDuty User Email` across incident actions.

## Why
`From Email` can be interpreted as an email sender address rather than the PagerDuty user performing the action. The new label makes the field purpose clearer while preserving the existing authentication-specific description and behavior.

## How
Updated only the user-facing label. No validation, authentication, or PagerDuty API behavior was changed.

## Tests
- `go test ./...` from `pkg/integrations/pagerduty`

Closes #1898

### DIFF
--- a/pkg/integrations/pagerduty/acknowledge_incident.go
+++ b/pkg/integrations/pagerduty/acknowledge_incident.go
@@ -77,7 +77,7 @@ func (c *AcknowledgeIncident) Configuration() []configuration.Field {
 		},
 		{
 			Name:        "fromEmail",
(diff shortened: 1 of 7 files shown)


holahaikuflow:
Hi! This PR has been open for a few weeks and is still mergeable with no conflicts. Happy to make any adjustments if needed — just let me know.

PR #6662, opened 2026-08-12
fix: stop failed runs showing Needs input

forestileao:
## Summary
- Automations list/detail aggregates no longer treat finished `RESULT_FAILED` as **Needs input** — Idle when nothing is active
- Lines phase aggregate stays idle for finished-only failures; run rows use `failed` kind
- Failed status dots are red (`#ef4444`); true waiting stays orange

## Test plan
- [x] `factoryAutomationStatus.spec.ts` / `linePhaseRuns.spec.ts`
- [ ] Automations list: app with only passed/failed runs shows Idle (gray)
- [ ] Automations run row: Failed shows red tick
- [ ] Lines phase header idle when nothing executing; failed run row red


forestileao:
👍

PR #6666, opened 2026-08-12
fix: show one Lines board card per work order

forestileao:
## Summary
- Lines phase board now places each work order once, in the column for its furthest current step (prefer active executions, else furthest finished).
- Passed/failed labels come from that single execution; phase ticks still aggregate only cards left in the column.
- Specs cover multi-step progress, mid-line failure, and unique `workOrderId` across the board.

## Test plan
- [x] `npx vitest run src/pages/factories/lib/linePhaseRuns.spec.ts` (in app container)
- [ ] Manual: Bugs (or any) line — each WO title appears once; progressing WOs only under their active phase


forestileao:
👍

PR #6672, opened 2026-08-13
fix: Scope Factory Run Status to Selected Run

forestileao:
## Summary

Factory Live canvases painted each node from canvas-wide `lastExecutions`. That made work-order and automation views show stale status (for example PASSED from a run 6h earlier) on nodes the current run never reached.

This change keeps factory run views scoped to `?run=`, hides runtime status on factory Live, and opens the right inspector only after the user selects a node.

## Test plan

- [ ] From a work order, open a run that has not reached a later node; confirm that node has no PASSED/timestamp from another run
- [ ] Close the run inspector; confirm `?run=` stays and node status stays run-scoped
- [ ] Open factory Live without `?run=`; confirm nodes show topology only (no status footers)
- [ ] Click a node in a factory run; confirm the right inspector opens for that node only
- [ ] `make check.lint.ui` passes


forestileao:
👍

PR #6716, opened 2026-08-14
fix(workorder): Persist unassigned work order users

### LINKED ISSUE #6681: Workorder: Cannot unassign users
**Description:**
Deselecting assignees and clicking Save shows a success toast, but the users stay assigned. Assigned users are also sorted alphabetically with everyone else, so they are hard to find in the picker.

**Steps to reproduce:**
1. Open a work order that has one or more assignees.
2. Open the assignee picker, deselect those users, and click Save.

**Expected / Actual:**
Expected: Those users are unassigned, and assigned users appear at the top of the picker. Actual: Toast says assignees updated, but they remain assigned.

**Context:**
Picker and save live in `WorkOrderAssigneesPopover` / `WorkOrderAssigneePicker` (overview sidebar). `useUpdateWorkOrderAssignees` sends `assigneeIds` and always toasts “Assignees updated.” Keep assigned members at the top of the list so unassign does not require a search.

mytmlt:
## Summary

Users could not unassign people from a work order. Deselecting assignees and clicking Save showed a success toast, but the users stayed assigned. This change fixes that bug and keeps assigned users at the top of the picker.

The root cause is in the backend. `FactoryWorkOrder.UpdateAssignees` writes `updated_at` with `tx.Model(o).Update(...)`, and GORM auto-saves the loaded `Assignees` association on that update. The join rows that `ReplaceAssignees` just deleted are upserted back, so every removal is undone. The frontend also discarded in-progress selections while the picker was open, because the popover re-synced its draft from the `selectedIds` prop on every render.

## UI

- WorkOrderAssigneesPopover syncs its draft only when the popover opens, so refetches and re-renders no longer clobber in-progress edits.
- WorkOrderAssigneePicker keeps selected users at the top of the list, then alphabetical.

## API / Backend

- factory_work_order.go: `UpdateAssignees` now writes `updated_at` with `Omit(clause.Associations)` so the association save that re-added removed assignees no longer runs.

## Tests

- factory_work_order_assignees_test.go: `TestFactoryWorkOrder_UpdateAssigneesUnassignsAll` clears all assignees and asserts the join rows are gone and the `order.assignees.updated` event records the full unassign diff.

Closes #6681

### DIFF
--- a/pkg/models/factory_work_order.go
+++ b/pkg/models/factory_work_order.go
@@ -11,6 +11,7 @@ import (
 	"github.com/superplanehq/superplane/pkg/models/factory"
 	"gorm.io/datatypes"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 const (
@@ -193,7 +194,7 @@ func (o *FactoryWorkOrder) UpdateAssignees(tx *gorm.DB, assigneeIDs []uuid.UUID,
 
 	now := time.Now()
 	o.UpdatedAt = now
-	if err := tx.Model(o).Update("updated_at", now).Error; err != nil {
+	if err := tx.Model(o).Omit(clause.Associations).UpdateColumn("updated_at", now).Error; err != nil {
 		return err
 	}
 
--- a/pkg/models/factory_work_order_assignees_test.go
+++ b/pkg/models/factory_work_order_assignees_test.go
@@ -1,10 +1,13 @@
 package models
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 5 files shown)


mytmlt:
Hello, I started this PR yesterday as a part of the requirement for the job application and I was genuinely interested in   the project once I learned about it from the job posting. It took some time to onboard and get up and running and in the meantime I saw that there is already similar fix by a bot. Let  me know if you want to close this or not.

andrecalil:
Hey @mytmlt thanks for working on this PR. We'll keep the other one, as that whole feature is under current development and test. I appreciate your interest.

PR #6720, opened 2026-08-14
feat: Consistent Workspace Header

andrecalil:
## Summary

Each workspace page had its own header. Titles used different sizes, spacings, and typography. Detail pages put the back link in the body instead of the header. Work order detail did not show the `SP-42` identifier. This made the product feel inconsistent and hurt navigation.

This change adds one shared `WorkspacePageHeader` component with two variants and moves every workspace page onto it:

- **Section variant** — Overview, Work Orders list, Lines list, Automations list, Wiki, Velocity, Settings, Coming Soon. Title, optional subtitle, optional leading slot for scope pills, trailing actions, and an optional row below for filter chips.
- **Entity variant** — work order, line, and automation detail. Back link and optional identifier kicker above the entity title, with trailing actions in the header.

Other changes that come with the migration:
- Work order detail now shows the `SP-42` kicker above the title. The sticky header and negative-margin hack are gone.
- Line and automation detail move their back link out of the body and into the header. The duplicate emphasized automation card is removed; status and description move to the subtitle.
- Velocity moves its 7d / 30d toggle from the body into the header.
- Work Orders list drops the extra `h-8` on toolbar triggers and uses the standard `sm` size.
- Create, secondary, and icon buttons follow one style: filled `sm` with `Plus` for create (`New work order`, `New line`, `New automation`), outline `sm` for secondary (`Refresh knowledge`, line `Edit`), and ghost `icon-xs` for Copy and overflow.

### DIFF
--- a/web_src/src/pages/factories/WorkOrderDetailHeader.stories.tsx
+++ b/web_src/src/pages/factories/WorkOrderDetailHeader.stories.tsx
@@ -1,20 +1,23 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
 
 import { ComponentStoryShell } from "./__fixtures__/ComponentStoryShell";
+import { withFactoriesTheme } from "./__fixtures__/factoriesStoryTheme";
 import { WorkOrderDetailHeader } from "./WorkOrderDetailHeader";
 
 /**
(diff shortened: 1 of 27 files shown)


andrecalil:
👍🏼

PR #6726, opened 2026-08-15
fix: Stop SSH command timeout goroutine leak on early return

alxn787:
## Summary

SSH command timeouts leaked a goroutine when the command finished early.
That goroutine held the session in memory until the full timeout ended.
Canvases that run many SSH nodes then grew one leak per execution.

This change stops the timer when the command returns.
A command that does not complete still hits the deadline.

## Test plan

- [x] Run `make test PKG_TEST_PACKAGES=./pkg/components/ssh`.
- [x] Confirm `TestExecuteScriptDoesNotLeakTimeoutGoroutine` passes.


alxn787:
@lucaspin @markoa

PR #6731, opened 2026-08-15
feat: Replay a single node from a past run

### LINKED ISSUE #3320: Test mode: run a single node against stored payloads
**Describe the request:**

When something breaks in a canvas, the debugging loop today looks like this: read the logs, guess what went wrong, edit the node, push an update, then wait for a new external event (a git push, a webhook, a cron tick) to see if your fix worked. If the event doesn't come, you manufacture one. If it does come but the data is slightly different from the original failure, you're not even testing the same scenario. Repeat until it works or you give up.

This is the single biggest time sink when building event-driven workflows. The payload that caused the failure *already exists* in the execution history. We should be able to point at it and say "run this node again with that exact input."

What we need is a **test mode**: the ability to pick any node in a canvas, select a payload from a previous execution (or paste one manually), and run just that node in isolation. No full workflow re-execution, no waiting for external events, no side effects on other nodes.

**Describe your use-case:**

When a node breaks in an event-driven workflow, every fix attempt requires a full cycle: edit the node, push the update, wait for a new external event (webhook, cron), then check the result. The payload that caused the failure already exists in execution history, but there's no way to reuse it. You end up manufacturing new events just to re-test the same scenario, and each cycle takes minutes. Replay would turn that into seconds.

**Describe functionality:**

**Core: single-node replay**
- Select any node in a canvas (from the UI or CLI)
- Choose an input payload: pick from a previous execution's stored payloads, or provide a custom one
- Run that node in isolation. It executes exactly as it would in a real run, but nothing upstream or downstream fires
- See the full output: result data, stdout, stderr, errors, duration, same as a normal execution detail view

**Optional: payload overrides**
- Before replaying, let the user edit specific fields in the stored payload (e.g., change `before`/`after` commit SHAs, swap the repository name, modify a configuration value)
- This lets you test edge cases without needing to manufacture real events that produce those exact values
- Show a diff between the original payload and the modified version so it's clear what changed

**Replay lineage**
- Tag replay executions distinctly in execution history (e.g., a "replay" badge or label) so they don't get confused with real runs
- Link each replay back to the original execution it was derived from
- This matters for audit trails. You need to know which runs were real events and which were test iterations

**Scope boundaries**
- Start with single-node replay. This covers 90% of the debugging pain
- Subgraph replay (run a chain of 2-3 connected nodes) is a natural follow-up but not required for v1
- Replay should not trigger side effects on downstream nodes unless the user explicitly opts into a "continue from here" mode (future)

**What this is not:**
- Not a full workflow re-trigger. That already exists and is slow, which is the whole problem
- Not a mock/stub framework. The node runs for real with real integrations, just with a controlled input
- Not a unit test runner. There's no assertion language or pass/fail criteria, it's an interactive debugging tool

**Impact:**
Turns the debugging loop from "edit, push, wait for event, check" into "edit, replay, check." The feedback cycle drops from minutes to seconds. This is especially critical for nodes that depend on external events (webhooks, cron) where you can't control when or how the next trigger arrives.

aartem1:
## Summary

- To debug one node today, an operator starts the full run and repeats the side
  effects of every other node.
- This change replays one node against a past input, isolated from the graph, so
  no downstream node starts.
- The operator can edit the payload before the launch and compare it with the
  original payload.
- Replay works from the component sidebar, the Run inspector, and the CLI, and
  each replay run keeps a link to its origin.

Closes #3320

## Test plan

- [x] Run `make test`, `make lint`, and `make check.build.app`.
- [x] Run `make check.build.ui`, `make check.lint.ui`, and `make check.test.ui`.
- [x] Run `make check.db.migrations` and `make check.db.structure`.
- [x] Replay a node from the component sidebar. Confirm that no downstream node
      starts.
- [x] Replay a step from the Run inspector. Confirm that the editor shows the
      difference from the original payload.
- [x] Replay an aggregating node. Confirm that the modal asks for one input per
      live incoming source.
- [x] Open the runs list. Confirm that replay runs stay hidden until the filter
      shows them.
- [x] Open a replay run in the inspector. Confirm that it shows its lineage.
- [x] Replay an aggregating node with two sources. Confirm that the Input card
      counts both sources.


aartem1:
@forestileao @lucaspin @shiroyasha — a review ping when you have a moment.

The branch is rebased onto `main` (`0da51ffc5`) and green: `make lint`, `check.build.app`, the Go suite (13045 tests), `check.build.ui`, `check.lint.ui`, `check.test.ui` (6598 tests), plus the migration and `db/structure.sql` checks. The four replay migrations were reissued so they stay newer than everything on `main`.

The PR is `MERGEABLE` and blocked only on approvals (policy-bot, 0/6). Happy to split it up or walk through any part of it if that makes the review easier.

PR #6753, opened 2026-08-17
chore: make queue item processing an opt-in capability

lucaspin:
Every component was required to implement ProcessQueueItem(), and all but two of them (merge and loop) contained the same boilerplate call to ctx.DefaultProcessing(). This duplicated ~2,400 lines across components and integrations, and every new component had to repeat it. Remove ProcessQueueItem from the core.Action interface. Add an optional core.QueueItemProcessor interface for components that manage their own queue items (merge, loop). The node queue worker now runs the default processing itself — create the execution, consume the item, mark the node as processing — unless the component implements the interface. DefaultProcessing is removed from ProcessQueueContext, because the engine now owns that path. The registry's panic-recovery wrapper exposes the underlying processor so worker dispatch keeps the panic protection. Behavior is unchanged for all components.


lucaspin:
👍

PR #6771, opened 2026-08-18
feat: add per-node concurrency configuration

lucaspin:
Nodes run one execution at a time. For long-running components, such as coding agents, this limit blocks the whole backlog behind one execution. This change adds an optional `concurrency` block to action nodes:
- `max` sets the maximum number of concurrent executions. The default is 1, which keeps the current behavior.
- `key` is an optional template expression. Each resolved value gets an independent queue with its own `max` slots.

## Engine and workers

- The node `state` column no longer acts as a dispatch mutex. The worker derives capacity from counts of active executions per resolved queue name.
- The poll query returns only actionable nodes. Nodes with a full backlog stay out of the dispatch pass until an execution finishes.
- The worker resolves each queue item's `key` expression once, then stores the resolved queue name on the item and on its execution.
- If a key expression fails or resolves to null, the item moves to the node's implicit queue and the failure is logged.

## Components

- Merge and loop manage their own queue items. The worker does not capacity-gate them.
- Loop supports `max` as its parallel session cap. Active sessions include pending executions, so one dispatch pass cannot exceed the cap.
- Merge rejects the `concurrency` block. Its throughput is unbounded across runs by design. Loop rejects `key`, because its backlog mixes session starts with feedback for running sessions.

## API

- New `ConcurrencySpec` proto message with optional `key` and `max` fields on `Node`. Absent means one execution at a time.
- Canvas YAML validation rejects `max` below 1, concurrency on non-action nodes, and the unsupported fields on merge and loop.

## Database

- `workflow_nodes` gets typed `concurrency_key` and `concurrency_max` columns with a `max >= 1` check constraint.
- `workflow_node_queue_items` and `workflow_node_executions` get a `queue_name` column. A partial index on active executions backs the capacity counts.
- The migration resets nodes stuck in `processing` and backfills `queue_name` for in-flight executions, so capacity counts see them.
- KV-to-execution lookups now log an error when one key/value pair matches more than one execution. This exposes webhook routing collisions that were silent before.

## UI

- The node settings sidebar gets a "Concurrency" section with the key and max fields. The key input supports expression autocomplete.
- Merge nodes do not show the section. Loop nodes show only `max`.


lucaspin:
👍

PR #6772, opened 2026-08-18
fix(retry): apply InitialDelay once before the first attempt

EBEN4REAL:
## What

`retry.WithConstantWait` calls `time.Sleep(options.InitialDelay)` at the **top of the retry loop**, so the delay is applied before *every* attempt and stacks with the regular `Wait` between retries (`InitialDelay + Wait` per retry). That contradicts the field's name, which implies a **one-time** delay before the first attempt.

This PR moves the sleep outside the loop so `InitialDelay` runs exactly once, and adds the first unit tests for `pkg/retry`.

```go
// before
for attempt := 1; ; attempt++ {
    time.Sleep(options.InitialDelay) // runs before every attempt
    err := f()
    ...
    time.Sleep(options.Wait)
}

// after
time.Sleep(options.InitialDelay) // runs once
for attempt := 1; ; attempt++ {
    err := f()
    ...
    time.Sleep(options.Wait)
}
```

## Why it's low-risk

The only caller (`pkg/integrations/jira/webhook_handler.go`) leaves `InitialDelay` at its zero value, so runtime behavior is unchanged today; this fixes the latent bug and makes the field behave as its name promises.

## Testing

Added `pkg/retry/retry_test.go` (the package previously had no tests):

- succeeds on first attempt
- retries until success
- returns an error after giving up
- **regression test** asserting `InitialDelay` is applied once, not per-attempt (verified to fail against the previous implementation and pass after the fix)

```
ok  github.com/superplanehq/superplane/pkg/retry
```

Signed off per DCO.


hmadhsan (commented):
LGTM! Excellent bugfix separating InitialDelay from the per-attempt retry loop, complete with comprehensive unit tests.

PR #6780, opened 2026-08-18
fix(sns): cache SNS signing certificate instead of re-fetching every webhook call

sloonzz:
## Summary
- Every incoming SNS webhook message re-fetched and re-parsed the signing certificate from `SigningCertURL`, even though AWS reuses the same certificate for a long time (there was already a `TODO` noting this).
- Add an in-memory cache in `pkg/integrations/aws/sns/on_topic_message.go`, keyed by the certificate URL (which is already validated as a genuine AWS SNS domain before use), bounded by a 24h TTL and the certificate's own `NotAfter`. Cache hits are re-checked against `NotBefore`/`NotAfter` so an expired certificate is never served stale.
- The cache is a package-level `sync.Map` rather than a field on `OnTopicMessage`, since a fresh `OnTopicMessage{}` is constructed on every trigger lookup (`Registry.GetIntegrationTrigger`) — a struct field would never actually persist across requests. `sync.Map` is safe for the concurrent webhook requests that share it within a process.
- Scope note: caching is per-process, so with multiple server replicas each one builds its own cache independently, and a restart clears it. That matches the intent of the original TODO (avoid re-fetching on every call) rather than a cluster-wide shared cache.

## Test plan
- [x] `go test ./pkg/integrations/aws/sns/...` — existing tests pass, plus 3 new tests covering: cache hit avoids a second HTTP fetch, an expired cache entry triggers a re-fetch, and different certificate URLs are cached independently.
- [x] `go vet ./pkg/integrations/aws/sns/...` and `gofmt -l` clean.
- [x] `go build ./...` for the full repo.

### DIFF
--- a/pkg/integrations/aws/sns/on_topic_message.go
+++ b/pkg/integrations/aws/sns/on_topic_message.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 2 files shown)


hmadhsan (commented):
LGTM! Excellent performance improvement caching SNS signing certificates with proper expiration checks and thorough test coverage.

PR #6787, opened 2026-08-18
fix: Report the cause of gateway server errors

LevRiabov:
The gateway sanitizes handler errors before it writes the response. The cause was not logged and not sent to the error tracker. A failed request thus created an issue that contains the status code and the path only. Issues 6693 to 6697 are examples. Each one is titled "HTTP 500 /api/v1/..." and shows nothing about the failure.

Keep the cause in a per-request recorder in the request context. The two gateway error paths record it, and the HTTP logging middleware sends it as an exception. If a handler writes a 5xx response without an error, the middleware sends the previous message.

Record server errors (5xx) only. Client errors are caused by the request and are not reported.

The response to the client does not change. It stays sanitized.


hmadhsan (commented):
LGTM! Great design preserving client response sanitization while capturing full root-cause exceptions in Sentry.

PR #6789, opened 2026-08-19
fix: Restore Global Notification Event Types

forestileao:
## Summary

All workspaces hid the event type list. Users could not turn off assignment, comment, or status emails.

This change stores global event types when the scope is All. Filtered workspaces keep a type list for each workspace.

## Test plan

- [ ] Open workspace settings Notifications with All workspaces selected
- [ ] Confirm the Notify me about event types are visible
- [ ] Turn off one type, save, and reload the page
- [ ] Confirm that type stays off
- [ ] Select Filtered, add one workspace, and set a different type list
- [ ] Save and confirm All uses the global list and Filtered uses the workspace list
- [ ] Select None and confirm no event type emails are sent


forestileao:
👍

PR #6814, opened 2026-08-19
feat: add per-step max parallelism to factory lines

lucaspin:
A factory line step starts a run for every work order that reaches it.
Long-running steps, for example coding agents, can overload the system
or external tools. This change adds a `maxParallelism` limit to each
line step. When a step is at its limit, new work orders wait in a
first-in-first-out queue for that step. The default limit is 10.

## Database
- New table: `factory_work_order_queue_items`. One row is one line
  dispatch that waits for admission into one step. Each row stores the
  line dispatch, the step index, and the step name.
- A dispatch can wait at only one step at a time (unique index on
  `line_dispatch_id`).

## API

- `FactoryLine.Step` has a new optional `max_parallelism` field. The
  value must be at least 1. An unset value means the server default
  (10).
- `WorkOrderLineDispatch` has a new optional `queue_item` field. It
  shows the step the dispatch waits at and its 1-based queue position.
- Line create and update actions validate the new field.

## Models

- `FactoryWorkOrderLineDispatch.EnqueueOrStartStep` counts the active
  runs for a step. It starts the step when a slot is free. It creates a
  queue item when the step is at its limit.
- Admission decisions take a row lock on the factory line. Two
  concurrent decisions for the same line cannot both use the last free
  slot.
- A `step.execution.queued` event is added to the work order timeline
  when a dispatch starts to wait.
- When a work order closes, its queue items are deleted and the
  related dispatches are cancelled.
- The factory resource cleaner deletes queue items before the other
  factory resources.

## Workers

- When a run for a line step completes, the run finalizer admits the
  oldest queued work order for that step, if a slot is free.
- Admission uses the same lock and capacity check as dispatch.

## UI

- The line step editor has a new "Max parallelism" number input next
  to the Automation and Trigger inputs.
- The work order details page shows a queued step row with its queue
  position, and a "Queued at <step> — waiting for a free slot" entry in
  the activity timeline.
- The Lines phase board shows a "Queued #N — waiting for a free slot"
  note under the card of each queued work order. The board column tick
  also shows the queued state.


lucaspin:
👍

PR #6821, opened 2026-08-19
feat: add run-level error recording

lucaspin:
Canvases need a way to record a business failure on a run without failing a node execution. This change stores those messages on the run and shows them when a person opens that run.
The Add Run Error component appends a message to the current run. Downstream nodes still receive an event. When the run finishes, any recorded error marks the run as failed.

### Database

- Adds an errors JSONB column on workflow_runs. Drops result_message on that table. The new list of errors replaces the single result message.
- Each append uses FOR NO KEY UPDATE. Concurrent writes do not overwrite each other. Child foreign-key inserts on the run can still proceed.
- Limits: 50 errors per run, and a size cap from the payload limit.

### API

- CanvasRun and CanvasRunRef now include errors as a list of strings. Describe-run, list-runs, and run refs return that list.

### Components and workers

- Add Run Error is a SuperPlane building block. It records the configured message on the run. The step itself passes unless the store rejects the message (too many or too large).
- Run initialization records fail messages through the same AddError path. Run-finished callbacks include the error list. Run App forwards child-run errors on the failed output channel.
- CalculateResult returns failed when the run has one or more recorded errors.

### UI

- The run inspector shows a red This run has an error / This run has errors card with the full messages.
- Regular apps show the card at the top of the run inspector. Factory apps show the same card in the node inspector. If no node is selected, the card appears on the canvas.
- Add Run Error is listed in the SuperPlane building-block category.


lucaspin:
👍

PR #6834, opened 2026-08-20
fix: Load admin org flags via admin API

forestileao:
## Summary

Fixes the installation admin organization page so it can show experimental features.
The page used the member organization API. Installation admins are not always members, so that request returned 404.
The page now loads flags from the admin experimental-features route.

## Test plan

- [ ] Open `/admin/organizations/{id}` as an installation admin who is not a member of that organization.
- [ ] Confirm the Experimental Features table loads.
- [ ] Enable a feature and confirm the switch stays on after reload.
- [ ] Disable the feature and confirm the switch stays off.


forestileao:
👍

PR #6872, opened 2026-08-24
fix: Route factory false forks into left handles

forestileao:
## Summary

The factory split-run canvas drew the `false` branch from If up and over the next node.
That path looked wrong because the side node already sits on the same row.
This change attaches that branch to the left handle of the side node.

## Test plan

- [ ] Open a work order split-run popup with an If node that has a `false` branch.
- [ ] Confirm the `false` edge leaves the right of If and enters the left of the side node.
- [ ] Confirm the `true` spine still goes down from the bottom of If.
- [ ] Confirm channel labels still show on the paths.


forestileao:
👍

PR #6951, opened 2026-08-27
fix: Open the Unique Line on Workspace Visit

forestileao:
## Summary

ListFactories omitted line ids. The workspace switcher and the organization workspace index then opened the workspace index.

A workspace with one line must open that line board in one step.

ListFactories now returns lines. The client builds `/workspaces/{key}/lines/{id}` without a stop on the Lines list.

## Test plan

- [ ] Open an organization whose last workspace has one line.
- [ ] Confirm the browser opens `/workspaces/{key}/lines/{lineId}`.
- [ ] Switch to that workspace from another workspace.
- [ ] Confirm the browser does not show the Lines list.
- [ ] Open `/workspaces/{key}/lines` and confirm the redirect to the line board.
- [ ] Open a workspace with no line and confirm overview still opens.

### DIFF
--- a/pkg/grpc/actions/factories/list_factories.go
+++ b/pkg/grpc/actions/factories/list_factories.go
@@ -3,6 +3,7 @@ package factories
 import (
 	"context"
 
+	"github.com/google/uuid"
 	"github.com/superplanehq/superplane/pkg/database"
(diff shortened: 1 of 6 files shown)


forestileao:
👍

PR #7002, opened 2026-08-28
feat: track LLM usage on PR runs

lucaspin:
Factory pull request runs did not record LLM usage against the work order.
That hid cost on the work order and undercounted work-order totals.
This change records factory-canvas spend when the work order is known.
The work order log shows cost, tokens, and duration on line and pull request phases.
Backlog analysis spend is recorded in the ledger and included in work-order totals.
This change does not show cost on each analysis run.
This change does not backfill old events.


lucaspin:
👍

PR #7021, opened 2026-08-30
fix: Stop the JSON viewer from truncating long string values

### LINKED ISSUE #6407: Run inspector truncates long string values even in fullscreen — failure errors unreadable
### Describe the bug

The run inspector's JSON viewer truncates long string values with an ellipsis — including in the **fullscreen** view — so failure payloads whose message is one long string are unreadable in the UI.

### Steps to reproduce

1. Produce a failed HTTP execution (e.g. a blocked/unreachable URL). The failure event is `{"data": {"error": "error executing request: Get \"…\": dial tcp …: connection blocked: …"}, "type": "http.request.failed"}`.
2. Open the run inspector → OUTPUT · FAILURE section: the `error` value renders as `error executing request: Get \"...` — truncated.
3. Open fullscreen: still truncated. The only ways to read the actual error are the Copy button (paste elsewhere) or querying `workflow_events` directly.

### Expected

Long string values should be readable in place — wrap, expand-on-click, or at minimum untruncated in fullscreen. Error strings are exactly the values users need whole.

Environment: self-hosted dev at current main.

liamoconnor87:
## What

Every JSON payload viewer in the app now shows string values in full and wraps long lines. They previously truncated strings after 30 characters with an ellipsis.

There were 12 render sites across 7 files, each configured slightly differently. This replaces all of them with one shared `<JsonPayload>` component (`web_src/src/ui/JsonPayload.tsx`), so `@uiw/react-json-view` is now imported in exactly one non-test file.

| Surface | Before | After |
|---|---|---|
| Run inspector timeline (OUTPUT / FAILURE) | truncated | full + wrapped |
| Node detail Payload / Config tabs | truncated, no wrapping | full + wrapped |
| Building-blocks component payload preview | truncated | full + wrapped |
| Component-sidebar event payload (inline + modal) | truncated, hardcoded light-mode colors | full + wrapped + dark mode fixed |

## Why

Fixes #6407. When a node fails, its error is one long string in the payload (`{"error": "error executing request: Get \"…\": dial tcp …: connection blocked: …"}`). The viewer cut it at 30 characters, fullscreen included, so the part of the message that explains the failure was only reachable via the Copy button or a direct `workflow_events` query.

The viewers had also diverged over time: some wrapped, some didn't, and the two in the component-sidebar event view hardcoded `#ffffff` / `#24292e`, making them unreadable in dark mode.

## How

The library truncates via `shortenTextAfterLength` (default 30) and hands the already-truncated text to any custom string renderer, so wrapping alone can't recover the missing characters. `shortenTextAfterLength={0}` disables truncation and passes the full string through.

- New `web_src/src/ui/JsonPayload.tsx`: `shortenTextAfterLength={0}`, a string renderer that wraps and keeps JSON escapes visible, and the theme style resolved from `useTheme()` internally.
- All 12 render sites migrated. `runInspectorJson.ts` folded into `lib/jsonViewTheme.ts`.
- `JsonPayload` defaults to `collapsed={2}`, carried over from the previous component. The three sites that rendered a bare `JsonView` with no `collapsed` now pass `collapsed={false}` so they stay fully expanded as they were.
- 13 files changed, +154 / -143.

## Tests

`JsonPayload.spec.tsx` and `RunNodeDetailTabSection.spec.tsx` render a long error string and assert the full value is present with no ellipsis. I checked they fail against the pre-fix code.

Rebased on main today, which brought in the bun test migration (#7408), so both specs import from `bun:test`. The 14 specs covering these files pass, 96 tests via `make check.test.ui`.


liamoconnor87:
Rebased on main and this is ready for another look.

Main moved a fair bit while this sat. #7408 migrated the UI tests to bun, so both my specs now
import from bun:test rather than vitest.

Also, migrating the building-blocks preview and the two event payload viewers onto the shared
component would have collapsed them below depth 2, since JsonPayload defaults to collapsed={2}.
They rendered a bare JsonView with no collapsed prop before, so they now pass collapsed={false}
to stay as they were. The fullscreen event modal keeps its 14px as well.

The 14 specs that touch these files pass, 96 tests.

Thanks,

Liam.

PR #7076, opened 2026-09-01
fix: Align factory automation canvas layout

forestileao:
## Summary

The settings Automation tab did not show the same leaf-right graph as run inspect. The preview also hid nodes instead of fitting the full graph.

Edit automation reused live positions and then moved nodes into the edit layout. That looked like a slide.

Factory edit plus-stems used handle-border gray on a matching canvas, so the stem disappeared. The plus sat off the 10px factory port.

This change uses the run-inspect layout in the settings preview, fits the graph at once, holds a loading cover until Configure layout is ready, and paints plus-stems with the edge stroke, centered on the port.

## Test plan

- [ ] Open a factory line. Open Settings. Open the Automation tab.
- [ ] Confirm the graph matches Automations run inspect and fits in the pane.
- [ ] Click Edit automation. Confirm a short loading cover, then the edit graph with no slide.
- [ ] Confirm unused channel plus buttons show a visible stem through the center of the plus.
- [ ] Run `make check.lint.ui` and `make check.build.ui`.


forestileao:
👍

PR #7083, opened 2026-09-01
feat: Add linked accounts, separate from sign-in methods

shiroyasha:
## Summary

Velocity reports credit a merged pull request to the GitHub login that
authored it. Until now the only way to give SuperPlane that login was to
sign in with GitHub. A member who signs in with a password or with Google
appeared twice in the People table: once as themselves, and once as the
GitHub author of their own work.

A sign-in method is the wrong place to hold this. It answers how a member
signs in, not who the member is on another service. A member also needs to
state a GitHub identity without adding a way to sign in.

This pull request adds **linked accounts** as a separate concept:

- A linked account is an identity the member owns on another service. It
  has its own table and carries no token.
- The connect flow proves ownership through GitHub OAuth, but it issues no
  session and writes no credential. A linked account never becomes a way to
  sign in.
- A unique index keeps one identity to one account, so two members cannot
  claim the same author.
- The Velocity query prefers the linked account and falls back to the GitHub
  sign-in identity, so members who joined through GitHub keep their
  attribution.
- Only GitHub is linkable, because only GitHub holds activity to attribute.

The Security page and its sign-in methods do not change.

## Separate bug fix

The first commit fixes a defect this work uncovered. GORM maps the
`GitHubLogin` field to the column `git_hub_login`, so the `github_login`
alias in the member query never bound. Every member looked like they had no
GitHub identity, which means the People table could never match a member to
an author. The field now names its column.

## Test plan

- [x] `TestDescribeFactoryVelocity_JoinsLinkedGitHubAccount` shows the
      duplicate row before the link and one row after it
- [x] `TestAccountLinkedAccount` covers link, replace, refuse-in-use, remove
- [x] `FactorySettingsLinkedAccountsPage.spec.tsx` covers connect, remove,
      cancel, and both redirect results
- [x] `make lint`, `make check.build.app`, `make check.lint.ui`,
      `make check.build.ui`
- [x] `./pkg/authentication/...` and `./pkg/public/...` pass
- [ ] Link a GitHub account in Settings, then confirm the People table shows
      one row for the member
- [ ] Confirm the link adds no GitHub sign-in method on the Security page


shiroyasha:
Closing this pull request. #7088 shipped the same capability with a larger scope.

`main` now links a provider to the signed-in account through
`/auth/<provider>?intent=link`, and it stores the provider login in
`account_providers.username`. This is the field that the velocity People
report joins on, so the duplicate person row is fixed after a member
connects GitHub in Account settings.

The version in `main` also supports Google, supports disconnect, and
prevents the removal of the last sign-in method. This pull request adds a
second OAuth link path with a different state format and no disconnect
support, so it is redundant.

PR #7165, opened 2026-09-03
feat: shut down gracefully, draining in-flight work on SIGTERM

### LINKED ISSUE #6987: Add graceful server and worker shutdown
## Summary

SuperPlane does not perform a graceful shutdown when the process receives a termination signal. The server starts workers with `context.Background()` and then blocks forever with `select {}`.

The public HTTP server, background workers, message consumers, and telemetry services do not receive a shared shutdown signal. Active work can stop in the middle of a database transaction or external API operation when a deployment replaces the process.

## Impact

- Deployments can interrupt active canvas and integration work.
- RabbitMQ consumers can stop without an orderly close.
- Background workers cannot stop accepting new work before process exit.
- HTTP requests do not receive a shutdown grace period.
- Telemetry data can remain buffered when the process exits.

## Expected behavior

SuperPlane handles termination signals and stops each service in a controlled order. The process waits for active work until a configured shutdown deadline expires.

## Current behavior

The process relies on immediate operating system termination. Worker contexts never receive cancellation, and the server has no shutdown lifecycle.

Vandit1604:
Closes #6987.

`pkg/server/server.go` ended in `select {}` and started its 22 workers with `go w.Start(context.Background())`. Nothing was ever cancelled, so SIGTERM killed the process wherever it happened to be.

Two things had to be true before that mattered.

The server never received SIGTERM. `docker-entrypoint.sh` ended with `/app/build/superplane` and no `exec`, so bash stayed PID 1 and kept the signal. I checked it in a container: without `exec` the process runs as PID 7 and exits 137 on SIGKILL, with `exec` it runs as PID 1 and exits 0.

Waiting for worker loops is not waiting for work. Sixteen workers hand each task to a goroutine behind a weighted semaphore. Returning on `ctx.Done()` abandoned up to 25 of them per worker, mid-transaction, which is the first impact the issue lists. `drainTasks` acquires the full weight, which succeeds only once every task has released its slot.

## What changed

`Start()` derives a context from SIGINT and SIGTERM and blocks on it. `startWorker` tracks each worker in a `WaitGroup`. Workers drain in-flight tasks before returning, and stop handing out new ones once shutdown starts. The semaphore limits live in `pkg/workers/shutdown.go` and size both the semaphore and the drain, so the two cannot drift.

The public API gets `Shutdown(ctx)` beside the existing `Close`, and `http.ErrServerClosed` no longer reaches `log.Fatal`. The two email consumers take a context. Their reconnect loop used a plain `time.Sleep`, so `Start()` never returned and would have hung the drain.

One deadline covers the drain, the tracing shutdown and the Sentry flush. Separate budgets could run past the grace period between them. Default 25s, `SHUTDOWN_TIMEOUT` overrides.

## Not refusing requests on the way out

Closing the listener on SIGTERM would have been a regression. Kubernetes removes the pod from Endpoints asynchronously, and the chart polls `/` every 10s (`api.yaml:146`), so requests still arrive. Today bash swallows the signal and those requests are served. A naive fix would answer them with connection refused.

`SHUTDOWN_DRAIN_DELAY` keeps the API listening for 5s after the signal, then drains. Against the built binary, `/health` returns 200 at t+1s, t+2s and t+3s, then refuses. Development skips the delay.

## Races

`tackle.Consumer.Stop()` returns early while `State == StateNotListening` (`consumer.go:77`), and `State` becomes listening only at `consumer.go:173`, after `connect()` finishes its network I/O. A `Stop` in that window is discarded, and `consume()` then blocks on `<-c.shutdown` with nobody left to close it. Shutdown retries `Stop` until the consumer returns.

`Server.Shutdown` had the same shape when it ran before `Serve` assigned `httpServer`. `Serve` now refuses to start once `Shutdown` has run.

A deliberate `Stop` also made `Consumer.Start` return nil, which the reconnect path reported as a dropped connection. Every deploy would have logged a reconnect that never happens.

## Left out

The EventDistributer stays outside the drain. `Start` blocks on its own channel while the real work runs in per-route consumers created inside a loop, with no handle to stop them. Joining the drain would report every worker stopped while those consumers still ran. The pprof server keeps no handle either. Both want their own change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,13 @@
 # PPROF_ENABLED=yes
 # PPROF_PORT=6060
 
+## Graceful shutdown
+##
+## How long the process waits on SIGTERM/SIGINT for in-flight work to drain,
+## tracing to shut down and Sentry to flush. Keep it below the deploy target's
+## termination grace period, which is 30s by default on Kubernetes.
+# SHUTDOWN_TIMEOUT=25s
+
 ## Database timeout overrides
 ##
 ## Values can be Go durations such as 60s/2m or integer milliseconds.
--- a/.env.example
+++ b/.env.example
@@ -53,6 +53,12 @@
 ## tracing to shut down and Sentry to flush. Keep it below the deploy target's
 ## termination grace period, which is 30s by default on Kubernetes.
 # SHUTDOWN_TIMEOUT=25s
+##
+## How long the public API keeps serving after the signal, before it stops
+## listening. Kubernetes removes the pod from its Endpoints asynchronously, so
+## traffic still arrives for a moment after SIGTERM. Skipped when
+## APP_ENV=development.
+# SHUTDOWN_DRAIN_DELAY=5s
 
 ## Database timeout overrides
 ##
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -32,4 +32,10 @@ migrate -lock-timeout "${DB_MIGRATION_LOCK_TIMEOUT}" -source file:///app/db/migr
 migrate -lock-timeout "${DB_MIGRATION_LOCK_TIMEOUT}" -source file:///app/db/data_migrations -database "${DB_URL}&x-migrations-table=data_migrations" up
 
 echo "Starting server..."
-/app/build/superplane
+
+#
+# exec replaces this shell, so the server runs as PID 1 and receives SIGTERM
+# directly. Without it the shell keeps the signal, the server is only killed
+# when the grace period expires, and no graceful shutdown ever runs.
+#
+exec /app/build/superplane
--- a/pkg/public/server.go
+++ b/pkg/public/server.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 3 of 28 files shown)


Vandit1604:
Heads up that this PR is on the larger side (~1.1k lines), mostly because graceful shutdown is cross-cutting: it touches the server, every worker, and the email consumers. A fair chunk of that is one small drain pattern repeated across ~15 worker files plus the tests, so the actual logic to review is smaller than the diff suggests.

If it is easier to review in pieces, I am happy to split it into three stacked PRs:

1. **Drain in-flight worker tasks on context cancellation.** The `drainTasks` helper plus the same guard applied across the worker loops. Self-contained, no server changes.
2. **Add `Server.Shutdown` to the public server.** Graceful HTTP shutdown that stops accepting new connections and waits for in-flight requests. Self-contained, with its own test.
3. **Wire SIGTERM orchestration.** The signal handler, the WaitGroup, and the shutdown deadline that cancels the workers, drains the consumers, and calls `Shutdown`. This is where the first two pieces get activated.

Phases 1 and 2 are independent and can go in either order. Phase 3 builds on both.

Just let me know which you would prefer to review and I will restructure it. Happy to keep it as one if that is simpler on your end.

PR #7166, opened 2026-09-03
docs: Front-load generated-file rules in AGENTS.md

lucaspin:
Factory agents often read only the first part of AGENTS.md. They then search git for pb.gen output that Git ignores. Move those facts to the top, move model transaction rules to pkg/models/AGENTS.md, and move Cursor Cloud notes out of the factory briefing.


lucaspin:
👍

PR #7294, opened 2026-09-09
fix: Bound worker poll queries with SQL LIMIT

### LINKED ISSUE #6989: Bound database work for polling workers
## Summary

Several background workers load every eligible database row before they apply their in-process concurrency limit. Most workers allow 25 active tasks, but their polling queries have no row limit.

This pattern affects pending events, node executions, node requests, integration requests, webhooks, app messages, and deleted-resource cleanup. Each worker replica can read the same complete backlog before record locking selects the worker that performs the work.

## Impact

- Large backlogs create large query results and memory allocations.
- Worker replicas repeat reads for the same records.
- Replicas compete for locks on records that another replica already selected.
- A single poll can spend a long time waiting for local semaphore capacity.
- Database load increases during incidents, when backlog size is already high.

## Expected behavior

Each poll reads and claims a bounded batch that matches available worker capacity. Additional worker replicas should divide the backlog without reading all eligible rows.

## Current behavior

Polling queries return the complete eligible set. The concurrency limit applies only after the records are loaded into the worker process.

siddiqueirshad:
## Summary

Polling workers capped concurrency in process, but their ListPending and ListDeleted queries still loaded every eligible row. On a large backlog, each replica repeated that full read before SKIP LOCKED chose a worker.

This change passes a batch limit into those list helpers and applies LIMIT in SQL. The batch size is 100, the same sweep size used by the repository provisioner and run finalizer. Queries also order oldest-first so a deep backlog does not starve early work. The pending-run sweep no longer loads the full set and then slices in memory.

Closes #6989

## Test plan

- [ ] Run `make test PKG_TEST_PACKAGES=./pkg/models` and confirm the new poll-limit tests pass
- [ ] Create more than 100 pending webhooks and confirm one WebhookProvisioner tick claims at most 100 rows
- [ ] Create more than 100 due integration requests and confirm ListIntegrationRequests returns at most the batch size
- [ ] Confirm pending executions and pending runs still progress under load with the new oldest-first order
- [ ] Confirm deleted-resource cleanup workers still drain soft-deleted rows across ticks

### DIFF
--- a/pkg/models/app_message.go
+++ b/pkg/models/app_message.go
@@ -39,15 +39,20 @@ func CreateAppMessage(tx *gorm.DB, canvasID uuid.UUID, nodeID string, payload an
 	return tx.Create(message).Error
 }
 
-func ListAppMessages(tx *gorm.DB) ([]AppMessage, error) {
+func ListAppMessages(tx *gorm.DB, limit int) ([]AppMessage, error) {
+	if limit <= 0 {
+		limit = 100
+	}
+
 	var messages []AppMessage
 
 	query := tx.
 		Table("app_messages").
-		Select("app_messages.*")
+		Select("app_messages.*").
+		Order("app_messages.created_at ASC").
+		Limit(limit)
 
 	err := withActiveCanvas(query, "app_messages.canvas_id").
-		Order("app_messages.created_at ASC").
 		Find(&messages).
 		Error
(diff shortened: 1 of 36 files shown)


siddiqueirshad:
Hi maintainers, when you get a chance, could you please review this PR? Thanks!

